PR #172, opened 2026-04-26
Fix crosscheck appendTail buffer comparison

### LINKED ISSUE #171: Crosscheck divergence in appendTail
The following example produces a `CrosscheckException`.

If I try with just SafeRE or `java.util.regex` it seems to produce the same result. I haven't debugged this carefully, I'm not sure if it's a bug in SafeRE or in crosscheck.

It looks like maybe crosscheck is creating a separate `StringBuilder` for `java.util.regex` operations and writing directly to the input `StringBuilder` for SafeRE operations, but that assumes that the client isn't also operating directly on the `StringBuilder`. If that's the problem, it could create temporary `StringBuilder`s for both SafeRE and `java.util.regex` operations, check if they produce the same result, and then write them to the actual `StringBuilder` instead.

```java
public class T {
    String regex = "\\{\\{(.+?)\\}\\}";
    var p = org.safere.crosscheck.Pattern.compile(regex);
    var text =
"""
{{option1_string}};
{{option2_string}};
{{option3_string}};
{{option4_string}};
{{option5_string}};
hello
""";
    var matcher = p.matcher(text);
    StringBuffer expandedText = new StringBuffer();
    while (matcher.find()) {
      var g = matcher.group(1);
      matcher.appendReplacement(expandedText, "");
      expandedText.append(g);
    }
    matcher.appendTail(expandedText);
    System.err.println(expandedText);
  }
}
```

```
Exception in thread "main" org.safere.crosscheck.CrosscheckException: Crosscheck divergence in appendTail():
  SafeRE: option1_string;
option2_string;
option3_string;
option4_string;
option5_string;
hello

  JDK:    ;
;
;
;
;
hello


API call trace (22 calls):
  [1] find() → true
  [2] find() → start=0, end=18
  [3] group(1) → option1_string
  [4] appendReplacement("") → (appended)
  [5] find() → true
  [6] find() → start=20, end=38
  [7] group(1) → option2_string
  [8] appendReplacement("") → (appended)
  [9] find() → true
  [10] find() → start=40, end=58
  [11] group(1) → option3_string
  [12] appendReplacement("") → (appended)
  [13] find() → true
  [14] find() → start=60, end=78
  [15] group(1) → option4_string
  [16] appendReplacement("") → (appended)
  [17] find() → true
  [18] find() → start=80, end=98
  [19] group(1) → option5_string
  [20] appendReplacement("") → (appended)
  [21] find() → false
  [22] appendTail() → DIVERGENCE
        SafeRE: option1_string;
option2_string;
option3_string;
option4_string;
option5_string;
hello

        JDK:    ;
;
;
;
;
hello


	at org.safere.crosscheck.Matcher.throwDivergence(Matcher.java:445)
	at org.safere.crosscheck.Matcher.checkEqual(Matcher.java:434)
	at org.safere.crosscheck.Matcher.appendTail(Matcher.java:234)
	at org.safere.crosscheck.T.main(T.java:25)
```

eaftan:
## Summary
- compare crosscheck appendReplacement and appendTail using temporary JDK buffers seeded from the caller buffer
- preserve user mutations between append operations for StringBuilder and StringBuffer
- add regression coverage for direct buffer mutation during appendReplacement loops

Fixes #171

## Tests
- mvn -pl safere-crosscheck test -q
- mvn -pl safere,safere-crosscheck test -q

Full mvn test -q was attempted, but safere-ffm-re2 failed while trying to clone Abseil from GitHub because sandbox DNS/network access is blocked.

### DIFF
--- a/safere-crosscheck/src/main/java/org/safere/crosscheck/Matcher.java
+++ b/safere-crosscheck/src/main/java/org/safere/crosscheck/Matcher.java
@@ -194,11 +194,11 @@ public final class Matcher implements MatchResult {
    * {@link #appendTail(StringBuilder)} to finalize and crosscheck.
    */
   public Matcher appendReplacement(StringBuilder sb, String replacement) {
+    String before = sb.toString();
     safereMatcher.appendReplacement(sb, replacement);
-    // JDK side uses its own buffer — compared at appendTail.
-    jdkSb().append(""); // ensure initialized
-    jdkMatcher.appendReplacement(jdkSb(), replacement);
-    trace.recordMatch("appendReplacement", quote(replacement), "(appended)");
+    StringBuilder jdkSb = new StringBuilder(before);
+    jdkMatcher.appendReplacement(jdkSb, replacement);
+    checkEqual("appendReplacement", quote(replacement), sb.toString(), jdkSb.toString());
     return this;
   }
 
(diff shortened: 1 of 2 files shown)


cushon:
crosscheck is going to have performance overhead, and the best approach could depend on usage patterns, but in general I wonder if it'd be more optimal to avoid `toString` the entire existing `StringBuilder` and only compare the part that's being appended? Something like

```java
  public Matcher appendReplacement(StringBuilder sb, String replacement) {
    StringBuilder jdkSb = new StringBuilder();
    StringBuilder safeSb = new StringBuilder();
    safereMatcher.appendReplacement(safeSb, replacement);
    jdkMatcher.appendReplacement(jdkSb, replacement);
    checkEqual("appendReplacement", quote(replacement), safeSb.toString(), jdkSb.toString());
    sb.append(safeSb);
    return this;
  }
```

eaftan:
Good feedback.  Filed as #177 so I don't lose track of it.

PR #365, opened 2026-05-07
Recognize IsUnknown as a character class

### LINKED ISSUE #357: IsUnknown Unicode group
java.util.regex recognizes `IsUnknown` as a unicode group, SafeRE does not.

I think it is defined as a script property here: https://www.unicode.org/reports/tr24/

```java
public class Repro {
        java.util.regex.Pattern.compile("\\p{IsUnknown}").matcher("").find();
        org.safere.Pattern.compile("\\p{IsUnknown}").matcher("").find();
    }
}
```

cushon:
Fixes https://github.com/eaftan/safere/issues/357


eaftan (approved):
LGTM, thanks!

Issue #360, opened 2026-05-07
Crosscheck divergence in hitEnd


cushon:
```java
        Pattern p = Pattern.compile("^fooseball");
        Matcher m = p.matcher("fort");
        m.find();
        m.hitEnd();
```

```
Exception in thread "main" org.safere.crosscheck.CrosscheckException: Crosscheck divergence in hitEnd():
  SafeRE: true
  JDK:    false

API call trace (2 calls):
  [1] find() → false
  [2] hitEnd() → DIVERGENCE
        SafeRE: true
        JDK:    false
```

cushon:
Support for `hitEnd` was removed in https://github.com/eaftan/safere/pull/443, see also https://github.com/eaftan/safere#not-supported

PR #366, opened 2026-05-07
Fix hitEnd() behavior

cushon:
https://github.com/eaftan/safere/issues/360

SafeRE's Matcher.hitEnd() was returning true for all failed find() operations, java.util.regex.Matcher only returns true if the search actually reached the end of input.

This change updates the DFA and NFA engines to track whether they reached the end of input, and use this in Matcher to correctly set lastHitEnd on failure. Acceleration paths also set this flag when scanning to the end.


cushon:
This seems like maybe SafeRE isn't following the spec and worth fixing. I haven't checked if the changes here affect performance at all, since it does have to track some additional state.

safere/src/main/java/org/safere/Matcher.java
eaftan: I believe you need to wire this path into the new `hitEnd` tracking.  

`onePass.search(...)` still only returns `int[]` or `null`, unlike the updated DFA/NFA paths that now return whether the engine reached the end of input. When OnePass returns `null`, `Matcher` records a failed match, but it has no OnePass-specific `hitEnd` value to copy into `lastEngineHitEnd`. Since `updateEndState()` now trusts `lastEngineHitEnd` for failed `find()` calls, this path can report `hitEnd=false` for failures that should report `true`.

Repro on this PR:

```java
Pattern p = Pattern.compile("^forts");
Matcher m = p.matcher("fort");

assertThat(m.find()).isFalse();
assertThat(m.hitEnd()).isTrue(); // JDK behavior; currently false on this PR
```

OnePass likely needs to expose the same `{groups, hitEnd}` shape as the updated DFA/NFA paths, or this path needs another principled way to set `lastEngineHitEnd` from actual OnePass search state.  Please also add the above test to the test cases.

eaftan, line 1939: I think there's a similar problem here where we need to thread through the `hitEnd` state.  Fortunately it's easier to do here.

`keywordAlternation` is SafeRE’s internal optimization for a narrow regex shape: fixed literal alternatives used as whole words, e.g. `(?i)\b(?:error|warning)\b`. When `Pattern` recognizes that shape, `Matcher.doFindCore()` bypasses the normal DFA/NFA/OnePass engines and calls `findKeywordAlternation(...)`, which scans the input directly.

That means this helper is acting as the search engine for this path. If the loop reaches the end without finding a keyword, it should set `lastEngineHitEnd = true` before returning `NoMatchResult`; otherwise `updateEndState()` reads the default/stale value and `hitEnd()` can be wrong.

Repro:

```java
Pattern p = Pattern.compile("(?i)\\b(?:error|warning)\\b");
Matcher m = p.matcher("ok");

assertThat(m.find()).isFalse();
assertThat(m.hitEnd()).isTrue(); // currently false
```

This one probably does not need a new result type because the helper owns the whole scan and the final no-match return only happens after exhausting the input.

Also, can you add the above to the test cases?

```suggestion
    this.lastEngineHitEnd = true;
    return applyEngineResult(new NoMatchResult());
```

eaftan (changes requested):
My agent found a couple of bugs including repros which you can turn into test cases.  

Also you'll need to rebase on main and resolve conflicts.  Sorry about that.

safere/src/main/java/org/safere/Matcher.java
cushon: Thanks, fixed and added the test

cushon: Done

cushon:
Thanks for the review!

I think I've addressed the comments and added the suggested test coverage.

I merged in the latest changes, and had to make a couple more tweaks including adding hitEnd handling for the character class fast path added in https://github.com/eaftan/safere/commit/e9488d5a8101266a827ad11ced2969f8f3fe69bd.

safere/src/main/java/org/safere/Matcher.java
eaftan, line 441: One more path where we need to set `lastEngineHitEnd` to true.

My agent says the following test case will fail without this:

```java
Pattern p = Pattern.compile("[abc]");
Matcher m = p.matcher("xyz");

assertThat(m.find()).isFalse();
assertThat(m.hitEnd()).isTrue(); // currently false
```

```suggestion
    this.lastEngineHitEnd = true;
    return applyEngineResult(new NoMatchResult());
```

eaftan (changes requested):

safere/src/main/java/org/safere/Matcher.java
cushon: Fixed, thanks

eaftan (changes requested):
I think the tests are failing now, you should be able to see it in the CI output.

cushon:
> I think the tests are failing now, you should be able to see it in the CI output.

Sorry, I did some more work on this and the tests are passing.

I added some more test coverage and investigated some divergences from the JDK. In some of these cases SafeRE is more correct, and this may be a case where following the spec makes more sense than try to match JDK behaviour exactly. I also took a stop back to think about how `hitEnd` is used and am having doubts about whether it is worth supporting in SafeRE, I will file a separate issue.

The 'lastEngineHitEnd' stuff here is actually fixing a bug and makes sense to do if SafeRE supports `hitEnd`.

cushon:
I pulled in the latest changes and fixed a merge conflict.

Given #411 this is low priority, and there are edge cases where this doesn't fully match `java.util.regex` behaviour. But as long as SafeRE supports these methods I think this PR is an improvement over the status quo.

eaftan (approved):

Issue #361, opened 2026-05-07
Crosscheck divergence in find - surrogates in negated character classes


cushon:
```java
 org.safere.crosscheck.Pattern.compile("[^a-z]..").matcher("\ud801\ud800\udc00xyz").find();
```

```
Exception in thread "main" org.safere.crosscheck.CrosscheckException: Crosscheck divergence in find → start()():
  SafeRE: 1
  JDK:    0

API call trace (2 calls):
  [1] find() → true
  [2] find.start() → DIVERGENCE
        SafeRE: 1
        JDK:    0
```

cushon:
I also filed https://bugs.openjdk.org/browse/JDK-8384092 to clarify if this is deliberate behaviour

PR #367, opened 2026-05-07
Include surrogates in negated character classes

cushon:
SafeRE's CharClassBuilder.negate() was explicitly excluding surrogate code points when negating a character class. This caused a divergence from java.util.regex, which treats isolated surrogates as normal characters that can match a negated class.

https://github.com/eaftan/safere/issues/361


cushon:
I'm not sure if the java.util.regex behaviour this matches is correct or not, this could wait for clarification in https://bugs.openjdk.org/browse/JDK-8384092

cushon:
See https://github.com/eaftan/safere/issues/361#issuecomment-4399504439

PR #426, opened 2026-05-16
Fix JDK syntax compatibility for dangling quantifiers on reluctant quantifiers (`^+?+`)

### LINKED ISSUE #391: SafeRE accepts a regex that java.util.regex rejects - dangling meta character
This looks similar to #272 but still reproduces at head

```java
package org.safere;

import java.util.regex.PatternSyntaxException;

public class R {

    String regex = "^+?+";
    System.err.println("java.util.regex");
    try {
      java.util.regex.Pattern.compile(regex);
      System.err.println("ok");
    } catch (PatternSyntaxException e) {
      System.err.println(e.getMessage());
    }
    System.err.println("SafeRE");
    try {
      org.safere.Pattern.compile(regex);
      System.err.println("ok");
    } catch (PatternSyntaxException e) {
      System.err.println(e.getMessage());
    }
  }

  private R() {}
}
```

```
java.util.regex
Dangling meta character '+' near index 3
^+?+
   ^
SafeRE
ok
```

cushon:
Standard `java.util.regex` rejects applying a quantifier directly to a reluctant quantifier (such as `+` applied to `+?` in `^+?+` or `a+?+`) with a `PatternSyntaxException` ("Dangling meta character '+'").

SafeRE correctly rejected nested quantifiers on standard atoms (`a+?+`) as invalid nested repetition operators. However, its coalescing rule for repeated zero-width assertions (`canRepeatAfterUnary()`) inspected only whether the target expression was zero-width (`isQuantifiedZeroWidth`), without verifying whether the preceding quantifier was greedy (`+`) or reluctant (`+?`). As a result, SafeRE incorrectly permitted `+` to coalesce with `+?` on zero-width assertions, accepting invalid patterns like `^+?+`.

**Changes:**
- Updated `canRepeatAfterUnary()` in `Parser.java` to require that the preceding quantifier on the parse stack was strictly greedy (`stacktop.re.op == RegexpOp.PLUS && !stacktop.re.nonGreedy()`).
- Attempts to quantify a reluctant quantifier on an anchor or empty group (e.g., `^+?+`, `()+?+`) are now immediately rejected with a `PatternSyntaxException` ("invalid nested repetition operator").
- Existing valid coalescing behaviors (like `^++`, `^{3,}`, or comments-mode coalescing `(?x) ^+ \n\n +^`) remain fully supported.
- Added parameterized test coverage in `ParserTest.java` verifying rejection across `^+?+`, `$*?+`, `\b+?+`, `()+?+`, and `a+?+`.

Fixes https://github.com/eaftan/safere/issues/391

### DIFF
--- a/safere/src/main/java/org/safere/Parser.java
+++ b/safere/src/main/java/org/safere/Parser.java
@@ -684,7 +684,11 @@ final class Parser {
   }
 
   private boolean canRepeatAfterUnary(RegexpOp op) {
-    return op == RegexpOp.PLUS && stacktop != null && isQuantifiedZeroWidth(stacktop.re);
+    return op == RegexpOp.PLUS
+        && stacktop != null
+        && stacktop.re.op == RegexpOp.PLUS
+        && !stacktop.re.nonGreedy()
+        && isQuantifiedZeroWidth(stacktop.re);
   }
 
   private boolean isQuantifiedZeroWidth(Regexp re) {
--- a/safere/src/test/java/org/safere/ParserTest.java
+++ b/safere/src/test/java/org/safere/ParserTest.java
@@ -764,6 +764,13 @@ class ParserTest {
       Regexp re = parse("a*\\{");
       assertThat(re.op).isEqualTo(RegexpOp.CONCAT);
     }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"^+?+", "$*?+", "\\b+?+", "()+?+", "a+?+"})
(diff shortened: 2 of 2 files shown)


eaftan:
My agent found a regression introduced by this patch: #428 

That said, I'm going to merge this and fix forward.  I think the right way to approach that bug is by adding a new exhaustive sweep for this entire class of problem (#427), then drive that to zero divergences.  Every time we point fix something, we risk introducing bugs for things we don't have tests for.

Issue #404, opened 2026-05-11
Parser character-class performance regression test times out


eaftan:
Found while validating issue #374 after rebasing on current `main`.

## Repro

```bash
mvn -pl safere test -q
```

The full suite consistently hit the 10s preemptive timeout in:

```text
org.safere.ParserTest$CharacterClasses.manyPlainCharacterClassesCompileWithinRegressionBound
```

## Observed

```text
execution timed out after 10000 ms
at org.safere.ParserTest$CharacterClasses.manyPlainCharacterClassesCompileWithinRegressionBound(ParserTest.java:345)
Caused by: Execution timed out in thread junit-timeout-thread-2
  at java.util.Arrays.copyOf(Arrays.java:3589)
  at org.safere.Regexp.literalString(Regexp.java:108)
  at org.safere.Parser.maybeConcatString(Parser.java:988)
  at org.safere.Parser.pushRegexp(Parser.java:509)
  at org.safere.Parser.doParse(Parser.java:262)
  at org.safere.Parser.parse(Parser.java:154)
```

This appears unrelated to the grapheme-cluster change in issue #374: the stack is in the character-class/literal parser path, and the focused grapheme tests and new grapheme exhaustive sweep pass.

## Notes

The run was stopped after the timeout was observed to avoid continuing the remaining exhaustive tests after Maven had already recorded the failure.

PR #442, opened 2026-05-19
Add support for Python style named capturing groups

cushon:
This adds support for Python style `?P<GROUP_NAME>` named capturing groups, with the `P` and allowing underscores in group names.

This syntax isn't allowed by `java.util.regex`, but is supported by RE2 family implementations like RE2J. Supporting it here provides an easier migration path for users of RE2 regex engines.

This PR unconditionally allows the syntax in `Pattern.compile`. I thought about making it configurable, but the current set of `Pattern` flags exactly matches `java.util.regex`, and it seems nice to keep them in sync (especially if `Pattern` ever added another flag in the future). It'd be possible to have a separate set of SafeRE-specific configuration and expose that in the `Pattern` API. I also don't see significant downsides to unconditionally supporting the syntax, I think the `?P` syntax has been [claimed by Python](https://web.archive.org/web/20221023212326/http://markmail.org/message/oyezhwvefvotacc3) and is unlikely to collide with anything or cause problems.

### DIFF
--- a/safere-fuzz/src/test/java/org/safere/fuzz/DialectSyntaxFuzzer.java
+++ b/safere-fuzz/src/test/java/org/safere/fuzz/DialectSyntaxFuzzer.java
@@ -13,7 +13,6 @@ final class DialectSyntaxFuzzer {
 
   private static final String[] CONTEXT_PREFIXES = {"", "^", "(?:", "a|", "[", "[^"};
   private static final String[] DIALECT_FRAGMENTS = {
-    "(?P<name>a)",
     "(?P=name)",
     "(?'name'a)",
     "\\g{name}",
@@ -42,7 +41,6 @@ final class DialectSyntaxFuzzer {
           "", "a", "b", "a\u0300", "\r\n", "Braille", "Latin", ":", "[", "]", "l", "o", "w", "e",
           "r");
   private static final String[] REGRESSION_REGEXES = {
-    "(?P<name>a)",
     "\\p{Braille}",
     "\\p{Latin}",
     "[[:lower:]]",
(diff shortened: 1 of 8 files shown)


cushon:
I can't reproduce the test failure locally and it's not clear to me how it's related, I wonder if the test could be flaky.

```
Error:  org.safere.ParserTest$CharacterClasses.manyPlainCharacterClassesParseWithLinearScaling -- Time elapsed: 0.152 s <<< FAILURE!
java.lang.AssertionError: 
[4x input should stay near linear, smaller=2833794ns larger=33575616ns] 
Expecting actual:
  33575616L
to be less than:
  28337940L 
	at org.safere@0.3.0-SNAPSHOT/org.safere.ParserTest$CharacterClasses.manyPlainCharacterClassesParseWithLinearScaling(ParserTest.java:364)
```

eaftan:
Yeah, probably a flaky test.  I don't have a good answer for how to write tests that catch scaling performance or memory regressions without being flaky.  I had another one that I thought was flaky (#404), but turned out to catch an actual performance regression.  

Will look at this tonight, but don't worry about that test. I'll do something about it as well.

eaftan:
@cushon I made a few small changes directly and will merge this if it passes CI. I did it that way for expediency, mostly so we can get this in before I go to bed rather than adding another timezone round trip.

For future PRs, let me know what you prefer. I’m happy either to leave feedback and have you make the follow-up changes, or to make small fixups myself when that’s faster.

cushon:
Thanks! The changes look good, and please feel free to make changes to PRs in the future to minimize round trips.

PR #446, opened 2026-05-23
Fix leftmost find start for overlapping suffixes

### LINKED ISSUE #445: Crosscheck divergence in find
This one affects real world code. It doesn't seem to be very common.

```java

public class R {
    String input = "Foo '{0}' Bar: {0}";
    String regex = "[^{']*(?:'[^']*'[^{']*)*\\{([^}]*)\\}";
    Pattern p = Pattern.compile(regex);
    Matcher m = p.matcher(input);
    while (m.find()) {
      System.err.println(m.start());
    }
  }
}
```

```
Exception in thread "main" org.safere.crosscheck.CrosscheckException: Crosscheck divergence in find → start()():
  SafeRE: 5
  JDK:    0

API call trace (2 calls):
  [1] find() → true
  [2] find.start() → DIVERGENCE
        SafeRE: 5
        JDK:    0

	at org.safere.crosscheck.Matcher.throwDivergence(Matcher.java:499)
	at org.safere.crosscheck.Matcher.checkMatchState(Matcher.java:431)
	at org.safere.crosscheck.Matcher.find(Matcher.java:66)
	at org.safere.crosscheck.R.main(R.java:9)
```

eaftan:
## Summary

- Preserve leftmost `find()` starts when greedy leading context overlaps a later required literal.
- Stop using unanchored OnePass as an authoritative `find()` producer; fall through to the linear DFA/BitState/NFA pipeline instead.
- Guard the DFA sandwich so a later reverse-DFA start falls through to the submatch engine.
- Add matcher regression coverage and FindSequenceFuzzer coverage for the issue shape.

Fixes #445

## Verification

- `mvn -pl safere -Dtest=MatcherTest test`
- `mvn -pl safere -Dtest=EnginePathEquivalenceTest test`
- `mvn -pl safere -Dtest=MatcherTest,EnginePathEquivalenceTest test`
- `mvn -pl safere-fuzz -am -Dtest=FindSequenceFuzzer -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -pl safere test -q`

Not run: full public API crosscheck, because another heavy crosscheck JVM was already running in a separate workspace and running another in parallel would compete for CPU.

### DIFF
--- a/safere-fuzz/src/test/java/org/safere/fuzz/FindSequenceFuzzer.java
+++ b/safere-fuzz/src/test/java/org/safere/fuzz/FindSequenceFuzzer.java
@@ -7,22 +7,41 @@ package org.safere.fuzz;
 
 import com.code_intelligence.jazzer.api.FuzzedDataProvider;
 import com.code_intelligence.jazzer.junit.FuzzTest;
+import org.junit.jupiter.api.Test;
 
 final class FindSequenceFuzzer {
+
+  @Test
+  void delimitedPrefixBeforeRequiredSuffixRegression() {
+    FuzzSupport.CompiledPattern pattern =
+        FuzzSupport.compileOrSkip("[^{']*(?:'[^']*'[^{']*)*\\{([^}]*)\\}", 0);
+
+    pattern.matcher("Foo '{0}' Bar: {0}").find();
+  }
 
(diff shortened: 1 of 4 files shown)


cushon:
Thanks!

Issue #425, opened 2026-05-15
Evaluate generated Unicode tables for compile-time performance and reproducibility


eaftan:
## Problem

SafeRE currently derives several Unicode tables from the running JDK's `Character` data so regex behavior tracks the JDK Unicode version automatically. That has a nice compatibility property, but the current case-folding work exposed a performance risk: even when base tables are cached once per JVM, derived folded expansions can still do substantial work during `Pattern.compile()`.

Recent investigation during the case-folding character-class sweep showed workers spending most CPU in compile-time Unicode case expansion:

- `Parser.addUnicodeFoldedRange`
- `Parser.unicodeCaseVariantInRange`
- `Character.toUpperCase` / `Character.toLowerCase` / `Character.toTitleCase`
- `Parser.lookupCaseFold`

This was not GC thrashing: GC time was negligible relative to elapsed time. The sweep was CPU-bound in repeated Unicode folding work.

The immediate bug fix can use runtime-derived caches, but we should reconsider whether SafeRE should dynamically track the running JDK's Unicode version at all for parser/runtime performance and reproducibility.

## Proposal to evaluate

Consider moving Unicode data to generated, checked-in tables with an explicit SafeRE Unicode version. This could include:

- General category and script range tables.
- Binary Unicode properties SafeRE supports.
- Java-style property tables such as `javaUpperCase`, `javaLowerCase`, and `javaTitleCase` if we continue supporting those names.
- Case-fold closure helper tables needed for `CASE_INSENSITIVE` and `UNICODE_CASE` character-class expansion.

This would trade automatic adaptation to the running JDK's Unicode version for stable, reproducible, fast Unicode semantics. Canary tests can compare the checked-in tables against the running JDK and report drift explicitly when upgrading JDKs, instead of silently changing behavior or paying runtime table construction costs.

A lighter-weight alternative is to keep runtime-derived source tables but cache every derived folded range/table result. That may be enough, but we should measure it against generated tables before deciding.

## Benchmark/workload plan

To make the decision evidence-based, add benchmarks that isolate compile-time and first-use costs for Unicode-heavy regexes. Suggested workloads:

### Compile-only microbenchmarks

Measure `Pattern.compile()` throughput/allocation for patterns such as:

- `\\p{L}+`
- `\\p{Lu}+`
- `\\p{Ll}+`
- `\\p{Lt}+`
- `\\p{javaUpperCase}+`
- `\\p{javaLowerCase}+`
[...]

PR #459, opened 2026-06-02
Generate Unicode property tables

### LINKED ISSUE #458: Consider moving Unicode data to generated, checked-in tables
This was previously raised as a possibility in #425, I'm re-opening to track it.

> SafeRE currently derives several Unicode tables from the running JDK's Character data so regex behavior tracks the JDK Unicode version automatically. That has a nice compatibility property, but the current case-folding work exposed a performance risk: even when base tables are cached once per JVM, derived folded expansions can still do substantial work during Pattern.compile().
>
> Consider moving Unicode data to generated, checked-in tables with an explicit SafeRE Unicode version.

The motivation for this was that migrating an example command line tool to SafeRE made performance measurably worse, and the profile showed that a lot of the time in SafeRE was spent in:

```
org.safere.UnicodeProperties.buildRanges(java.util.function.IntPredicate)
org.safere.UnicodeProperties$BinaryHolder$$Lambda.test(int)
org.safere.UnicodeGroups.buildAll()
```

I think short running command line tool may be a less important use-case than longer running processes like servers, and there's a still a tradeoff here with dealing with Unicode versioning. But I think this might be worth doing eventually.

eaftan:
## Summary

Fixes #458.
Refs #425.

This moves SafeRE-owned Unicode property data from runtime-derived `Character` scans to a checked-in generated table resource produced with `~/jdks/jdk-26.0.1` and recorded as Unicode 17.0 data.

The `java*` property family remains runtime-JDK based because the JDK defines those regex properties in terms of `java.lang.Character` predicates.

## Profile motivation

The motivating profile for a short-running command-line migration showed startup/compile time concentrated in runtime Unicode table construction:

```text
org.safere.UnicodeProperties.buildRanges(java.util.function.IntPredicate)
org.safere.UnicodeProperties$BinaryHolder$$Lambda.test(int)
org.safere.UnicodeGroups.buildAll()
```

This change removes those runtime construction paths for ordinary SafeRE Unicode regex properties:

- `UnicodeGroups.buildAll()` is replaced by generated category/script tables loaded from `unicode-tables.bin`.
- `UnicodeProperties.buildRanges(...)` and `BinaryHolder` are removed from non-`java*` Unicode property lookup.
- Binary properties, categories, scripts, and blocks are loaded from the generated resource.

## Implementation

- Adds `tools/unicode/UnicodeTableGenerator.java` and generation/verification scripts.
- Adds `safere/src/main/resources/org/safere/unicode-tables.bin` generated from JDK 26.0.1.
- Adds `UnicodeGeneratedTables` to load the binary resource once.
- Refactors `UnicodeGroups` and `UnicodeProperties` to use generated tables.
- Keeps bounded runtime JDK alias resolution for script/block aliases such as `GreekandCoptic`, but still maps to generated SafeRE tables.
- Documents the Unicode version policy in `DESIGN.md`.
- Adds Unicode first-compile/warm-compile benchmark scaffolding.

## Benchmark results

Benchmarks compare baseline `5ac18ec` against this branch. Raw JMH output was saved under `target/unicode-benchmark-results/{baseline,current}/` locally.

First-compile benchmark settings:

```text
-f 10 -wi 0 -i 1 -r 1 -bm ss
```

| Scenario | Before | After | Speedup |
|---|---:|---:|---:|
| `\p{IsAlphabetic}+` | 286.272 ms | 27.197 ms | 10.53x |
| `\p{IsExtended_Pictographic}+` | 280.473 ms | 24.412 ms | 11.49x |
| `\w+` + `UNICODE_CHARACTER_CLASS` | 276.448 ms | 26.315 ms | 10.51x |
| `\p{block=BasicLatin}+` | 88.297 ms | 25.178 ms | 3.51x |
| `\p{L}+` | 70.484 ms | 24.944 ms | 2.83x |
| `\p{script=Latin}+` | 67.506 ms | 22.296 ms | 3.03x |
| `[a-z]+` + `CASE_INSENSITIVE \| UNICODE_CASE` | 134.932 ms | 108.708 ms | 1.24x |

First-compile geomean across these scenarios: 4.65x faster.

Warm repeated-compile quick benchmark settings:

```text
-f 1 -wi 3 -w 1 -i 5 -r 1
```

| Scenario | Before | After |
|---|---:|---:|
| `\p{IsAlphabetic}+` | 30.470 us | 30.843 us |
| `\p{IsExtended_Pictographic}+` | 3.001 us | 5.690 us |
| `\w+` + `UNICODE_CHARACTER_CLASS` | 31.657 us | 32.912 us |
| `\p{block=BasicLatin}+` | 0.999 us | 1.059 us |
| `\p{L}+` | 26.860 us | 27.090 us |
| `\p{script=Latin}+` | 1.920 us | 1.753 us |
| `[a-z]+` + `CASE_INSENSITIVE \| UNICODE_CASE` | 1.606 us | 1.604 us |

Warm geomean: 0.91x. The intended win is cold first use; warm steady-state is roughly neutral in this quick run.

## Verification

Ran:

```text
tools/unicode/verify-unicode-tables.sh 17.0
mvn -pl safere -Dtest=UnicodeTablesTest,UnicodePropertySyntaxTest,UnicodeJdkConsistencyTest,JavaCharacterClassesTest test
mvn -pl safere -Dtest=UnicodeTablesTest test
git diff --check
```

Also built and ran focused JMH benchmarks as described above.

Not run:

- Full `mvn -pl safere test -q`.
- Full public API crosscheck validation.


eaftan:
@cushon: I didn't pregenerate the derived Unicode case-closure index used by patterns with `CASE_INSENSITIVE | UNICODE_CASE`. My assumption is that those flags are rare in real-world usage. If you see meaningful usage of that flag combination and cold-start cost shows up there, let me know and we can revisit.

PR #460, opened 2026-06-02
 Generate unicode tables at build time

cushon:
This allows the tables to be automatically updated when rebuilding SafeRE instead of relying on checked in data.

Also generate Java code for the table data instead of relying on Java  serialization, to reduce the security surface area.

https://github.com/eaftan/safere/issues/458


eaftan:
I agree with the direction of generating Java instead of loading serialized resources.

However, I'm a bit concerned about the Maven wiring. Adding the generator as a dependency of the main module means module-specific builds from a clean checkout fail, because Maven tries to resolve `org.safere:safere-unicode-generator:jar:0.3.0-SNAPSHOT` from the local/remote Maven repositories instead of building it first. For example, `mvn -pl safere test -q` and `cd safere && mvn test -q` fail with:

> Could not find artifact org.safere:safere-unicode-generator:jar:0.3.0-SNAPSHOT

I’d also like to preserve the option of generating the Unicode tables with a different JDK than the one used to build/test SafeRE.  This lets me support higher Unicode versions than JDK 21 does, which is what I target for compatibility.

Could we adjust this to keep the generator as a separate module, but make regeneration an explicit maintainer workflow instead of part of the default `safere` Maven lifecycle?

Concretely, I’m thinking:
- Keep `safere-unicode-generator` as a separate module.
- Keep `UnicodeGeneratedTables.java` checked in under `safere/src/main/java/org/safere/`.  The file should have a header explaining that it's generated, should not be edited by hand, and how to regenerate it.
- Remove the generator dependency/plugin execution from the main `safere` module’s default build.
- Add or keep a script/profile for maintainers to build the generator and regenerate the checked-in Java file with an explicitly selected JDK.

That would preserve Maven workflows like `mvn -pl safere test -q`, while still replacing the serialized table resource with generated Java source.

Would that meet your needs?

I’m guessing the build-time generation piece is mostly for wiring this into Bazel, where the generator can be a tool target and the generated Java can be an explicit build output. If we keep the generator module and generated Java format, I think you’d still have the pieces needed to do that on the Bazel side.

eaftan:
@cushon FYI, PR #463 included a change to start pregenerating Unicode category data for `Cn` and `C`.  I don't think you have to do anything other than rebase on main to pick it up, after that it should Just Work™️.  That PR added a test for it, so as long as you rebase and it passes tests, you should be good.

cushon:
> Could we adjust this to keep the generator as a separate module, but make regeneration an explicit maintainer workflow instead of part of the default safere Maven lifecycle?

That all SGTM, thanks, done

> I’m guessing the build-time generation piece is mostly for wiring this into Bazel, where the generator can be a tool target and the generated Java can be an explicit build output. If we keep the generator module and generated Java format, I think you’d still have the pieces needed to do that on the Bazel side.

Exactly, yes, that was the motivation. I was planning to write this up with Bazel to generate the code at build time using the latest JDK STS.

eaftan:
I added a script that runs spotless on the generated code.  Otherwise LGTM.  Enabled auto-merge, so this should merge after CI finishes.

Issue #465, opened 2026-06-04
Performance regression after 'Fix region boundary and Unicode category divergences'


cushon:
I'm seeing a performance regression after https://github.com/eaftan/safere/commit/7bc22a35e25560381bd455d2e10f4d23fe7dc0d0 (Fix region boundary and Unicode category divergences), profiling shows a regex that was previously using the DFA engine is now using the NFA engine. Is that expected?

I'll work on minimizing the specific regex I'm seeing this on.

cushon:
I got an agent to try to reproduce it in a benchmark, this seems like a plausible explanation.

---

## 1. Root Cause of Performance Regression
The performance regression occurs because the patterns used by `MoeScrubber` contain word boundaries (`\b`). 

In `Matcher.java`, the method `dfaSupportsProgram` returned `false` for any program containing a word boundary:
```java
private static boolean dfaSupportsProgram(Prog prog) {
  return !prog.hasGraphemeSemantics() && !prog.hasWordBoundary();
}
```
Because of this check, patterns with `\b` or `\B` were forced to completely bypass the DFA engine and run via the NFA/BitState engine. The NFA/BitState engine is significantly slower (~10x) because it has to spawn and track active search threads at every character position.

### Why was DFA disabled for word boundaries?
The `java.util.regex` engine allows matching word boundaries (such as `\B`) at any character boundary, including *inside* surrogate pairs (i.e. between high and low surrogates).
Because SafeRE's DFA engine is code-point-aligned, it only steps over whole code points and never stops inside surrogate pairs. Consequently, it cannot evaluate or match boundary assertions inside surrogate pairs, resulting in subtle behavior divergences from the JDK on input containing emoji/supplementary characters.

---

## 2. Implemented Optimization
Since the majority of input strings processed by regexes do not contain surrogate pairs, we can run the DFA safely for word boundary patterns on these inputs without any behavioral divergence from the JDK.

We modified `dfaSupportsProgram` to be instance-aware and perform a fast scan of the input string for surrogate characters:
```java
private boolean dfaSupportsProgram(Prog prog) {
  if (prog.hasGraphemeSemantics()) {
    return false;
  }
  if (prog.hasWordBoundary()) {
    return !hasSurrogate(text);
  }
  return true;
}

private static boolean hasSurrogate(String text) {
  int len = text.length();
  for (int i = 0; i < len; i++) {
    char c = text.charAt(i);
    if (c >= '\uD800' && c <= '\uDFFF') {
      return true;
[...]

PR #470, opened 2026-06-07
Port C++ RE2 lazy DFA optimizations and fix reverse search boundaries

cushon:
I came up with this investigating the performance gap for the `ScrubberBenchmark.scrubWithDirectives` benchmark added in #468 as part of #465.

I didn't do this by hand, but it seems like there might be something here. It verified it solves the performance gap for `ScrubberBenchmark`. SafeRE's own tests pass, I can do additional testing on it.

---

This commit ports the NFA instruction flattening and flat epsilon-closure optimizations from C++ RE2 to SafeRE to improve DFA matching performance, and resolves correctness issues in the reverse DFA search at region boundaries.

1. DFA Optimizations (Ported from C++ RE2):
   - NFA Instruction Flattening (Prog.java, Inst.java): Elides ALT/ALT_MATCH instructions, merging them into flat sequential instruction lists. Uses a 'last' flag on Inst to demarcate list boundaries. Flat programs are cloned for DFA usage while maintaining tree structure compatibility for backtracking NFA and BitState engines.
   - Flat Epsilon Closures (Dfa.java): Updates epsilon closure expansion (expand()) to iteratively traverse the flattened instruction layout using LIFO pushes, avoiding state explosion and recursive overhead.
   - Leftmost-First Match Pruning: Prunes lower-priority branches as soon as a match is found, honoring leftmost-first semantics during forward search.
   - Segregated DFA Caches (Pattern.java, Matcher.java): Splits the thread-local DFA cache into separate first-match (cachedForwardFirstMatchDfa) and longest-match (cachedForwardLongestMatchDfa) caches to prevent transition cache collisions due to differing pruning behaviors.

2. DFA Reverse Search and Region Boundaries:
   - Correct Boundary Context (Dfa.java): Extends reverse DFA scanning (pos > 0) beyond the startLimit to correctly read preceding characters for boundary assertions (\b, \B, ^, $) at region limits.
   - Zero-Width Start Bounds (Dfa.java): Fixes end-of-text sentinel transitions (cp < 0) to set FLAG_MATCH_BEFORE on match, allowing the reverse search to correctly identify match starts at the startLimit boundary.

---

ScrubberBenchmark with this change shows that it makes SafeRE faster than `java.util.regex` or RE2/J for `scrubWithDirectives`:

```
Benchmark                                     Mode  Cnt        Score        Error  Units
ScrubberBenchmark.scrubNoDirectives_jdk       avgt    5  7242496.840 ±  75709.674  ns/op
ScrubberBenchmark.scrubNoDirectives_re2j      avgt    5  3965837.984 ±  25789.250  ns/op
ScrubberBenchmark.scrubNoDirectives_safere    avgt    5   646384.066 ±  10229.906  ns/op
ScrubberBenchmark.scrubWithDirectives_jdk     avgt    5  9488435.193 ± 212145.632  ns/op
ScrubberBenchmark.scrubWithDirectives_re2j    avgt    5  7233247.677 ± 617873.610  ns/op
ScrubberBenchmark.scrubWithDirectives_safere  avgt    5  2108599.200 ±  30250.930  ns/op
```

### DIFF
--- a/safere-fuzz/src/test/java/org/safere/fuzz/MatchFuzzer.java
+++ b/safere-fuzz/src/test/java/org/safere/fuzz/MatchFuzzer.java
@@ -48,6 +48,7 @@ final class MatchFuzzer {
       assertFullMatchesSafeRe(regression.regex(), regression.flags(), regression.inputs());
     }
     assertUnicodeBoundaryStartCacheMatchesJdk();
+    assertTrailingLineTerminatorEndAnchorFindsMatchJdk();
 
     String regex;
     int flags;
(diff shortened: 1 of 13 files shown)


cushon:
I thought I had a clean test run for this but the `org.safere.CrossEngineExhaustiveTest$LineEndings.lineEndsExhaustive` failure shows this is now diverging from `java.util.regex`, and I think matching c++ RE2 behaviour. 

This is another place where I feel a bit conflicted between the `java.util.regex` compatibility goals, and the desire for good performance and 'good enough' `java.util.regex` compatible. I'm not sure this case has come up much in practice during previous migrations from `java.util.regex` to RE2 engines, and I wonder if we could get away with RE2 semantics here, or even an option for that.

eaftan:
I fixed the test failure and verified the benchmark results match what you reported.  This LGTM.  Will merge after CI runs.

PR #472, opened 2026-06-10
Update Error Prone version

cushon:
This include a fix (https://github.com/google/error-prone/commit/c3581e93a0db15e35f8f7b3b6bb5cfe5eadbe8dc) where some checks were crashing in `compilesWithFix` on projects that include `module-info.java`s, which SafeRE does.


eaftan (approved):
LGTM, thanks!

PR #473, opened 2026-06-10
Use Objects.requireNonNull

cushon:
This is mostly a tiny readability improvement.

There's no performance downside, `requireNonNull` gets special treatment from the JIT (it has a `@ForceInline` annotation), and it might actually be faster because because it results in slightly smaller bytecode which could affect other inlining decisions.


eaftan (approved):
Looks good.  I also updated the agent instructions to prefer `Objects.requireNonNull` over explicit comparisons to `null`.

PR #478, opened 2026-06-11
Optimize `replaceFirst` and `replaceAll` when there is no match.

eamonnmcmanus:
If there is no match then we can simply return the input string unaltered. There is no need to reconstruct a copy using a `StringBuilder`.
    
This showed up when comparing against RE2-over-JNI in a production-like environment. SafeRE was on average faster for all operations other than `replaceFirst`. With the changes here, it is faster there too.

Also fix a minor incompatibility in `replaceAll(String)`. It does not matter if the replacement string is invalid, when the match doesn't happen anyway. This is consistent with `java.util.regex`, but previously the precomputed `ReplacementSegment[]` would cause an exception even in the absence of a match.

### DIFF
--- a/safere/src/main/java/org/safere/Matcher.java
+++ b/safere/src/main/java/org/safere/Matcher.java
@@ -2129,10 +2129,11 @@ public final class Matcher implements MatchResult {
    */
   public String replaceFirst(String replacement) {
     reset();
-    StringBuilder sb = new StringBuilder();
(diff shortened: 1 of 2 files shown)


eaftan (approved):
LGTM, thanks @eamonnmcmanus!  I'm going to follow up by adding a benchmark for this case, since it represents a real production workload.

Issue #488, opened 2026-06-20
replaceAll benchmark shows performance gap with RE2


cushon:
Similar to #481, #478 this is an attempt to create a benchmark that reproduces real world performance results for SafeRE comparing to RE2.

The results here showed a ~3x gap between SafeRE and RE2-via-JNI in some `replaceAll` cases.

The analysis points to the issue being the semantic guards causing SafeRE to fall back to the NFA engine. This could be a reason to optimize the NFA engine, or to consider an [intentional divergence](https://github.com/eaftan/safere/blob/main/INTENTIONAL_DIVERGENCES.md) here and use the DFA even if it could result in theoretical divergences.

In the motivating examples here we were also crosschecking against RE2 and the behaviour was the same, so I think this may be a case where the diverges the semantics guards are preventing are theoretical enough that trading performance for compatibility is worth it. I am not certain that is a good tradeoff, I don't have a complete picture yet of how significant the divergences would be, or how much this performance gap matters in practice. It probably makes sense to be a bit cautious about chasing benchmarking wins that trade off compatibility with `java.util.regex`.

The rest of this is the summary of some agentic analysis and the actual benchmark.

---

These results show a performance gap between SafeRE and RE2 (both pure Java RE2J and RE2-via-JNI) when running `replaceAll` on patterns containing alternations or lazy quantifiers.

For example, on an input of **100,000 characters**:

* A pattern with a lazy quantifier and alternation (e.g., `(?i)(?s)\s*\n*(?:<block>|\[block\])\s*\n?renderElement\(.*?\)\s*\n?(?:</block>|\[/block\])`) takes **15ms** in SafeRE compared to **0.8ms** in RE2-via-JNI (**18.8x slower**) and **11.5ms** in RE2J.

* A pattern with alternation and capture groups (e.g., `(^|[^#])(#!customTag)`) takes **29ms** in SafeRE compared to **2ms** in RE2-via-JNI (**14.3x slower**) and **9.5ms** in RE2J.

---

### Root Cause Analysis

1. **DFA Sandwich Bypassed:**
   To guarantee 100% bug-for-bug compatibility with `java.util.regex`'s backtracking semantics, SafeRE implements **Semantic Guards**. When a pattern contains alternations (`hasAlternation`) or lazy quantifiers (`hasLazy`), the DFA start state is marked unreliable (`dfaStartReliable() == false`).
[...]

PR #489, opened 2026-06-20
Make `EnginePathOptions` and `Pattern.compile(..., EnginePathOptions)` public

cushon:
This allows disabling semantic guards as a performance experiment, see https://github.com/eaftan/safere/issues/488

I'm not sure how good an idea this is and am curious what you think of #488 in general, but this would leave the decision up to SafeRE clients instead of making it the default

### DIFF
--- a/safere/src/main/java/org/safere/EnginePathOptions.java
+++ b/safere/src/main/java/org/safere/EnginePathOptions.java
@@ -14,7 +14,7 @@ import java.util.Map;
  * <p>Production code uses {@link #allEnabled()}. Tests in this package can compile a pattern with
  * selected shortcuts disabled to compare observable API traces against the default engine cascade.
  */
-record EnginePathOptions(
+public record EnginePathOptions(
     boolean literalFastPaths,
     boolean charClassMatchFastPaths,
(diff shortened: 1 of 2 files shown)


cushon:
Per the updates in #488 I no longer think disabling semantic guards is that helpful, although this could still be useful as a way to have a public API to configure things like leftmost-longest semantics (#494)

It might be worth considering a smaller public API on Pattern for SafeRE-specific configuration options instead of exposing all of `EnginePathOptions`, though.

cushon:
I think this can be shelved for now.

There may eventually be use-cases for having SafeRE specific options in the API, and it seems maybe better to have that be separate from the `Pattern` constant flags to keep those exactly compatible with the `java.util.regex` API. But I think it'd be better to expose a narrower API than all of `EnginePathOptions`, and only provide options that need to be configured by clients of SafeRE.

The only uses-cases so far are disabling semantic guards (which I realized was misguided, see above) and `LONGEST_MATCH` which is very low priority (see the update in #494)

eaftan:
Sounds good.  I'm going to close this, but feel free to reopen if you change your mind.

PR #491, opened 2026-06-20
Simplify visited table clearing in BitState by removing incremental tracking

cushon:
My benchmarking showed a small (~1.7%) improvement from switching to `Array.fill here`, since the fill is well optimized and avoids book keeping for the dirty word tracking, and this also makes the code a bit simpler.

There's not much in this either way, it'd be worth verifying the results before accepting this.

I spent some time looking for micro-optimizations in BitState and didn't find anything else worthwhile.

### DIFF
--- a/safere/src/main/java/org/safere/BitState.java
+++ b/safere/src/main/java/org/safere/BitState.java
@@ -267,10 +267,7 @@ final class BitState {
 
   private int textSlots;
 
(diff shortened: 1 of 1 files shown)


eaftan (approved):
LGTM. Even if the performance improvement is negligible, this makes the code simpler, so it is a good contribution on its own.

PR #492, opened 2026-06-20
Implement recursive starting character class prefix extraction

cushon:
Previously, extractCharClassPrefixAscii only extracted starting ASCII character classes at the very top level of the AST. It did not see through alternations or case-insensitive literals.

This change extends extractCharClassPrefixAscii to recursively traverse the AST, supporting:

1. Unioning starting character classes for Alternation (RegexpOp.ALTERNATE) nodes.

2. Folding case for case-insensitive literal and literal string nodes, ensuring both upper and lower case ASCII variations are populated.

This allows unanchored searches for patterns like `(?:cat|dog|bird)s?` to leverage prefix acceleration and skip non-matching characters in a fast loop rather than running the full matching engine at every text offset.

Benchmarking show a ~1.46x speedup (46% faster search execution) on alternation-prefix patterns.

### DIFF
--- a/safere/src/main/java/org/safere/Pattern.java
+++ b/safere/src/main/java/org/safere/Pattern.java
@@ -1653,11 +1653,19 @@ public final class Pattern implements Serializable {
    * @return a {@code boolean[128]} ASCII bitmap, or {@code null} if no suitable prefix exists
    */
   private static boolean[] extractCharClassPrefixAscii(Regexp re) {
+    boolean[] bitmap = new boolean[128];
+    if (populateCharClassPrefixAscii(re, bitmap)) {
+      return bitmap;
+    }
+    return null;
+  }
+
+  private static boolean populateCharClassPrefixAscii(Regexp re, boolean[] bitmap) {
     Regexp node = re;
 
     // See through leading captures and concat wrappers.
     while (node != null) {
-      if (node.op == RegexpOp.CAPTURE) {
+      if (node.op == RegexpOp.CAPTURE || node.op == RegexpOp.NON_CAPTURE) {
         node = node.sub();
         continue;
       }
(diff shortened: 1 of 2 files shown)


cushon:
Example:

```java
  @Test
  void benchmarkAlternationPrefix() {
    String pattern = "(?:cat|dog|bird)s?";
    String input = "The quick brown cat and the lazy dog and a bluebird, plus some birds in a tree. ".repeat(100);

    Pattern withAccel = Pattern.compile(pattern);
    Pattern noAccel = Pattern.compile(pattern, 0, 
        EnginePathOptions.builder().startAcceleration(false).build());

    // Warmup
    for (int i = 0; i < 1000; i++) {
      runFind(withAccel, input);
      runFind(noAccel, input);
    }

    // Measure With Start Acceleration
    int countWith = 0;
    for (int i = 0; i < 5000; i++) {
      countWith += runFind(withAccel, input);
    }
    long timeWith = System.nanoTime() - start;

    // Measure Without Start Acceleration (Baseline)
    start = System.nanoTime();
    int countNo = 0;
    for (int i = 0; i < 5000; i++) {
      countNo += runFind(noAccel, input);
    }
    long timeNo = System.nanoTime() - start;

    System.out.printf("WithAccel: %d matches, %d ms\n", countWith, timeWith / 1_000_000);
    System.out.printf("NoAccel: %d matches, %d ms\n", countNo, timeNo / 1_000_000);
    System.out.printf("Speedup: %.2fx\n", (double) timeNo / timeWith);
  }

  private int runFind(Pattern p, String text) {
    Matcher m = p.matcher(text);
    int count = 0;
    while (m.find()) {
      count++;
    }
    return count;
  }
```

*   **Baseline (No start acceleration):** `3728 ms` (DFA evaluated at every index)
*   **Optimized (Recursive prefix acceleration):** `2545 ms` (skipping non-candidate indices)
*   **Speedup:** **1.46x (46% faster)**

eaftan (commented):
We have to avoid recursion in SafeRE because it can introduce stack overflows, which violate SafeRE's stack safety guarantee.  I apologize that this wasn't documented anywhere in the repo or agent instructions.

This example would stack overflow with this PR:

```java
 import org.safere.Pattern;

  public final class PrefixStackOverflowRepro {
      int depth = 2_000;
      String regex = nestedAlternationPattern(depth);

      System.out.println("compiling depth=" + depth + ", regex length=" + regex.length());
      Pattern.compile(regex);
      System.out.println("compiled");
    }

    private static String nestedAlternationPattern(int depth) {
      StringBuilder regex = new StringBuilder(depth * 5 + 1);
      for (int i = 0; i < depth; i++) {
        regex.append("(?:");
      }
      regex.append('a');
      for (int i = 0; i < depth; i++) {
        regex.append("|b)");
      }
      return regex.toString();
    }
  }
```

Run with `java -Xss256k PrefixStackOverflowRepro`.

I've added the no-recursion rule to the agent instructions in https://github.com/eaftan/safere/pull/497.  

I've pushed a change to this PR that converts it to an iterative algorithm.

eaftan (approved):
I was able to reproduce the performance improvement on my benchmark case `RegexBenchmark.alternationFind`.  The performance improvement is even bigger on that workload: 6.36x.

Thank you for this optimization!

Issue #474, opened 2026-06-10
Add lightweight SafeRE diagnostics listener for engine selection


eaftan:
Refs #469

## Background

In #469 we discussed whether optimizing Pike VM allocation is worth prioritizing. A useful next step is to add lightweight diagnostics for which engine path SafeRE actually uses in real workloads, so we can tell whether Pike NFA execution is common enough to justify allocation-focused work.

The first version should keep the public surface small. It should expose stable strategy-level information, not internal branch details that may change as the engine cascade evolves.

## Proposed API

Add a listener-style diagnostics hook:

```java
public interface SafeReDiagnostics {
  SafeReDiagnostics NONE = new SafeReDiagnostics() {};

  default void onPatternCompiled(PatternDiagnostics event) {}

  default void onMatchCompleted(MatchDiagnostics event) {}
}
```

Register it globally on `Pattern`:

```java
Pattern.setDiagnostics(SafeReDiagnostics diagnostics);
Pattern.diagnostics();
```

Implementation should store the listener in a `volatile` static field and use `SafeReDiagnostics.NONE` when diagnostics are disabled. Event sites should check for `NONE` before allocating event records.

## Pattern Diagnostics

Emit once when a pattern is compiled, after metadata is known.

Sketch:

```java
public record PatternDiagnostics(
    PatternFeatureSet features,
    int programSize,
    int captureCount,
    boolean onePassEligible,
    boolean dfaEligible,
    boolean bitStateEligible) {}
```

Initial feature set:

```java
public enum PatternFeature {
  LITERAL,
  CAPTURES,
  ALTERNATION,
  LAZY_QUANTIFIER,
  BOUNDED_REPEAT,
  ANCHOR,
  WORD_BOUNDARY,
  GRAPHEME,
  CASE_INSENSITIVE,
  UNICODE_CHARACTER_CLASS
}
```

Do not include the pattern text by default. Users can maintain their own association externally if they need per-pattern attribution.

## Match Diagnostics

Emit once per public match operation completion: `matches()`, `lookingAt()`, and `find()`.

Sketch:

```java
public record MatchDiagnostics(
    MatchOperation operation,
    MatchOutcome outcome,
    RegexEngine primaryEngine,
    CaptureMode captureMode,
    int inputLength) {}
```

Enums:

```java
public enum MatchOperation {
  MATCHES,
  LOOKING_AT,
  FIND
}

public enum MatchOutcome {
  MATCH,
  NO_MATCH
}

[...]

PR #466, opened 2026-06-05
Optimize SafeRE performance by running the DFA for word boundary patterns

cushon:
- Restore the DFA fast-path for patterns containing word boundary assertions (\b and \B), eliminating the NFA/BitState fallback. This fixes a ~10x performance regression for affected patterns.
- Align NFA search stepping for word boundary patterns to code point boundaries (switching from doSearchEveryCharPosition to doSearch). This ensures both the NFA and DFA engines are consistent and stop evaluating assertions inside surrogate pairs, avoiding JDK-style character-splitting and string corruption.
- Update UnicodeMatchBoundsTest trace expectations to align with the new, correct code-point boundary matching behavior.

https://github.com/eaftan/safere/issues/465

PR #493, opened 2026-06-20
Specialize EmptyOp anchor and boundary checking in BitState

cushon:
Previously, when processing an OP_EMPTY_WIDTH instruction (such as \b, ^, $, \A, or \z), BitState called Nfa.emptyFlags(...). This method computed the full bitmask of all possible empty-width assertions at that position, including expensive operations like line terminators, unicode word character lookups, and grapheme boundaries.

Since each OP_EMPTY_WIDTH instruction represents exactly one target constraint (e.g. only word boundaries or multiline line-start), we specialize the check inside the backtracking loop. We now evaluate only the specific required constraint directly, avoiding unnecessary character lookups, branch checks, and grapheme evaluations.

Benchmarks show a ~1.61x speedup (61% faster matching) on boundary/anchor heavy patterns.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,10 +73,10 @@ jobs:
           cache: maven
 
       - name: Smoke test Java benchmarks
-        run: ./run-java-benchmarks.sh --smoke RegexBenchmark
+        run: ./run-java-benchmarks.sh --smoke '^org\.safere\.benchmark\.RegexBenchmark\.'
 
       - name: Smoke test Java memory benchmarks (GC profiler)
-        run: ./run-java-memory-benchmarks.sh --smoke RegexBenchmark
+        run: ./run-java-memory-benchmarks.sh --smoke '^org\.safere\.benchmark\.RegexBenchmark\.'
 
       - name: Smoke test Java memory benchmarks (pattern sizes)
         run: |
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -46,7 +46,7 @@ mvn -pl safere test -q
 mvn install -DskipTests -q
 
 # Run benchmarks (see Benchmarking section below)
-./run-java-benchmarks.sh RegexBenchmark
+./run-java-benchmarks.sh '^org\.safere\.benchmark\.RegexBenchmark\.'
 ```
 
 ## Code Style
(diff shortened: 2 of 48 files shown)


cushon:
Example:

```java
  @Test
  void benchmarkWordBoundary() {
    String pattern = "\\b(?:cat|dog|bird)\\b";
    String input = "The quick brown cat and the lazy dog and a bluebird, plus some birds in a tree. ".repeat(100);

    Pattern p = Pattern.compile(pattern);

    // Warmup
    for (int i = 0; i < 1000; i++) {
      runFind(p, input);
    }

    // Measure
    int count = 0;
    for (int i = 0; i < 10000; i++) {
      count += runFind(p, input);
    }
    long time = System.nanoTime() - start;

    System.out.printf("Matches: %d, Time: %d ms\n", count, time / 1_000_000);
  }

  private int runFind(Pattern p, String text) {
    Matcher m = p.matcher(text);
    int count = 0;
    while (m.find()) {
      count++;
    }
    return count;
  }
```

*   **Baseline (Evaluating full emptyFlags mask):** `33380 ms`
*   **Optimized (Specialized switch block for WORD_BOUNDARY checks):** `20665 ms`
*   **Speedup:** **1.61x (61% faster)**

eaftan (commented):
I can't reproduce the performance improvement using your benchmark from the comment.  Here are the numbers I'm seeing on my machine:

main:
Trial 1: 10500 ms
Trial 2: 10453 ms
Trial 3: 10642 ms
Trial 4: 10510 ms
Trial 5: 10512 ms
average: ~10523 ms

this PR branch:
Trial 1: 10505 ms
Trial 2: 10529 ms
Trial 3: 10556 ms
Trial 4: 10558 ms
Trial 5: 10670 ms
average: ~10564 ms

I wonder if your result comes from an older version of SafeRE?  The current normal `Matcher.find()` path disables BitState on patterns with word boundary assertions here: https://github.com/eaftan/safere/blob/470f96bdfe39626761b565f785e484161c7a1c43/safere/src/main/java/org/safere/Matcher.java#L1863

This came out of a correctness sweep in PR #463, and then there was a round of optimization in PR #466 that restored DFA support for word-boundary patterns. 

The reason I disabled BitState for patterns with word-boundary assertions is that matching `java.util.regex` behavior requires considering UTF-16 character offsets, including offsets inside supplementary code points.  BitState’s search stepping is code-point based, so it can skip observable `\b`/`\B` positions. It’s possible to fix that, but it didn’t seem worthwhile at the time because the Pike NFA path was the canonical correctness path, and PR #466 restored the DFA fast path for these patterns.

cushon:
Sorry, I think what happened here is that the original benchmark results for word boundaries were obtained by forcing the use of BitState.

I think there are other cases where empty-width assertions (like ^, $, \A, \z) are evaluated by BitState and this optimization would help those cases.

I found a real world example `(\s*\[\d+\.\d+\.\d+(?:,\s*\d+\.\d+\.\d+)*\])+$` which uses BitState and shows a 2-3% speedup with this change. A synthetic example like `(([a-z])?$)+` has a larger ~14.6% speedup.

In the case of the real world example it's using BitState because of the outer capturing group, which is actually unused, the pattern is only used with `replaceAll("")`. The best improvement in that case would be to remove the unused capturing group so it could be executed with the DFA (which would be >10x faster). So that might be motivation for static or dynamic analysis to find inefficient patterns and statically rewriting them (some overlap with #474).

Overall I think there is a small benefit to be had here but maybe not enough to justify the additional logic, and not relying onBitState is probably a better approach.

cushon:
> The current normal `Matcher.find()` path disables BitState on patterns with word boundary assertions here:
> 
> https://github.com/eaftan/safere/blob/470f96bdfe39626761b565f785e484161c7a1c43/safere/src/main/java/org/safere/Matcher.java#L1863
> 
> This came out of a correctness sweep in PR #463, and then there was a round of optimization in PR #466 that restored DFA support for word-boundary patterns.
> 
> The reason I disabled BitState for patterns with word-boundary assertions is that matching `java.util.regex` behavior requires considering UTF-16 character offsets, including offsets inside supplementary code points. BitState’s search stepping is code-point based, so it can skip observable `\b`/`\B` positions. It’s possible to fix that, but it didn’t seem worthwhile at the time because the Pike NFA path was the canonical correctness path, and PR #466 restored the DFA fast path for these patterns.

Thanks, considering this again, #466 already made the decision to intentionally diverge from `java.util.regex` on not allowing word boundaries to split surrogate pairs, and so I think it makes sense to also remove the `hasWordBoundary` check here and falling back to BitState on word boundaries, right?

I have updated this PR to include the change to remove `hasWordBoundary` and add slightly more test coverage.

cushon:
I tried benchmarking the two parts of this independently, for `unprefixedWordBoundary` removing `!prog.hasWordBoundary()` is a 2.7x speedup, and the `matchEmptyWidth` stuff was another 5-8%.

So most of the benefit here is just removing `!prog.hasWordBoundary()` and that part is also simpler, it might make sense to just take that part and defer `matchEmptyWidth` unless there's more evidence that optimization is worth the extra code.

eaftan (approved):
Oh right, I forgot that #466 included that.  Sorry!

I took your advice and backed out everything except removing `!prog.hasWordBoundary()`, then ran the `unprefixedWordBoundary` benchmark you mentioned.  I was able to repro your speedups.

| Case | main | simplified branch | Speedup |
|---|---:|---:|---:|
| inputSize=1000, match=true | 762.457 ns/op | 299.139 ns/op | 2.55x |
| inputSize=1000, match=false | 2899.972 ns/op | 2910.212 ns/op | 1.00x |
| inputSize=10000, match=true | 741.844 ns/op | 348.225 ns/op | 2.13x |
| inputSize=10000, match=false | 29348.604 ns/op | 28251.776 ns/op | 1.04x |

So LGTM, thanks!

PR #494, opened 2026-06-20
Add leftmost-longest matching support to SafeRE

cushon:
Introduces the `longestMatch` option to `EnginePathOptions`, enabling leftmost-longest matching behavior compatible with RE2 semantics.

I started looking at this as part of #488 as a workaround for divergences with RE2 with semantic guards disabled, which was misguided because it can work around specific cases but in general doesn't ensure RE2 semantics with SafeRE's semantics guards disabled.

Longest match support can be enabled in RE2 and RE2J API, so supporting this could make it easier to migrate from RE2 engines to SafeRE. It doesn't seem that invasive to support:

https://github.com/google/re2j/blob/951a6159fdc0a4931593c70442926f86031edf40/java/com/google/re2j/Pattern.java#L51

---

Changes:
- Added `longestMatch` option to `EnginePathOptions` builder and accessors.
- Wired the `longestMatch` flag into `Matcher.java` to configure matching behavior across DFA, BitState, and NFA engines.
- Anchored forward DFA passes in `Matcher` now run in longest-match mode when enabled, allowing the DFA sandwich to resolve the longest match starting from a leftmost index.
- If `longestMatch` is enabled, the match boundaries resolved by the DFA sandwich are considered reliable even when `semanticGuards` is active, bypassing the fallback engine.
- Sorted keyword arrays in keyword-alternation fast path in descending length order when `longestMatch` is enabled to ensure the longest prefix is matched first.
- Added a new `LongestMatchTest.java` suite to verify leftmost-longest match correctness and performance benefits when semantic guards are disabled on simple alternation patterns.

### DIFF
--- a/safere-crosscheck/pom.xml
+++ b/safere-crosscheck/pom.xml
@@ -125,6 +125,7 @@
                         <exclude name="MatcherDeferredCaptureStateTest.java"/>
                         <exclude name="MatcherTransitionInventoryTest.java"/>
                         <exclude name="NfaTest.java"/>
+                        <exclude name="LongestMatchTest.java"/>
                         <exclude name="OnePassTest.java"/>
                         <exclude name="ParseFlagsTest.java"/>
                         <exclude name="ParserTest.java"/>
--- a/safere/src/main/java/org/safere/Dfa.java
+++ b/safere/src/main/java/org/safere/Dfa.java
@@ -880,7 +880,7 @@ final class Dfa {
               isMatchValid = true;
             }
           }
-          if (isMatchValid) {
+          if (isMatchValid && !longest) {
             break; // Prune all lower-priority branches!
           }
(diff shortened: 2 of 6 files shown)


cushon:
I analyzed usages of longest-match semantics in google’s codebase. There are <10 usages of RE2J’s `LONGEST_MATCH`, and the JNI wrapper around RE2 doesn’t expose the option at all. Of the RE2J usages some were using `matches()` and so `LONGEST_MATCH` wasn’t doing anything, and another could easily be rewritten to swap the order of alternatives to put the longest first.

So I think this is probably a very low priority for now, and could be revisited if there’s demand for it in the future.

eaftan:
Thanks Liam. I’m also going to close this, but feel free to reopen if you want to revisit.

PR #495, opened 2026-06-20
Optimize compiler to detect safe alternations and bypass NFA fallbacks

cushon:
This change implements disjoint-set lookahead analysis in the pattern compiler to identify structural alternations (e.g. `(^|[^<])`) that are guaranteed to have mutually exclusive starting characters or boundaries.

Because safe alternations can never produce leftmost-first vs. leftmost-longest conflicts, the pattern compiler optimizes away the semantic guard and promotes execution from NFA fallback to the pure DFA engine path.

Helps with:
- ReplaceAll benchmarks containing safe boundary prefix alternations (specifically Issue #488 replaceAll benchmarks like `opentag`), resulting in a 14.7x speedup (runtimes drop from 11.1 ms to 770 us on 100KB inputs). This makes SafeRE 3.1x faster than native JNI RE2.


eaftan:
Update from this review pass:

1. Correctness fix

The unsafe-alternation scan is now stack-safe and still checks nested alternations under safe outer alternations. The previous recursive AST walk could overflow during `Pattern.compile` on deeply nested generated regexes, and an intermediate iterative version stopped at a safe outer alternation. The current version uses an explicit worklist carrying the lookahead context, and adds regressions for both cases:

- deeply nested quantified alternation compiles without depending on the Java call stack
- `(?:a(?:b|bc)|z)` still marks DFA start as unreliable because the nested `b|bc` alternation is unsafe

2. Benchmark evidence

I used the `Issue488ReplaceAllBenchmark` added on PR #496 as a temporary benchmark only; it was not added permanently to this PR. I applied that benchmark to detached worktrees for `main` and this PR branch, then ran:

```bash
./run-java-benchmarks.sh Issue488ReplaceAllBenchmark
```

For the SafeRE `altCapture` workload (`(^|[^#])(#!customTag)` with `replaceAll`), results were:

| text size | main SafeRE | this PR SafeRE | speedup |
|---:|---:|---:|---:|
| 1,000 | 44.069 us/op | 4.200 us/op | 10.49x |
| 10,000 | 447.534 us/op | 38.340 us/op | 11.67x |
| 100,000 | 18,722.723 us/op | 568.161 us/op | 32.95x |

So for the issue #488 safe boundary-prefix alternation case, this PR is a substantial SafeRE-only improvement, especially at 100 KB input size.

Issue #481, opened 2026-06-19
Additional benchmarks comparing java.util.regex / RE2 / RE2J / SafeRE


cushon:
I tried to extract a benchmark showing some regexes and inputs that are representative of real-world examples.

Similar to https://github.com/eaftan/safere/pull/478 the goal here was to investigate some cases where there was a gap between SafeRE and using RE2-via-JNI. In practice we're seeing a larger gap on average between SafeRE and RE2-JNI than [RE2-FFM results in the SafeRE benchmarks](https://github.com/eaftan/safere/blob/main/BENCHMARKS.md#vs-re2-ffm-c-re2-via-ffm).

I did some analysis of these benchmark results. I found a couple small optimization opportunities I'll open a PR for.

When I had an agent look at it, it claimed that the remaining gap is mostly due to fundamental limitations of the approach since `java.util.regex` can scan once and SafeRE has to do the 3-Pass DFA Sandwich, and also compared to RE2-via-JNI RE2 is vectorized and can optimize prefix scans better than SafeRE. If that's true it might be fun to think about vectorizing SafeRE with the [Vector API](https://openjdk.org/jeps/537) especially if/when it's out of preview.

I also don't have a full sense of whether the performance gaps here would matter for real-world usage of SafeRE, mostly I want to understand to what extent these gaps could be optimized and to what extent they're necessary with the approach SafeRE takes.

---

```java
import java.util.concurrent.TimeUnit;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;

/** JMH benchmark comparing the performance of SafeRE, RE2J, JNI RE2, and JDK regex. */
@State(Scope.Benchmark)
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@Warmup(iterations = 3, time = 1, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
@Fork(1)
public class RegexEngineBenchmark {

[...]

PR #496, opened 2026-06-20
Optimize NFA VM by reducing GC allocations in Pike VM execution

cushon:
This change replaces the stack-based DFS and heap-allocated capture arrays in the NFA execution path with recursive backtracking utilizing pre-allocated, recycled thread and queue arrays.

This completely eliminates JVM garbage collection allocations during character stepping inside the NFA VM.

Helps with:
- Scan-heavy forced-NFA workloads (e.g. Issue #481 scaling benchmarks) where SafeRE Pike VM scanning speed improves by 2.3x - 2.5x.
- Splitting benchmarks (e.g. splitWords) where throughput improves by 1.6x.
- General NFA fallback operations (e.g. replaceAll with lazy quantifiers in Issue #488) where throughput improves by 2.2x.

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -83,6 +83,24 @@
     }
   },
 
+  "issue488ReplaceAll": {
+    "textSizes": [1000, 10000, 100000],
+    "lazyAlt": {
+      "pattern": "(?i)(?s)\\s*\\n*(?:<block>|\\[block\\])\\s*\\n?renderElement\\(.*?\\)\\s*\\n?(?:</block>|\\[/block\\])",
+      "replacement": "",
+      "prefixUnit": "Some prefix text preceding the render block element. ",
+      "match": "\n<block>\nrenderElement(id=\"el-42\", width=100, height=200, label=\"hello world\")\n</block>\n",
+      "suffixUnit": "Some suffix text following the render block element. "
+    },
+    "altCapture": {
+      "pattern": "(^|[^#])(#!customTag)",
+      "replacement": "$1<tag type=\"custom\"",
+      "hitUnit": "This is some sample prose text that contains a tag marker #!customTag and more text here. ",
+      "missUnit": "This is some sample prose text that contains a tag marker #customTag and more text here. ",
+      "hitInterval": 5
+    }
+  },
+
   "captureScaling": {
     "capture0": {
       "pattern": "[0-9]+-[0-9]+-[0-9]+",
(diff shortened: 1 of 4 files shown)


eaftan:
Posted a follow-up commit on the PR branch after merging current `main`.

1. Correctness fix

The NFA epsilon traversal is now iterative instead of recursive. This keeps deeply nested empty/capture-transition chains from depending on the Java call stack, preserving the stack-safety guarantee for the Pike NFA path. I also added a regression test with 30,000 nested capturing groups running on a small stack.

2. Added benchmark

Added `Issue488ReplaceAllBenchmark` plus shared `benchmark-data.json` workloads for the issue 488 replace-all cases:

- `lazyAlt`: the block/renderElement lazy alternation pattern
- `altCapture`: the `(^|[^#])(#!customTag)` capture replacement pattern

3. Benchmark results

I compared current `main` against this PR using the standard Java benchmark script, applying only the new benchmark harness to the `main` worktree so both branches could run the same workloads:

```bash
./run-java-benchmarks.sh 'Issue481ScalingBenchmark.*_safere' 'Issue488ReplaceAllBenchmark.*_safere'
```

Lower is better. Speedup is `main / PR`.

Issue 488 results:

| Workload | Size | main us/op | PR us/op | Result |
|---|---:|---:|---:|---:|
| `altCapture` | 1,000 | 46.850 | 44.224 | 1.06x faster |
| `altCapture` | 10,000 | 467.142 | 455.606 | 1.03x faster |
| `altCapture` | 100,000 | 21084.458 | 8674.722 | 2.43x faster |
| `lazyAlt` | 1,000 | 28.324 | 27.981 | 1.01x faster |
| `lazyAlt` | 10,000 | 1103.770 | 552.113 | 2.00x faster |
| `lazyAlt` | 100,000 | 10398.172 | 5354.069 | 1.94x faster |

Issue 488 geomean: PR is 1.48x faster.

Issue 481 also showed targeted wins in the scheme-prefix workloads:

| Workload | Size | main us/op | PR us/op | Result |
|---|---:|---:|---:|---:|
| `schemeExtract` | 10,240 | 25.593 | 12.709 | 2.01x faster |
| `schemeExtract` | 102,400 | 32.628 | 20.262 | 1.61x faster |
| `schemeFindNegative` | 10,240 | 12.414 | 6.432 | 1.93x faster |
| `schemeFindNegative` | 102,400 | 20.006 | 13.849 | 1.44x faster |

Across all `Issue481ScalingBenchmark.*_safere` rows, the geomean was 1.07x faster. The block and split-word rows were effectively flat to slightly slower, mostly low single-digit differences and within the observed benchmark error/noise range.

Validation run before pushing:

```bash
git diff --check
mvn -pl safere -Dtest=NfaTest test -q
```

eaftan:
LGTM, thanks!

PR #484, opened 2026-06-19
Add issue 481 scaling benchmarks

eaftan:
## Summary

Adds issue 481 benchmark coverage for representative application-style regex workloads:

- `\W+` split over prose-like text
- `(?s)<block>.*?</block>` block extraction
- `<<tag.*?>>` tag extraction
- custom-scheme Markdown-style link extraction with capture retrieval

The Java benchmark scales each workload across `128 B`, `1 KB`, `10 KB`, `100 KB`, and `1 MB`. The shared `benchmark-data.json` inputs are also wired into the C++ RE2 and Go regexp harnesses. The Java RE2-FFM wrapper now has `split(...)` support so the Java benchmark calls pattern APIs consistently.

Fixes #481

## Quick Benchmark Signal

Ran:

```bash
./run-java-benchmarks.sh --quick Issue481ScalingBenchmark
```

Runtime: 20m03s. These are quick development results, not publication-quality `BENCHMARKS.md` results.

### Split Words: `\W+`

| Text Size | SafeRE | JDK | RE2/J | RE2-FFM | vs JDK | vs RE2/J | vs RE2-FFM |
|--:|--:|--:|--:|--:|---|---|---|
| 128 B | 2.91 | 0.63 | 5.23 | 4.80 | 4.6x slower | **1.8x faster** | **1.7x faster** |
| 1 KB | 20.6 | 4.89 | 40.5 | 32.8 | 4.2x slower | **2.0x faster** | **1.6x faster** |
| 10 KB | 215 | 49.0 | 427 | 464 | 4.4x slower | **2.0x faster** | **2.2x faster** |
| 100 KB | 2,153 | 489 | 4,250 | 30,163 | 4.4x slower | **2.0x faster** | **14.0x faster** |
| 1 MB | 26,075 | 5,532 | 43,933 | 5,472,706 | 4.7x slower | **1.7x faster** | **209.9x faster** |

### Block Find: `(?s)<block>.*?</block>`

| Text Size | SafeRE | JDK | RE2/J | RE2-FFM | vs JDK | vs RE2/J | vs RE2-FFM |
|--:|--:|--:|--:|--:|---|---|---|
| 128 B | 1.21 | 0.43 | 4.42 | 0.45 | 2.8x slower | **3.7x faster** | 2.7x slower |
| 1 KB | 5.21 | 4.24 | 43.3 | 2.59 | 1.2x slower | **8.3x faster** | 2.0x slower |
[...]

PR #498, opened 2026-06-21
Flatten DFA transition table matrix to contiguous array layout

cushon:
Flatten DFA transition table from jagged int[][] to a contiguous one dimensional array. Restructures match loops to use primitive-only state/offset lookups. This improves cache locality, avoids pointer indirection, and reduces heap fragmentation.

This is a bit of a complexity / readability tradeoff, but it seems like it does benchmark faster.

Benchmark results (100 KB inputs):

* For pure DFA execution (e.g., blockFind using `(?s)<block>.*?</block>`):
  - Baseline: 1170.679 us/op
  - Flattened: 959.720 us/op (~22% speedup)
* For search patterns with prefix optimizations (e.g., tagFind), the vectorized literal search dominates, showing no significant change.

### DIFF
--- a/safere/src/main/java/org/safere/Dfa.java
+++ b/safere/src/main/java/org/safere/Dfa.java
@@ -90,18 +90,16 @@ final class Dfa {
      */
     final int[] wordBoundaryMatchIds;
 
-    /** Transitions indexed by equivalence class; null entry = not yet computed. */
-    final State[] next;
+    int id = -1; // Added: unique index in Dfa states array
 
-    State(int[] insts, int flags, int numClasses) {
-      this(insts, flags, null, numClasses);
+    State(int[] insts, int flags) {
+      this(insts, flags, null);
     }
 
-    State(int[] insts, int flags, int[] wordBoundaryMatchIds, int numClasses) {
+    State(int[] insts, int flags, int[] wordBoundaryMatchIds) {
       this.insts = insts;
       this.flags = flags;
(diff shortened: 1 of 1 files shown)


eaftan:
Hmm, I can't repro the performance improvement from this PR.  The `blockFind` workload maps to the benchmark added in #484.  When I run:

```bash
./run-java-benchmarks.sh 'Issue481ScalingBenchmark\.blockFind_safere'
```

I get:

| textSize (bytes) |  main d11ae6a us/op   |     dfa-flattening us/op   |    ratio |
|---------|--------------------------|------------------------------|--------|
|  128    |    1.250 +/- 0.051      |     1.241 +/- 0.080        |    0.99x |
|  1024   |    5.386 +/- 0.100      |     4.981 +/- 0.147        |    0.92x |
|  10240   |   53.875 +/- 1.390    |      49.021 +/- 0.692     |      0.91x |
|  102400  |   567.933 +/- 27.237  |      754.496 +/- 12.872 |        1.33x |
|  1048576  |  5754.054 +/- 284.349  |    7396.326 +/- 278.465  |     1.29x |

Ratio is `dfa-flattening / main`, so values above 1.0 mean the PR branch is slower.

I wonder if some of the other optimizations we've merged into main have changed the effectiveness of this optimization?

cushon:
Sorry about this, the earlier results appear to have been the result of bad benchmarking methodology, where there was insufficient warmup. With sufficient warmup I'm seeing better performance from the unflattened layout.

PR #499, opened 2026-06-21
Statically link libstdc++ and libgcc into FFM RE2 shim

cushon:
Statically link the C++ standard library and GCC runtime library into the compiled libre2_shim.so binary.

This prevents UnsatisfiedLinkErrors in environments where the running Java VM's dynamic library loader and the compiled native library target mismatched glibc/libstdc++ versions.  Static linking increases the portability of the FFM benchmark wrapper across different Linux installations.

### DIFF
--- a/safere-ffm-re2/src/main/native/CMakeLists.txt
+++ b/safere-ffm-re2/src/main/native/CMakeLists.txt
@@ -33,3 +33,4 @@ FetchContent_MakeAvailable(abseil-cpp re2)
 # Build our C shim as a shared library, statically linking RE2 into it
 add_library(re2_shim SHARED re2_shim.cpp)
 target_link_libraries(re2_shim PRIVATE re2::re2)
+target_link_options(re2_shim PRIVATE "-static-libstdc++" "-static-libgcc")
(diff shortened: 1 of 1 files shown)


eaftan (approved):
I made a change that guards this to Linux with GCC or Clang.  

My agent says that MacOS doesn't support static linking against Apple system libraries, including `libSystem`.  That was surprising but seems to be supported by some of Apple's docs:  https://developer.apple.com/forums/thread/715385#:~:text=Other%20Unix%2Dy%20platforms%20support%20the%20notion%20of%20a%20statically%20linked%20executable%2C%20one%20that%20makes%20system%20calls%20directly.%20This%20is%20not%20supported%20on%20Apple%20platforms.%20Apple%20platforms%20provide%20binary%20compatibility%20via%20system%20dynamic%20libraries%20and%20frameworks%2C%20not%20at%20the%20system%20call%20level.

> Other Unix-y platforms support the notion of a statically linked executable, one that makes system calls directly. This is not supported on Apple platforms. Apple platforms provide binary compatibility via system dynamic libraries and frameworks, not at the system call level.

PR #501, opened 2026-06-22
Optimize DFA cache lookups to be allocation-free on cache hits

cushon:
This PR optimizes the DFA state transition caching logic to avoid object allocations during cache hits. 

This doesn't improve benchmark performance after warmup, the hope is that it might improve performance while the DFA is warming up, and it also reduces allocations.

---

Previously, `Dfa.getOrCreate()` allocated a new `StateKey` record on every transition lookup. In hot matching loops where transition lookups frequently hit the cache, this generated substantial GC allocation pressure.

This change:
1. Converts the internal `StateKey` from a record to a mutable class with `reset()` and `copy()` methods.
2. Adds a private instance-local `lookupKey` to the `Dfa` class.
3. Rewrites `getOrCreate()` to reuse `lookupKey.reset(insts, flags)` for the cache query. We only allocate a persistent `StateKey` object on a cache miss by calling `lookupKey.copy()`.

Because `Dfa` instances are cached inside a `ThreadLocal` wrapper, reusing an instance-level key is completely thread-safe and requires no synchronization.

### DIFF
--- a/safere/src/main/java/org/safere/Dfa.java
+++ b/safere/src/main/java/org/safere/Dfa.java
@@ -110,9 +110,26 @@ final class Dfa {
   }
 
   /** Cache key for state deduplication. */
-  // TODO(#98): Replace int[] with Guava ImmutableIntArray to get proper value semantics.
-  @SuppressWarnings("ArrayRecordComponent")
-  private record StateKey(int[] insts, int flags) {
+  private static final class StateKey {
+    private int[] insts;
+    private int flags;
+
+    StateKey() {}
+
+    StateKey(int[] insts, int flags) {
+      this.insts = insts;
+      this.flags = flags;
+    }
+
+    void reset(int[] insts, int flags) {
+      this.insts = insts;
+      this.flags = flags;
+    }
+
+    StateKey copy() {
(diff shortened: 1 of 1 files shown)


cushon:
This avoids some garbage creating the DFA cache but don't change anything once the cache is created, and looking at this again the benefits seem negligible. I think we can shelve this for now. The memory savings from #501 seem more worthwhile.

cushon:
Considering this more, I think it's worth trying.

It should avoid allocations on the hot path through the DFA for lookups, and could help with tail latency for regex matches, since an allocation on a cache lookup could cause a GC pause. Some of the results suggest SafeRE has worse tail latency than RE2.

This doesn't result in a benchmark win, but the benchmarks runs aren't really GC or memory constrained and are doing a lot of repetitions to smooth out the results.

And this is a pretty small change that doesn't seem like it could make anything worse, it's just re-using a `StateKey` instance of lookups and copying it if necessary, instead of unconditionally creating a new instance.

eaftan (approved):
I agree with your reasoning -- this is a small change, doesn't seem like it could make anything worse, and could help memory usage.

PR #500, opened 2026-06-22
Add real-world comparative regex scaling microbenchmarks

cushon:
Introduce `RealWorldRegexBenchmark` to compare matching and replacement performance of `SafeRE`, `JDK` (`java.util.regex`), `RE2J`, and native `RE2_FFM` on representative regex patterns and input sizes.

The benchmark defines 10 common real-world regex patterns (e.g. map field path lookups, markup entity replacements, wildcards, metadata blocks, tags) and tests them on both matching and non-matching inputs scaling from 1,000 to 10,000 characters.

PR #502, opened 2026-06-22
Implement sparse state transitions table optimization for DFA

cushon:
This PR implements a sparse transition table representation for DFA states to reduce their memory footprint.

Previously, every `State` allocated a flat `State[] next` array of size `numClasses` (the number of equivalence classes). For large programs, this flat array dominated memory usage, even though the vast majority of transitions were never computed.

This change introduces a hybrid transition table:
* **Sparse Mode ($\le 8$ transitions)**: The state uses small, packed primitive arrays (`sparseClasses` and `sparseStates`). Lookups are performed via a linear scan.
* **Inflated Mode ($> 8$ transitions)**: If a state accumulates more than 8 transitions, it inflates to a full `inflatedNext` array of size `numClasses`, copying over the existing sparse transitions and reclaiming the sparse arrays.

This hybrid approach keeps the state memory representation extremely compact for typical patterns while dynamically scaling for high-density transition states.

The impact on the benchmarks from https://github.com/eaftan/safere/pull/500 is

* **Memory Footprint**: Reduces the memory footprint per DFA state instance significantly, saving up to **1.6 KB per state** on typical patterns. This reduces heap usage and GC collection frequency.
* **Execution Performance**: Achieves an **overall geometric mean speedup of 1.07x** (7% faster) across the microbenchmark suite.
* **Trade-offs**: There is a minor regression (8–12%) on some matching loops due to the linear scan overhead when looking up sparse transitions, but this is outweighed by the overall memory savings and reduced GC overhead.

### DIFF
--- a/safere/src/main/java/org/safere/Dfa.java
+++ b/safere/src/main/java/org/safere/Dfa.java
@@ -198,7 +198,6 @@ final class Dfa {
   /** State cache: maps instruction-set + flags to canonical State instance. */
   private final Map<StateKey, State> cache = new HashMap<>();
 
-
   /** Sentinel dead state: no instructions, no transitions possible. */
   private final State deadState = new State(new int[0], 0);
 
--- a/safere/src/main/java/org/safere/Dfa.java
+++ b/safere/src/main/java/org/safere/Dfa.java
@@ -90,22 +90,65 @@ final class Dfa {
      */
     final int[] wordBoundaryMatchIds;
 
-    /** Transitions indexed by equivalence class; null entry = not yet computed. */
-    final State[] next;
+    // Sparse transitions representation (for <= 8 transitions)
+    private int[] sparseClasses;
+    private State[] sparseStates;
+    private byte sparseCount;
 
-    State(int[] insts, int flags, int numClasses) {
-      this(insts, flags, null, numClasses);
+    // Inflated transitions representation (for > 8 transitions)
+    private State[] inflatedNext;
+
+    State(int[] insts, int flags) {
+      this(insts, flags, null);
     }
 
-    State(int[] insts, int flags, int[] wordBoundaryMatchIds, int numClasses) {
+    State(int[] insts, int flags, int[] wordBoundaryMatchIds) {
       this.insts = insts;
       this.flags = flags;
       this.wordBoundaryMatchIds = wordBoundaryMatchIds;
-      this.next = new State[numClasses];
     }
 
     boolean isMatch() {
       return (flags & FLAG_MATCH) != 0;
+    }
+
+    State getTransition(int cls) {
+      if (inflatedNext != null) {
+        return inflatedNext[cls];
+      }
+      for (int i = 0; i < sparseCount; i++) {
+        if (sparseClasses[i] == cls) {
+          return sparseStates[i];
+        }
+      }
+      return null;
+    }
+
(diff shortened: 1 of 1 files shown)


cushon:
I did some more measurement of this, and the claim is that this is a significant savings to the retained heap size of the DFA cache, and that it should be CPU neutral. So it doesn't make the benchmarks faster, but seems worth it for the memory savings.

eaftan:
I took a closer look at the memory/performance tradeoff for this PR.  The memory improvement you cited matches what I see, but I am seeing some significant performance regressions on some benchmarks.  In particular: 

| benchmark | size | main | PR | change |
|---|---:|---:|---:|---:|
| `RegexBenchmark.emailFind_safere` | 44 bytes | 280.460 ns/op | 365.457 ns/op | 30.3% slower |
| `searchEasyFail_safere` | 1,024 | 0.495 us/op | 0.563 us/op | 13.7% slower |
| `searchEasyFail_safere` | 10,240 | 4.895 us/op | 5.251 us/op | 7.3% slower |
| `searchEasyFail_safere` | 102,400 | 235.635 us/op | 244.942 us/op | 3.9% slower |
| `searchEasyFail_safere` | 1,048,576 | 2633.841 us/op | 2530.907 us/op | 3.9% faster |
| `searchHardFail_safere` | 1,024 | 0.496 us/op | 0.559 us/op | 12.7% slower |
| `searchHardFail_safere` | 10,240 | 4.887 us/op | 5.283 us/op | 8.1% slower |
| `searchHardFail_safere` | 102,400 | 235.123 us/op | 231.897 us/op | 1.4% faster |
| `searchHardFail_safere` | 1,048,576 | 2636.574 us/op | 2523.348 us/op | 4.3% faster |
| `searchMediumFail_safere` | 1,024 | 0.496 us/op | 0.561 us/op | 13.1% slower |
| `searchMediumFail_safere` | 10,240 | 4.881 us/op | 5.241 us/op | 7.4% slower |
| `searchMediumFail_safere` | 102,400 | 232.151 us/op | 238.156 us/op | 2.6% slower |
| `searchMediumFail_safere` | 1,048,576 | 2637.097 us/op | 2531.506 us/op | 4.0% faster |

My read of that is that it seems like this might regress DFA matching on short inputs.  Tomorrow, I'll explore whether we can keep the memory win without regressing short-input performance.

eaftan:
For the record, here are the memory numbers I measured:

| workload | main bytes | PR bytes | saved | reduction |
|---|---:|---:|---:|---:|
| schemeExtractLarge | 8,320 | 2,192 | 6,128 | 73.7% |
| urlExtraction | 3,520 | 1,152 | 2,368 | 67.3% |
| stackTraceExtract | 3,712 | 1,696 | 2,016 | 54.3% |
| scrubLineNoDirectives | 280 | 256 | 24 | 8.6% |
| scrubBlockWithDirectives | 432 | 416 | 16 | 3.7% |
| csvFieldScan | 1,056 | 1,024 | 32 | 3.0% |
| lazyAltReplaceLarge | 1,536 | 1,504 | 32 | 2.1% |

cushon:
Thanks, I agree this needs more work. If there are CPU/RAM tradeoffs I think we'd probably want to prioritize saving CPU, within reason.

cushon:
#533 gets some of these benefits with fewer tradeoffs, closing

PR #503, opened 2026-06-22
Implement Vector SIMD scanning and required literal precheck for unanchored matches

cushon:
See https://github.com/eaftan/safere/issues/504

---

This commit introduces a two-phase candidate filtering cascade to optimize unanchored matches in SafeRE, bridging the performance gap with native C++ RE2's vectorized pre-filtering:

1. **Required Substring Precheck**: Computes a mandatory literal substring (if present in the AST) during compile time. If the substring is absent in the input text, unanchored matches terminate immediately without invoking the DFA state machine. Restricted to case-sensitive patterns to preserve match correctness.
2. **Vector SIMD Scanning**: Accelerates unanchored searches starting with a single character class using the Java Vector API (incubating in JDK 26) compiling down to native AVX-2/Neon instructions.
3. **Modular Integration & Zero-Copy**:
   * Decouples the Vector API code to an optional `safere-vector` module.
   * Loads the scanner class via a reflection loader bridge; falls back to high-performance inline scalar loops if module or JVM preview flags are absent.
   * Uses `VarHandle` reflection to access `java.lang.String` internal backing array and `coder` byte, executing zero-copy `ByteVector` (Latin-1) and `ShortVector` (UTF-16) scans. Falls back to a lazy `char[]` copy if reflection is blocked.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,6 +39,7 @@
   <modules>
     <module>safere-unicode</module>
     <module>safere</module>
+    <module>safere-vector</module>
     <module>safere-crosscheck</module>
     <module>safere-fuzz</module>
     <module>safere-exhaustive</module>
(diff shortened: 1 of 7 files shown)


cushon:
Closing this, see #504, https://github.com/eaftan/safere/commit/516a1ff76141bc83c3910426332c7824acdf007c, #656

The current prototype for vector scanning for String APIs is #656

PR #506, opened 2026-06-23
Remove start-reliability semantic guards from the DFA sandwich path

cushon:
This depends on the fix in https://github.com/eaftan/safere/pull/505 to disable leftmost-first pruning during reverse DFA scan. After that change the DFA sandwich's reverse scan can now be trusted unconditionally on all regular expressions.

This change also fixes a boundary recording bug in the reverse DFA (`Dfa.java:doSearchReverse`): when a state contains both a deferred boundary match (FLAG_MATCH_BEFORE) and a consuming match (FLAG_MATCH_AFTER_DEFERRED), we now correctly record the leftmost start position (prevPos) instead of unconditionally recording the deferred match position (pos).

This allows patterns like the 'OVERLAPPING_URL' example in https://github.com/eaftan/safere/pull/500 to run on the DFA, which in my benchmarking is ~100x faster and makes SafeRE competitive with RE2 on that benchmark.

This matches RE2's approach, which trusts the reverse DFA in leftmost-longest mode, and doesn't have equivalent semantic guards. The RE2 behaviour here matches `java.util.regex` in practice for google's codebase, so the hope is that this makes SafeRE's approach closer to RE2 performance without regressing compatibility with `java.util.regex`.

The change removes:
1. The `dfaStartReliable()` check and associated AST compilation checks (lazy quantifier flags, bounded repeats, anchors inside quantifiers, unsafe alternations)
2. The direct fallback path in `Matcher.java` that bypassed the DFA precheck for short inputs on "unreliable-start" patterns
3. The `authoritativeStart` logic in the DFA sandwich, since the start position returned by the reverse DFA is now always authoritative.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,13 @@ jobs:
           -Dtest.parallelism=4
           --batch-mode --no-transfer-progress
 
+      - name: Run work-counter scaling tests
+        run: >
+          mvn clean test -pl safere
+          -Pwork-counters
+          -Dtest.parallelism=4
+          --batch-mode --no-transfer-progress
+
   fuzz-regression:
     needs: changes
     if: needs.changes.outputs.code == 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,32 @@ jobs:
           -Dtest.parallelism=4
           --batch-mode --no-transfer-progress
 
+  fuzz-regression:
+    needs: changes
+    if: needs.changes.outputs.code == 'true'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up JDK 25
+        uses: actions/setup-java@v5
+        with:
+          java-version: '25'
+          distribution: 'temurin'
+          cache: maven
+
+      - name: Build SafeRE for fuzz tests
+        run: >
+          mvn -pl safere -am install
+          -DskipTests
+          --batch-mode --no-transfer-progress
+
+      - name: Run fuzz regression seeds
+        run: >
+          mvn -pl safere-fuzz test
+          --batch-mode --no-transfer-progress
+
   benchmark-smoke-test-java:
     needs: changes
     if: needs.changes.outputs.code == 'true'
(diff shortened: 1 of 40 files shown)


cushon:
I've merged in the latest changes here. I've been doing some more internal testing of this, and haven't found any cases where removing the semantics guards introduces divergences with other engines.

cushon:
I found a case where removing the guards was causing divergences. I may have found a fix in https://github.com/eaftan/safere/pull/506/changes/eac31b037626af83fae1676f172c97ce9923805e, but I'm going to move this back to draft a do more due diligence.

I split some of the pre-work for this out into separate PRs (#531, #532)

cushon:
I confirmed I can still reproduce a benefit from this.

The benefit is specifically for patterns that would previously have executed on the NFA because the DFA start position was unreliable. With the fixes in #532, we can always trust the DFA start positions and remove the semantic guards.

I've been doing more testing and validation of this, and for google's codebase there are no patterns where this introduces divergences between SafeRE and RE2, and for production-like performance experiments this seems to help avoid performance gaps compared to RE2-JNI (where RE2 is able to use the DFA for long inputs and SafeRE falls back to the NFA).

For non-pathological cases this is more of a tradeoff, and the NFA might be competitive or slightly better for some cases with simple patterns and short inputs. But in general I think always being able to use the DFA is consistent with SafeRE's goals.

So of the potential performance improvements I've found, I think this is the one I'm most confident in the performance benefit of. I think the question is mostly whether we can fully trust DFA start positions with the fixes in #532, but again the testing I've done shows no divergences.

`lazyAlt` Benchmark Comparison (`textSize=100000`)

```
./run-java-benchmarks.sh '^org\.safere\.benchmark\.Issue488ReplaceAllBenchmark\.lazyAlt_safere$' --     -p textSize=100000
```

| Platform | Baseline (us/op) | PR / Remove Guards (us/op) | Delta (% Change / Speedup) |
| :--- | :--- | :--- | :---: |
| **Linux VM** | `8,339.857 ± 111.715` | `785.293 ± 8.490` | **-90.6% (10.6x faster)** |
| **Mac (Apple Silicon)** | `3,319.761 ± 26.254` | `236.696 ± 8.218` | **-92.9% (14.0x faster)** |

eaftan:
I looked at this and found some correctness bugs in the DFA that are exposed when you remove the guards.  I filed https://github.com/eaftan/safere/issues/538 for it.  To be clear, they were latent bugs in the DFA that just happen to be exposed when you use the DFA for more things; they were not introduced by this PR.

I think I can fix them without giving up the performance benefit of this optimization, which I am able to reproduce 🎉 .  Let me take a cut at #538, then will come back to this.

cushon:
~~I'm seeing an interesting performance regression that may be related to this PR, the theory is that a pattern that previously wasn't DFA eligible became DFA eligible, and then exceeded the DFA budget and fell back to the NFA anyways. IIUC the DFA cache isn't reset so this only happens once per Pattern, but it means the first operation that pattern could be expensive. I don't think it's necessarily a reason not to do this optimization unless the semantic guards are a good predictor of patterns that are likely to exceed the DFA budget, but it was an unexpected side effect.~~

**edit:** I no longer suspect this change, I think most of what I was seeing above was an artifact of state that's lazily initialized the first time the pattern is used.

eaftan (approved):
LGTM.  Looks like a great performance improvement.  I see the same performance improvements on my machine:

For `Issue488ReplaceAllBenchmark.lazyAlt_safere` with `textSize=100000`:

Baseline:
5684.811 ± 277.635 us/op

This PR:
453.169 ± 8.918 us/op (12.x faster)

PR #507, opened 2026-06-23
Specialize EmptyOp checking in BitState and fix BEGIN_LINE end-of-text match boundary bug.

cushon:
Previously, when processing an `OP_EMPTY_WIDTH` instruction (such as `\b`, `^`, `$`, `\A`, or `\z`), BitState called `Nfa.emptyFlags(...)`. This computed the full bitmask of all possible empty-width assertions at that position, including expensive operations like line terminators, unicode word character lookups, and grapheme boundaries.

This change specializes the empty-width check inside the backtracking loop by evaluating only the specific required constraint directly, avoiding unnecessary character lookups, branch checks, and grapheme evaluations.

We also fix a boundary matching bug introduced by this specialization: in the specialized `BEGIN_LINE` check, the assertion was incorrectly matching at `pos == text.length()` when the preceding character was a newline. We resolve this by explicitly returning false when `pos >= textLen`.

Benchmarks show a ~1.61x speedup (61% faster matching) on boundary/anchor heavy patterns like `TEMPLATE_TAG_MATCH` in #500

### DIFF
--- a/safere/src/main/java/org/safere/BitState.java
+++ b/safere/src/main/java/org/safere/BitState.java
@@ -455,10 +455,7 @@ final class BitState {
         }
 
         case InstOp.OP_EMPTY_WIDTH -> {
-          int curFlags =
-              Nfa.emptyFlags(
-                  text, pos, prog.unixLines(), prog.hasGraphemeSemantics(), graphemeContext);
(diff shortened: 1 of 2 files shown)


cushon:
Closing, this is a dupe of #493

PR #508, opened 2026-06-23
Enable BitState execution for patterns containing word boundary assertions

cushon:
This is related to the comment in https://github.com/eaftan/safere/pull/493#pullrequestreview-4540041896

> This came out of a correctness sweep in PR #463, and then there was a round of optimization in PR #466 that restored DFA support for word-boundary patterns.

I think #466 already made the decision to accept an intentional divergence around not splitting surrogate pairs on word boundaries for `java.util.regex` bug compatibility, so there's no reason not to remove this case and allow using BitState here.

This results in a ~8x speedup for capture group patterns containing word boundaries in #500. This also dovetails with #507, which specializes empty-width checking in `BitState`.


cushon:
Closing, consolidated into #493

PR #509, opened 2026-06-23
Implement compile-time case folding expansion for Unicode case-insensitive literals

cushon:
Previously, a case-insensitive literal character range with UNICODE_CASE enabled was compiled to a CHAR_RANGE instruction with `foldCase=true`. At runtime, every character check evaluated the case-insensitive orbit dynamically by binary searching on Unicode tables (`simpleFold`), slowing down NFA/DFA transitions.

This change modifies the literal compilation path in `Compiler.java` to precompute the full case-folding orbit of a rune at compile-time when FOLD_CASE and UNICODE_CASE flags are set. The compiler compiles these orbits to static character class instructions (`InstOp.CHAR_CLASS`).

Because character class instructions use O(1) ASCII bitmaps for characters 0-127 and compact sorted binary search arrays for non-ASCII characters, matching is significantly faster and completely avoids dynamic simpleFold lookups.

This helps with the CASE_INSENSITIVE_KEYWORD and OVERLAPPING_URL cases in #500

### DIFF
--- a/safere/src/main/java/org/safere/Compiler.java
+++ b/safere/src/main/java/org/safere/Compiler.java
@@ -946,16 +946,58 @@ final class Compiler extends Walker<Compiler.Frag> {
 
   private Frag literal(int rune, int flags) {
     boolean foldCase = (flags & ParseFlags.FOLD_CASE) != 0;
-    if (foldCase && (flags & ParseFlags.UNICODE_CASE) == 0) {
-      if (isAsciiLetter(rune)) {
-        // Parser keeps these as folded literals so literal/prefix accelerators can see them.
-        // Lower to an explicit ASCII class here to avoid Unicode simple-fold matching.
-        return asciiFoldedLiteral(rune);
+    if (foldCase) {
+      if ((flags & ParseFlags.UNICODE_CASE) == 0) {
+        if (isAsciiLetter(rune)) {
+          // Parser keeps these as folded literals so literal/prefix accelerators can see them.
+          // Lower to an explicit ASCII class here to avoid Unicode simple-fold matching.
+          return asciiFoldedLiteral(rune);
+        }
+        foldCase = false;
+      } else {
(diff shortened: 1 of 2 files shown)


eaftan:
I couldn't reproduce a meaningful improvement on the `caseInsensitiveKeyword` and `overlappingUrl` cases from #500. Those patterns use `(?i)` rather than `(?iu)`, so they don't exercise the new `FOLD_CASE && UNICODE_CASE` compiler path.

One nuance I noticed while checking the code: Unicode-folded literals with actual case variants already appear to be expanded by the parser into character classes before compilation. This change seems most relevant to remaining literals under `CASE_INSENSITIVE | UNICODE_CASE` that the parser leaves as literals, such as digits, `_`, punctuation, or other code points with identity folding, avoiding runtime `simpleFold` checks for those instructions.

Do you have a workload or benchmark case you can share that demonstrates the intended benefit?

cushon:
You're right, I went back and reviewed how I ended up wtih this:

* this was the result of comparing SafeRE's optimizations to RE2 / RE2J and looking for differences that could explain performance gaps, but it doesn't actually explain any of the gaps in the benchmarks I shared
* modifying `OVERLAPPING_URL` to use `(?iu)` shows a negligible difference
* on the synthetic example of `(?iu)ключевое_слово` matching a 10k character repeat of `"КЛЮЧЕВОЕ_СЛОВО"`, and then disabling DFAs in EngineOptions to actually exercise the optimization, showed a 7% latency reduction from this optimization

So it might help some cases and I think shouldn't hurt performance, but also it seems low priority compared to other potential optimizations. (I think #506 is the biggest opportunity of the ones I've found right now, if it's correct.)

cushon:
I think there is a small NFA-only win to be had here, but it's negligible for the examples I've seen, especially since with #506 the NFA will be used even less. Dropping this for now.

PR #510, opened 2026-06-23
Optimize Pike NFA VM execution performance and garbage collection overhead

cushon:
This came out of work to optimize benchmarks in #500, some of those cases were slow because they fell back to the NFA, and this makes the NFA faster in those cases and also significantly reduces allocations.

That investigation also found some opportunities to use the DFA more (#506) and to use BitState more (#493). I think those other changes are higher priority, and they make the NFA performance less important. This should probably be re-benchmarked after the other changes land to see if it's still worthwhile.

---

This change ports several NFA execution optimizations inspired by RE2J to SafeRE's fallback NFA engine.

* NFA VM Instance Pooling: Cache and reuse `Nfa` instances along with pre-allocated thread state arrays (`q0`, `q1`, `visitedInst`) via a ThreadLocal cache. Avoids allocations during fallback capture-group extraction. Equivalent to RE2J's caching of `Machine` instances.

* Generational Visited Array. Replaces the `Arrays.fill(visitedInst, false)` reset at each character step with an integer-based `visitedGeneration` counter. A state is visited if its entry equals the current generation. Matches RE2J's step generation index tracking in `Machine.java`.

* Combined Code Point Step Decoding: Replaces separate calls to `inputCodePointAt` and `inputNextPos` with a single unified `inputStep` method. The code point value and the next text position index are packed into a single `long` return value. Reduces duplicate index bounds checks and character/surrogate decoding lookups in string bounds. Directly inspired by RE2J's `MachineInput.step` encoding (which packs both the decoded rune and character width into a single integer via `rune << 3 | width`).

* Lazy Empty-Width Assertion Cache (CPU): Caches lookahead/lookbehind empty-width boundary flags (`emptyFlags`) per step, avoiding redundant re-evaluations of word boundaries and text anchors during the epsilon-transition queue traversal for a single character step. Matches how RE2J's NFA simulation precalculates the step condition (`cond`) once per character step and passes it as a parameter, avoiding redundant boundary evaluations during epsilon transitions.

### DIFF
--- a/safere/src/main/java/org/safere/Matcher.java
+++ b/safere/src/main/java/org/safere/Matcher.java
@@ -128,6 +128,11 @@ public final class Matcher implements MatchResult {
   private boolean bitStateBorrowed;
   private int[] bitStateResult;
 
+  /** Cached Nfa instance borrowed from the parent Pattern's thread-local cache. */
+  private Nfa cachedNfa;
+
+  private boolean nfaBorrowed;
+
   /**
    * Whether all capture groups have been resolved. When the DFA sandwich determines match
    * boundaries (group 0), inner captures (groups 1+) are deferred until explicitly requested. This
@@ -305,6 +310,10 @@ public final class Matcher implements MatchResult {
       bitStateBorrowed = false;
       cachedBitState = null;
     }
+    if (nfaBorrowed && cachedNfa != null) {
+      nfaBorrowed = false;
+      cachedNfa = null;
+    }
     bitStateResult = null;
     graphemeContextText = null;
     graphemeContext = null;
@@ -314,6 +323,8 @@ public final class Matcher implements MatchResult {
     bitStateBorrowed = false;
     cachedBitState = null;
     bitStateResult = null;
+    nfaBorrowed = false;
+    cachedNfa = null;
     graphemeContextText = null;
     graphemeContext = null;
   }
@@ -786,19 +797,18 @@ public final class Matcher implements MatchResult {
     // may accept a match ending before a trailing \n. In that case, fall back to the NFA
     // which uses longest-match mode for FULL_MATCH and finds the correct full-text match.
     if (result != null && result[1] != text.length()) {
-      Nfa.SearchResult nfaResult =
-          Nfa.search(
+      int[] nfaResult =
+          searchNfa(
               prog,
-              text,
               0,
               text.length(),
               text.length(),
               0,
+              prog.numCaptures(),
               Nfa.Anchor.ANCHORED,
               Nfa.MatchKind.FULL_MATCH,
-              prog.numCaptures(),
-              graphemeContextFor(prog));
-      result = nfaResult.groups();
+              false);
+      result = nfaResult;
     }
     return applyEngineResult(new FullMatchResult(result));
   }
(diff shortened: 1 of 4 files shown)


cushon:
I did some more benchmarking of this, using `EnginePathOptions` to force the usage of the NFA to directly measure performance

```java
EnginePathOptions options = EnginePathOptions.builder()
    .dfa(false)
    .onePass(false)
    .bitState(false)
    .build();

Pattern pattern = Pattern.compile(regex, flags, options);
```

This does appear to avoid most allocations for the NFA, and improve latency.

It may not matter that much in practice because DFA performance is much more important, but this may still be worth doing eventually if the results are reproducible, because it will still help when the NFA is used. It may also help more with tail latency in real world cases where there there may be more GC pressure, compared to the benchmarks where it's usually less of a factor.

| Case (Pattern Name) | Match | Baseline (ns/op) | Experiment (ns/op) | Speedup (%) |
| **blockedTags1** | true | 4,176.67 | 3,225.81 | **+22.8%** |
| **unprefixedWordBoundary** | true | 968.94 | 712.02 | **+26.5%** |
| **wildcardSearch** | true | 164,091.93 | 144,155.87 | **+12.1%** |
| **blockedTags1** | false | 72,178.88 | 52,970.85 | **+26.6%** |
| **unprefixedWordBoundary** | false | 42,161.20 | 32,384.09 | **+23.2%** |
| **wildcardSearch** | false | 135,932.47 | 114,009.13 | **+16.1%** |

| Case (Pattern Name) | Match | Baseline (B/op) | Experiment (B/op) | Allocation Reduction (%) |
| **blockedTags1** | true | 2,096.03 | 256.02 | **-87.8%** |
| **unprefixedWordBoundary** | true | 1,352.01 | 216.01 | **-84.0%** |
| **wildcardSearch** | true | 1,901.05 | 249.07 | **-86.9%** |
| **blockedTags1** | false | 1,936.51 | 192.37 | **-90.1%** |
| **unprefixedWordBoundary** | false | 1,296.30 | 192.23 | **-85.2%** |
| **wildcardSearch** | false | 1,600.96 | 192.80 | **-88.0%** |

cushon:
Re-opening for consideration. The NFA performance isn't nearly as important as the DFA performance, but I think the results in https://github.com/eaftan/safere/pull/510#issuecomment-4835004382 show this does help with general NFA performance when it is used, and that this has a reasonable return on complexity. Also similar to #501 reducing allocations may help with tail latency and avoid unlucky GC pauses, even if it doesn't show up as much in the benchmarks which are less memory constrained.

eaftan (approved):
I can repro your performance and allocation results.  Happy to take to this.

Latency table

| Case (Pattern Name) | Match | Baseline (ns/op) | Experiment (ns/op) | Speedup (%) |
|---|---:|---:|---:|---:|
| blockedTags1 | true | 2,892.68 | 2,332.18 | +19.4% |
| unprefixedWordBoundary | true | 729.05 | 510.97 | +29.9% |
| wildcardSearch | true | 115,625.86 | 100,610.75 | +13.0% |
| blockedTags1 | false | 45,117.02 | 38,113.88 | +15.5% |
| unprefixedWordBoundary | false | 28,676.48 | 25,234.55 | +12.0% |
| wildcardSearch | false | 84,041.81 | 77,019.34 | +8.4% |

Allocation table

| Case (Pattern Name) | Match | Baseline (B/op) | Experiment (B/op) | Allocation Reduction (%) |
|---|---:|---:|---:|---:|
| blockedTags1 | true | 2,096.04 | 256.03 | -87.8% |
| unprefixedWordBoundary | true | 1,352.01 | 216.01 | -84.0% |
| wildcardSearch | true | 1,897.60 | 249.38 | -86.9% |
| blockedTags1 | false | 1,936.62 | 192.52 | -90.1% |
| unprefixedWordBoundary | false | 1,288.40 | 192.35 | -85.1% |
| wildcardSearch | false | 1,609.16 | 193.06 | -88.0% |

PR #511, opened 2026-06-23
Add more cases to RealWorldRegexBenchmark

cushon:
Some of the regexes in RealWorldRegexBenchmark were adapted from examples, and in the process of doing that stopped showing the benefit of the optimizations they inspired.

* `sparseUrl` shows a 8.86x speedup with #506
* `unprefixedWordBoundary` shows a 2.16x speedup with #493


eaftan (approved):
LGTM.  I made a few changes to integrate these benchmarks in the native RE2 and Go `regexp` benchmarks.  Those are honestly unused at the moment, but I want to make a best effort to keep them running just in case I need them down the road.

PR #512, opened 2026-06-23
Rewrite Bash regular expressions to avoid issues.

eamonnmcmanus:
This might depend on the version of Bash, but I was getting this error:
```
run-java-benchmarks.sh: line 132: [[: invalid regular expression `[^$[]()|*+?]': Invalid preceding regular expression
```
Extracting the regular expression in question into a variable made the error go away.


eaftan:
Thanks @eamonnmcmanus.  This bug came out of a session yesterday where I got too clever with the wrapper script.  I think the longer term solution is to make the wrapper less clever, but this is a good fix for now.

PR #517, opened 2026-06-24
Add ASCII fast paths to DFA

cushon:
This change optimizes the DFA engine with fast paths for ASCII.

* Introduce nested, branch-free tight loops in `doSearch`, `doSearchReverse`, and `doSearchMulti` to scan ASCII characters
* Pre-calculates and caches the `isHighestPriorityMatch` flag directly inside the `State` class during NFA state expansion
* Combines loop exit conditions to avoid JIT branching overhead on matching states, keeping the scanning hot path branch-free for non-matching characters.

This allows the JVM C2 compiler to perform more aggressive loop unrolling and registry renaming.

Benchmark Results (1KB Input, Matching = True):
* wildcardSearch: 11,320 ns/op (vs. 14,534 ns baseline, a 22.1% speedup)
* mapFieldPath: 11,005 ns/op (vs. 17,506 ns baseline, a 37.1% speedup)

### DIFF
--- a/safere/src/main/java/org/safere/Dfa.java
+++ b/safere/src/main/java/org/safere/Dfa.java
@@ -1400,14 +1400,31 @@ final class Dfa {
             break;
           }
           if (ns.isMatch()) {
-            boolean useBefore =
-                (ns.flags & (FLAG_MATCH_BEFORE | FLAG_MATCH_AFTER_DEFERRED)) == FLAG_MATCH_BEFORE;
-            int startPos = useBefore ? pos : pos - 1;
-            if (!longest && !needEndMatch) {
-              return new SearchResult(true, startPos);
+            if ((ns.flags & FLAG_MATCH_BEFORE) != 0) {
+              if (pos >= startLimit && (!needEndMatch || pos == startLimit)) {
+                boolean alreadyMatched = matched;
+                matched = true;
+                matchStart = longest && alreadyMatched ? Math.min(matchStart, pos) : pos;
+                ambiguous |= (ns.flags & FLAG_MATCH_AFTER_DEFERRED) != 0;
+                if (!longest && !needEndMatch) {
+                  return new SearchResult(true, matchStart, ambiguous);
+                }
+              }
             }
-            matched = true;
-            matchStart = startPos;
+            boolean hasOnlyBeforeMatch =
+                (ns.flags & (FLAG_MATCH_BEFORE | FLAG_MATCH_AFTER_DEFERRED)) == FLAG_MATCH_BEFORE;
+            int prevPos = pos - 1;
+            if (!hasOnlyBeforeMatch) {
+              if (prevPos >= startLimit && (!needEndMatch || prevPos == startLimit)) {
+                boolean alreadyMatched = matched;
+                matched = true;
+                matchStart = longest && alreadyMatched ? Math.min(matchStart, prevPos) : prevPos;
+                ambiguous |= (ns.flags & FLAG_MATCH_AFTER_DEFERRED) != 0;
+                if (!longest && !needEndMatch) {
+                  return new SearchResult(true, matchStart, ambiguous);
+                }
+              }
+            }
           }
           s = ns;
           pos--;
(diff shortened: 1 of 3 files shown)


cushon:
This seems to meaningfully improve performance for a couple of benchmark cases where there's the largest gap between SafeRE and RE2-FFM. I also revisited approaches like flatting the DFA state array and that seemed to be worse than just the ASCII fast paths by themselves.

This obviously adds some more code and complexity, but seems potentially worth it to me for the speedup.

eaftan (changes requested):
Here as well, I couldn't reproduce the reported speedups.  

| Case | Baseline | PR head | Observed delta |
|---|---:|---:|---:|
| wildcardSearch | 9,713.308 ± 532.069 ns/op | 8,397.733 ± 272.044 ns/op | 13.5% faster |
| mapFieldPath | 9,420.295 ± 127.597 ns/op | 10,080.845 ± 413.143 ns/op | 7.0% slower |

I wonder if we're benchmarking differently?  Or if there are differences between our systems that are causing differences in our results?

For these I ran the following command at commits ea204c9eda4980a0bf49139641e736ba44d0f5c4 (baseline) and 022e731cace1ecd99f9c85513a77b76de387ad16 (PR):

```bash
    -p engine=SafeRE \
    -p inputSize=1000 \
    -p match=true \
    -p patternName=wildcardSearch,mapFieldPath
```

cushon:
Thanks for checking the numbers here. I currently don't have a good explanation for why we're getting different results, which is definitely worth understanding.

I tried again with the same `./run-java-benchmarks.sh` command on a linux VM and a macbook, and am consistently seeing an improvement from this PR compared to baseline. This was on JDK 26.

You mentioned in #517 your runs were under Windows Subsystem for Linux? The benefit of the ASCII fast path depends on how well the branch is getting predicted, and I wonder if the virtualization is making that less effective. Do you have another machine you could try on without virtualization?

The results I got are

Current Main (`188062d`) vs. Our Optimized Branch (`705e03e`)

**Linux VM**
| Case | Baseline (ns/op) | ASCII Head (ns/op) | Delta (% Speedup) |
| :--- | :--- | :--- | :--- |
| `wildcardSearch` | `14,486.841 ± 492.867` | `11,864.085 ± 132.798` | **+18.1%** |
| `mapFieldPath` | `15,706.759 ± 207.128` | `11,347.541 ± 222.946` | **+27.8%** |

**Mac (Apple Silicon)**
| Case | Baseline (ns/op) | ASCII Head (ns/op) | Delta (% Speedup) |
| :--- | :--- | :--- | :--- |
| `wildcardSearch` | `9,875.815 ± 113.559` | `9,061.220 ± 118.420` | **+8.2%** |
| `mapFieldPath` | `9,111.984 ± 77.170` | `8,633.692 ± 73.382` | **+5.2%** |

Original PR Base (`ea204c9`) vs. Raw PR Head (`022e731`)

**Linux VM**
| Case | Baseline (ns/op) | ASCII Head (ns/op) | Delta (% Speedup) |
| :--- | :--- | :--- | :--- |
| `wildcardSearch` | `14,713.133 ± 586.977` | `11,590.045 ± 89.259` | **+21.2%** |
| `mapFieldPath` | `14,573.517 ± 1,525.506` | `11,186.663 ± 160.649` | **+23.2%** |

**Mac (Apple Silicon)**
| Case | Baseline (ns/op) | ASCII Head (ns/op) | Delta (% Speedup) |
| :--- | :--- | :--- | :--- |
| `wildcardSearch` | `9,671.261 ± 74.495` | `8,928.980 ± 101.056` | **+7.7%** |
| `mapFieldPath` | `8,934.431 ± 79.896` | `8,595.971 ± 93.304` | **+3.8%** |

eaftan (approved):
I think I know why I couldn't repro the results before.  I had made a bugfix to this PR and it must have regressed performance.  I went back and revised it, and now I'm seeing similar performance similar to what you reported.

```java
    -p engine=SafeRE \
    -p inputSize=1000 \
    -p match=true \
    -p patternName=wildcardSearch,mapFieldPath
```

| Case | Baseline | PR | Delta vs baseline |
|---|---:|---:|---:|
| wildcardSearch | 9,713.308 ± 532.069 ns/op | 8,477.624 ± 146.188 ns/op | 12.7% faster |
| mapFieldPath | 9,420.295 ± 127.597 ns/op | 8,596.732 ± 184.054 ns/op | 8.7% faster |

PR #519, opened 2026-06-25
Optimize DFA character class lookup for non-BMP characters

cushon:
Optimize DFA execution performance on BMP characters by replacing binary searches in the hot search loops with a flat array lookup cache.

Performance Impact:
* CJK Search (1k chars): 3.5x speedup (reduced from 945 ns/op to 268 ns/op).
* CJK Non-Match (10k chars): 3.3x speedup (reduced from ~22,000 ns/op to 6,630 ns/op).
* ASCII and Emojis: Neutral (no regressions).

Memory & CPU Tradeoff:
* Memory Overhead: Allocates a single 128 KB `char[] bmpClassMap` array per compiled `Pattern` instance. This represents less than 3% of the existing runtime execution state cache budget (~2 MB to 5 MB).
* Compilation Overhead: Populating the `bmpClassMap` array is optimized via a single O(65536) linear scan sweep (two-pointer style scan) over the sorted boundaries array. This adds negligible overhead to pattern compilation:
  - Simple pattern compilation increases by only 34 nanoseconds (+1.4%).
  - Medium pattern compilation increases by only 1.1 microseconds (+6.2%).

### DIFF
--- a/safere/src/main/java/org/safere/Dfa.java
+++ b/safere/src/main/java/org/safere/Dfa.java
@@ -84,23 +84,28 @@ final class Dfa {
   private static final class State {
     final int[] insts; // sorted NFA instruction IDs (CHAR_RANGE, EMPTY_WIDTH, and MATCH only)
     final int flags;
+    final boolean isHighestPriorityMatch;
+    final State[] next;
 
     /**
      * Match IDs from word-boundary expansion (for PatternSet multi-match). Null if not applicable.
      */
     final int[] wordBoundaryMatchIds;
 
-    /** Transitions indexed by equivalence class; null entry = not yet computed. */
-    final State[] next;
-
-    State(int[] insts, int flags, int numClasses) {
-      this(insts, flags, null, numClasses);
+    State(int[] insts, int flags, int[] wordBoundaryMatchIds, int numClasses) {
+      this(insts, flags, wordBoundaryMatchIds, numClasses, false);
     }
 
-    State(int[] insts, int flags, int[] wordBoundaryMatchIds, int numClasses) {
+    State(
+        int[] insts,
+        int flags,
+        int[] wordBoundaryMatchIds,
+        int numClasses,
+        boolean isHighestPriorityMatch) {
       this.insts = insts;
       this.flags = flags;
       this.wordBoundaryMatchIds = wordBoundaryMatchIds;
+      this.isHighestPriorityMatch = isHighestPriorityMatch;
       this.next = new State[numClasses];
     }
 
@@ -163,20 +168,16 @@ final class Dfa {
   private final int numClasses;
 
   /**
-   * Fast code point to class index lookup table. For ASCII-only patterns, classMap is of size 128
-   * (covering ASCII characters 0–127). For Unicode-enabled patterns, classMap is of size 65536
-   * (covering BMP characters 0–65535).
+   * Fast ASCII-to-class lookup table. For code points 0–127, {@code asciiClassMap[cp]} gives the
+   * equivalence class index directly, avoiding binary search.
    */
-  private final char[] classMap;
+  private final int[] asciiClassMap;
 
   /**
-   * Default class for non-ASCII characters when the pattern is ASCII-only (equals boundaries.length
-   * - 2).
+   * Fast BMP-to-class lookup table. For characters 0–65535, bmpClassMap[ch] gives the class
+   * directly. Allocated lazily only if pattern contains non-ASCII transitions.
    */
-  private final int defaultNonAsciiClass;
-
-  /** Fast path character limit to avoid matching surrogates inside the optimized lookup loops. */
-  private final int fastPathLimit;
+  private final char[] bmpClassMap;
 
   /** State cache: maps instruction-set + flags to canonical State instance. */
   private final Map<StateKey, State> cache = new HashMap<>();
@@ -185,7 +186,7 @@ final class Dfa {
   private final StateKey lookupKey = new StateKey();
 
   /** Sentinel dead state: no instructions, no transitions possible. */
-  private final State deadState = new State(new int[0], 0, 0);
+  private final State deadState;
 
   /** Pre-allocated visited generation array for {@link #expand}, reused across calls. */
   private final int[] expandVisitedGen;
@@ -233,17 +234,7 @@ final class Dfa {
    */
   // TODO(#98): Replace int[] with Guava ImmutableIntArray to get proper value semantics.
   @SuppressWarnings("ArrayRecordComponent")
-  record Setup(int[] boundaries, int numClasses, char[] classMap, int defaultNonAsciiClass) {
-    Setup {
-      // Discard the boundaries array when it is not needed at runtime (i.e. when classMap size
-      // is less than 65536, meaning the pattern's max boundary is small and all out-of-cache
-      // characters map to the same default class). This allows the JVM to reclaim the boundaries
-      // array.
-      if (classMap.length < 65536) {
-        boundaries = null;
-      }
-    }
-  }
+  record Setup(int[] boundaries, int numClasses, int[] asciiClassMap, char[] bmpClassMap) {}
 
   /**
    * Builds a reusable {@link Setup} from a compiled program. The result is immutable and can be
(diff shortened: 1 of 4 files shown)


cushon:
This started off as an investigation into the tradeoffs between having the DFA engine use codepoints vs. UTF-16 code units vs. UTF-8 code units.

I investigated porting the DFA to a 16-bit code unit UTF-16 engine using a post-parse AST rewriter (which decomposed supplementary ranges into high/low surrogate alternations). However that approach couldn't distinguish between a high surrogate that is part of a valid pair (which must be matched as a pair) and an unpaired surrogate, which caused a divergence from JDK behavior on unpaired surrogates.

This approach optimizes the existing code-point based DFA to achieve similar performance to a UTF-16 code unit DFA, by expanding the `asciiClassMap` to cover all BMP characters as a `bmpClassMap`, bypassing dynamic decoding for BMP text while preserving JDK compatibility.

eamonnmcmanus:
Maybe I'm old-fashioned, but 128K per `Pattern` seems kind of a lot to me. I'm also concerned that when people unwisely construct `Pattern` instances inline rather than as constants, they are going to pay more of a performance penalty.

@lowasser has a change in the pipeline that should speed up the binary search here substantially. (For Googlers: cl/936273274.) I wonder if it might be worth waiting for that instead?

cushon:
That's a good point. It's be interesting to know how much of the cache budget is actually used in practice, presumably most patterns are not maxing out the cache and this could be a significant relative footprint increase.

I also don't know how common non-BMP characters are in practice, there was at least one outlier that I turned in the benchmark.

Waiting for @lowasser's change SGTM.

cushon:
I have revised the approach here a bit to only allocate the large `char[]` if the pattern contains non-ASCII ranges, which the vast majority of patterns don't. So there's now zero memory regression for ASCII patterns. There's still a CPU/memory tradeoff for non-ASCII patterns, but I think that makes it a lot more reasonable.

Also the latest version contains an optimization for the ASCII case as well, since we now check if the pattern is ASCII-only we can skip bisecting for the character class for non-ASCII characters. I saw a ~1.6x throughput improvement for ASCII patterns matched against Unicode inputs.

> Waiting for @lowasser's change SGTM.

Actually I think they're two different optimizations, the other improvement is to the unicode tables used compiling the DFA, but not executing it. So the other change is more of a boost of `Pattern.compile` and this helps with matcher/find, so they're complementary.

safere/src/main/java/org/safere/Dfa.java
```diff
@@ -338,15 +352,45 @@ private static void addCaseFoldBoundaries(NavigableSet<Integer> bounds, int lo,
...
-  private static int[] buildAsciiClassMap(int[] boundaries) {
-    int[] map = new int[128];
-    for (int cp = 0; cp < 128; cp++) {
-      int idx = Arrays.binarySearch(boundaries, cp);
-      map[cp] = (idx >= 0) ? idx : (-idx - 1) - 1;
+  private static char[] buildAsciiClassMap(int[] boundaries) {
+    char[] map = new char[128];
+    int classIdx = 0;
+    int boundariesLen = boundaries.length;
+    for (int ch = 0; ch < 128; ch++) {
```
eamonnmcmanus: I'm wondering if you need to special-case ASCII here or if the logic could simply allocate a `char` array that is just big enough to hold all the required values. Also, as a space optimization, when every character past the end of that array belongs to the same equivalence class, perhaps we don't need to retain the `boundaries` array.

cushon: Good ideas, thanks! I have updated it to avoid the ASCII special case and size the `char` array as you suggested, and also avoid retaining the `boundaries` array if we don't need it.

eaftan (dismissed):
Thanks for doing this work!  I ran the benchmarks in the PR description but wasn't able to reproduce the reported speedups.  I also saw some ~10%-ish regressions in some other benchmarks.

Any idea why we're seeing this mismatch?  I wonder if our machines have different performance characteristics.  I'm running these on an Intel Core i7-11700k under Windows Subsystem for Linux.

## PR description benchmarks

| Claim area | Benchmark | Params | Baseline | PR | Change | Result |
|---|---|---|---|---|---|---|
| CJK Search | RealWorldRegexBenchmark.runBenchmark | cjkSearch, inputSize=1000, match=true | 195.110 ± 13.189 ns/op | 159.264 ± 11.157 ns/op | 1.23x faster | Improved, but not 3.5x |
| CJK Non-Match | RealWorldRegexBenchmark.runBenchmark | cjkSearch, inputSize=10000, match=false | 4799.548 ± 188.945 ns/op | 4663.237 ± 66.996 ns/op | 1.03x faster | Roughly neutral, not 3.3x |
| Emoji Search | RealWorldRegexBenchmark.runBenchmark | emojiSearch, inputSize=1000, match=true | 57.282 ± 3.501 ns/op | 57.973 ± 3.023 ns/op | 1.2% slower | Neutral |
| Emoji Non-Match | RealWorldRegexBenchmark.runBenchmark | emojiSearch, inputSize=10000, match=false | 4799.549 ± 195.305 ns/op | 4777.452 ± 225.955 ns/op | 0.5% faster | Neutral |
| Simple Compile | CompileBenchmark.compileSimple_safere | n/a | 1.557 ± 0.099 us/op | 1.590 ± 0.106 us/op | +0.033 us/op | Matches +34 ns claim |
| Medium Compile | CompileBenchmark.compileMedium_safere | n/a | 11.935 ± 0.608 us/op | 11.781 ± 0.408 us/op | 1.3% faster | Did not reproduce +1.1 us overhead |

## Regressions in other benchmarks

| Benchmark | Params | Baseline | PR | Change |
|---|---|---|---|---|
| RealWorldRegexBenchmark.runBenchmark | sparseUrl, inputSize=1000, match=false | 2714.320 ± 46.026 ns/op | 3009.960 ± 51.273 ns/op | 10.9% slower |
| RealWorldRegexBenchmark.runBenchmark | sparseUrl, inputSize=10000, match=false | 27216.216 ± 3558.513 ns/op | 29569.251 ± 1058.090 ns/op | 8.6% slower |
| RealWorldRegexBenchmark.runBenchmark | unprefixedWordBoundary, inputSize=1000, match=false | 2765.880 ± 67.281 ns/op | 3003.699 ± 58.192 ns/op | 8.6% slower |
| RealWorldRegexBenchmark.runBenchmark | unprefixedWordBoundary, inputSize=10000, match=false | 26332.793 ± 605.304 ns/op | 29434.287 ± 1162.651 ns/op | 11.8% slower |

cushon:
I think I was seeing issues like in https://github.com/eaftan/safere/pull/523#issuecomment-4833508408 in previous benchmarks, where the JIT was CPU constrained and wasn't able to optimize well.

Rerunning without those CPU constraints, so the JIT can avoid the bounds checks, I'm also seeing a tradeoff where CJK is better but ASCII is worse. I think that is explained by the larger array resulting in worse memory locality for ASCII.

I've revised this to use a two-tiered approach that keeps a dedicated array for ASCII, and has a separate array for the part that doesn't fit in the ASCII array. This also dovetails with the ASCII fast paths (#517) because we're paying for a branch for ASCII anyways and then there's no additional branching cost to use a separate array for ASCII, and it solves the memory locality issue.

I reran the benchmarks with

```
./run-java-benchmarks.sh --long RealWorldRegexBenchmark.runBenchmark -- \
  -p patternName=cjkSearch -p engine=SafeRE -p inputSize=1000 -p match=true

./run-java-benchmarks.sh --long RealWorldRegexBenchmark.runBenchmark -- \
  -p patternName=sparseUrl,unprefixedWordBoundary -p engine=SafeRE -p inputSize=1000,10000 -p match=false  
```

For `cjkSearch`, 1k, match=true I'm seeing an improvement from 270.199±5.494 ns/op to 239.694±11.282 ns/op (1.13x faster)

With the ASCII fast paths from #517 included in this branch, I am seeing 1.46x-162.x speedups for the `sparseUrl`, `unprefixedWordBoundary` cases (the speed up is mostly #517, but with the latest version of this PR with the separate array for ASCII it should mean there's no regression from this change for ASCII).

I'm curious if that makes sense and you're able to reproduce those results? If you are (and also if you're able to reproduce the numbers for ASCII), then I think ASCII should be merged before this. But if you still can't reproduce I'll investigate more and try to figure out why we're seeing different numbers and if there's anything worthwhile here.

cushon:
This still shows a benchmark win for me but it's definitely a tradeoff with the cost of setting up these tables and the additional memory.

I think it probably warrants more consideration to what the initialization and memory cost of the additional tables is.

I've dropped this one internally for now, I'm going to shelve this and maybe come back to it later and try to get  more measurements on this change in isolation after some of the other performance fixes have stabilized.

cushon:
This area is still a bottleneck for patterns with non-ASCII character classes searching non-ASCII inputs, but I found another approach in #546 that seems to perform better and avoids allocating the giant lookup array. If #546 works out this should be closed.

cushon:
> This area is still a bottleneck for patterns with non-ASCII character classes searching non-ASCII inputs, but I found another approach in #546 that seems to perform better and avoids allocating the giant lookup array. If #546 works out this should be closed.

#546 has been merged, closing

Issue #516, opened 2026-06-24
Methods accepting byte[]


wendigo:
Is there a plan for this library to work with UTF-8-encoded byte[] arrays instead of just accepting Strings? When working with raw memory representations of Strings it's beneficial to avoid converting these back to String.

eaftan:
Hi @wendigo, thank you for the suggestion.

Can you help me better understand your use case? In particular, are you looking for boolean matching over UTF-8 bytes, match byte offsets, capture groups, replacement/splitting, integration with a specific byte-buffer representation, or something else?

SafeRE’s primary goal is to be a safe, mostly drop-in replacement for `java.util.regex`. Because `java.util.regex` is `CharSequence`/`String` based and reports UTF-16 offsets, direct matching over UTF-8 `byte[]` is somewhat outside that core goal.

Technically, this is possible: SafeRE’s compiled regex programs and engine transitions are code-point based. But the execution layer currently uses Java `String`/UTF-16 positions for regions, captures, boundary checks, reverse scans, and match offsets. Supporting UTF-8 bytes directly would require a separate byte-oriented API and a meaningful internal refactor, not just a small overload.

So I’m not sure this fits the near-term direction for SafeRE, but I’d like to understand the use case better before ruling it out.

wendigo:
@eaftan we (https://github.com/trinodb/trino) have been using re2j for a long time as a regex engine for our query engine. Safere seems like a good candidate to include in our engine but internally varchar/chars are represented as flat memory layouts so when we do regex matching we have byte[] for regex and byte[] for value we match against. Going back and fort to String object defeats the benefits of using safere. Ideally the API should allow to match against UTF-8 encoded byte[]

cushon:
I investigated this a bit and experimented with a prototype. Eddie is the decider here for whether this fits the direction for SafeRE, I just thought it was an interesting question and had fun looking in to it a bit :)

I can see the use-case for programs that want to match existing UTF-8 data that is read off the network, without having to re-encoding it as UTF-16 and allocate a copy of it as a Java `String`.

The prototype is in https://github.com/eaftan/safere/pull/520.

My understanding is that RE2J's NFA runs on code points, and it supports decoding UTF-8 `byte[]` inputs on the fly in `UTF8Input` here:

https://github.com/google/re2j/blob/951a6159fdc0a4931593c70442926f86031edf40/java/com/google/re2j/MachineInput.java#L63

[...]

PR #520, opened 2026-06-25
Implement native UTF-8 byte-matching engine for SafeRE

cushon:
This change introduces a specialized `ByteDfa` and `ByteNfa` engine that allows matching patterns directly on raw `byte[]` arrays. This avoids transcoding overheads and completely eliminates steady-state heap allocations.

- Overload `Pattern.compile` and `Pattern.matcher` to support `byte[]` inputs.
- Implement `Utf8Compiler` to decompose AST Unicode ranges to UTF-8 byte-range steps.
- Implement specialized `ByteDfa` and `ByteNfa` search loops.
- Integrate with `dfaSupportsProgram` to fall back on unsupported operations.


cushon:
I fixed a number of correctness issues, but also found two things the code unit engine supports that are difficult to support here:

* grapheme semantics (`\X`) - the byte engine compiles transitions directly on raw bytes. Matching grapheme clusters requires decoding bytes to code points, which is expensive especially for the reverse DFA.

* unicode word boundaries (under `UNICODE_CHARACTER_CLASS`) - checking boundaries requires inspecting the characters immediately before and after the index. Doing this in the byte engine requires expensive reverse UTF-8 decoding to find the preceding character.

The current approach throws an exception if the `byte[]` APIs are used with regexes with those features. Having a fast regex engine that supports `byte[]` and doesn't support those features would be useful for some cases, but is taking a step away from SafeRE's `java.util.regex` compatibility goals.

This could be a reason to reconsider the other approach I mentioned in the bug of re-using the code point engine and having separate loops that transcoded UTF-8 on the fly. If we're paying the cost of transcoding the UTF-8 instead of the DFA transitions just checking individual bytes, there's less point in having a separate byte-based DFA engine, and we may as well just use the code point one.

There would still be memory savings benefit with having to transcode the data because it avoids making a copy of the data to a `String`, but then the matching is much slower, so it would depend on the use-case whether it was worth the copy to get better matching performance.

cushon:
This happened in https://github.com/eaftan/safere/issues/516

Issue #521, opened 2026-06-25
WorkCounter instrumentation overhead


cushon:
I'm seeing a lot of time spent in WorkCounter (added in https://github.com/eaftan/safere/commit/3cead9eabd7cee21f87fdef72dbfccac1bcd087d) for some of the core loops in the DFA.

For an example like `(?is).*\b(options?.*for values|create.*samples?|test plans?|item lists?)\b.*` and a short 100-character input matching one of the alternative branches, `WorkCounter` increases latency per match from ~2.88 µs to ~3.50 µs, or 17.7%. I was looking at this because RE2 is faster than SafeRE for short inputs (for longer inputs the JNI overhead seems to add up more and SafeRE is very close or sometimes better).

Profiling that benchmark showed `WorkCounter.record` was the top hotspot.

Adding a check like this that short circuits the more expensive bookkeeping in `WorkCounter` makes the cost negligible, and it's only important in a couple of hot loops:

```java
if (WorkCounter.activeCounters != 0) {
  WorkCounter.record();
}
```

Longer term it might be worth considering other approaches to get the instrumentation for tests without touching the prod code, like doing testonly bytecode instrumentation to add hooks for tests, without having to add code to the production version.

Also just having a `static final boolean` constant that disabled the logic and was redefined for the tests to be `true` might have similar performance, since hotspot could inline the constant and elide the instrumentation code.

eaftan:
@cushon I put up https://github.com/eaftan/safere/pull/524 to completely remove the WorkCounter instrumentation from production builds as an alternative to #522.  WDYT?

cushon:
Thanks @eaftan, that approach looks great!

eaftan:
Fixed by #524

PR #522, opened 2026-06-25
Reduce WorkCounter instrumentation overhead

cushon:
Guard calls to WorkCounter APIs in some core loops to avoid overhead.

https://github.com/eaftan/safere/issues/521


safere/src/main/java/org/safere/WorkCounter.java
```diff
@@ -8,7 +8,7 @@
 /** Package-private deterministic work accounting for performance regression tests. */
 final class WorkCounter {
   private static final ThreadLocal<Counter> COUNTER = new ThreadLocal<>();
-  private static int activeCounters;
+  static int activeCounters;
```
eamonnmcmanus, line 11: Maybe add a getter method for this rather than exposing a mutable static field?

cushon:
Closing in favour of #524

PR #523, opened 2026-06-25
BitState stack consolidation optimization

cushon:
For a regex like `(?s)<markerA>.*?<markerB>` with lazy wildcard repetitions and a short 100-character input matching the pattern like `<markerA>Lorem ipsum dolor sit amet, consectetur adipiscing elit.<markerB>`, packing the `jobInstId` and `jobPos` into a single `long[]` in `BitState` improves cache locality and reduces the number of array bounds checks, and improved latency by ~17% for that benchmark.

I noticed this looking at cases where RE2 outperforms SafeRE for short ~100 character inputs.

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -30,6 +30,11 @@
       "pattern": "[a-zA-Z0-9._%+\\-]+@[a-zA-Z0-9.\\-]+\\.[a-zA-Z]{2,}",
       "text": "contact user.name+tag@example.co.uk for info",
       "op": "find"
+    },
+    "lazyWildcardFind": {
+      "pattern": "(?s)<markerA>.*?<markerB>",
+      "text": "<markerA>Lorem ipsum dolor sit amet, consectetur adipiscing elit.<markerB>",
+      "op": "find"
     }
   },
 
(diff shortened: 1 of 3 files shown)


eaftan (changes requested):
I added the benchmark you mentioned in the PR description and tried to reproduce your result, but I couldn't repro.  On my machine it seems to be about a 13% slowdown.

Would you mind testing with the benchmark I added and sharing your numbers?

| Revision | Benchmark | Result |
|---|---|---|
| PR base ea204c9 + benchmark patch | RegexBenchmark.lazyWildcardFind_safere | 765.827 ± 39.498 ns/op |
| PR head 569deac | RegexBenchmark.lazyWildcardFind_safere | 866.769 ± 15.639 ns/op |

cushon:
I did some more measurement of this, I am not seeing a benefit from this change and don't think it should be pursued.

I think the measurement I saw earlier is explained by JIT variance across different VM starts when there are CPU constraints. Some of the earlier benchmarking was done in a sandbox that had restricted access to CPU. Benchmarking the optimization under `taskset -c 0` reproduces similar variance where there's a 16% benefit for some forks of the benchmark.

Also if there was a benefit here, we'd expect it to just affect `BitState`. The `fruitSearchQuery` case in #527 is an example that should use `BitState`.

Re-running on a linux VM and my mac without resource constraints show a regression for `fruitSearchQuery`.

One possible theory for that is that if the JIT isn't resource constrained it'll be able to eliminate the bounds checks and the single array is a pessimization from the additional operations to pack/unpack the array values, but if it is resource contained it can't eliminate the bounds checks and then a single array avoids bounds checks for the second array and the packing/unpacking is worth it.

I also did some benchmarking together with the ASCII fast path (#517) to see if there was a combined benefit, and I do not see any combined benefit, although the ASCII fast path (with and without this change) is still showing a benefit compared to baseline for me.

PR #524, opened 2026-06-26
Disable WorkCounters in production builds

eaftan:
This is an alternative to #522 that completely removes WorkCounter instrumentation from production builds. I use a compile-time constant boolean that is set to false in the production config and true in a limited test config that applies only the the specific tests that need WorkCounters. javac optimizes any if (false) blocks away, so in the production build there is no more WorkCounter instrumentation.

I verified using javap that the generated bytecode in production config no longer contains any references to WorkCounter. I also added a test for this. Finally, I made it so that attempts to read work counters when they are disabled throw an exception.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,13 @@ jobs:
           -Dtest.parallelism=4
           --batch-mode --no-transfer-progress
 
+      - name: Run work-counter scaling tests
+        run: >
+          mvn clean test -pl safere
+          -Pwork-counters
+          -Dtest.parallelism=4
+          --batch-mode --no-transfer-progress
+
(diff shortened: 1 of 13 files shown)


cushon:
Thanks, this looks great. #522 was a quick hack, but having this completely disabled for production builds seems like the way to go.

For the CI failure I think `WorkCounterTest.java` just needs to be in the list of excluded files in `safere-crosscheck/pom.xml` under the `crosscheck-public-api-tests` profile (and also `@DisabledForCrosscheck`), since it tests the implementation and not the public API.

cushon (approved):

eaftan:
Thanks for fixing the CI failure!

Issue #525, opened 2026-06-26
BitState performance compared to RE2


cushon:
I was investigating a gap with RE2 performance on some patterns that use BitState.

The patterns use features that require BitState like capture groups, aren't OnePass eligible, and the matches had short inputs that were BitState eligible.

`(?is).*\b(items?.*for display|show.*entries?|entry lists?|display lists?|item lists?)\b.*`

`^\s*<(\QElement\E|\QElementContainer\E|\QElementGroup\E)>\s*$`

Edit:

* I explored one alternative to BitState in https://github.com/eaftan/safere/pull/526#issuecomment-4808739847, but after redoing the benchmarks it seems like a dead end
* I also experimented with a few other micro-optimizations and couldn't find anything that wasn't neutral or a regression (revisiting the packed `jobStack`, an ASCII fast path avoiding `codePointAt`, replacing `instId * textSlots` with `instOffsets[instId]`)

I'll open a separate PR for just the benchmarks that I used for this investigation.

The working theory for the ~2.5x gap on examples like `fruitSearchQuery` is that C++ RE2 can do raw pointer-arithmetic graph traversal, and SafeRE is paying for bounds-checking the array accesses. If so there may not be much to be done here, and I don't think this specific benchmark is that important a use-case, but I'd be interesting a second opinion on whether there are worthwhile opportunities here.

cushon:
RE2 still wins here, but I don't currently see any optimization opportunities for BitState that make sense.

There has also been a lot of other work (like the gap engine in #769) to use BitState less. I think there might still be some opportunities to expand that in ways that would help here, like for `fruitSearchQuery` if we were able to use the reverse DFA.

```
Benchmark                                                                                           (crossEngineTrial)  Mode  Cnt      Score      Error  Units                                                                                      

CrossEngineBenchmark.run                RealWorldRegexBenchmark.runBenchmark.fruitSearchQuery.match.1000@safere-string  avgt   10  22837.415 ± 1443.088  ns/op                                                                                      
CrossEngineBenchmark.run    RealWorldRegexBenchmark.runBenchmark.fruitSearchQuery.match.1000@re2-ffm-string-conversion  avgt   10  13341.189 ±  290.128  ns/op                                                                                      
CrossEngineBenchmark.run              RealWorldRegexBenchmark.runBenchmark.fruitSearchQuery.noMatch.1000@safere-string  avgt   10   1968.916 ±   10.835  ns/op                                                                                      
[...]

PR #526, opened 2026-06-26
Implement hybrid stack-and-recursion backtracking in BitState

cushon:
Introduce a hybrid backtracking engine in BitState that runs recursive matching on the JVM call stack for optimal latency (yielding 1.7x-2.7x speedup on non-pathological inputs), but dynamically falls back to a flat iterative loop if the recursion depth exceeds a safety limit (512), preventing StackOverflowError.

To ensure functional correctness and prevent priority inversion during the fallback:
- If a recursive search iteration hits the depth limit, we abort that start position iteration, completely unwind the recursive call stack, reset the registers/visited bitmap, and rerun that start position iteratively from the beginning using a flat loop.
- This "abort-and-restart" strategy preserves the correct LIFO backtracking choice points and register values, avoiding state corruption or missing matches on complex nested loop boundaries (e.g. \B patterns).

Also add EnginePathOptions support to force the iterative fallback loop, allowing equivalence verification of both paths against the JDK.

https://github.com/eaftan/safere/issues/525


cushon:
> yielding 1.7x-2.7x speedup

I did more rigorous benchmarking and can't reproduce this.

I think the idea of recursion was related to https://github.com/eaftan/safere/pull/496, but the final version of that PikeVM change also wasn't recursive (because of stack safety) and the win there was from avoiding allocations, and BitState already doesn't allocate.

I'll open a separate PR for the benchmarks that prompted this investigation, but recursive BitState seems like a dead end.

PR #527, opened 2026-06-26
Add more real-world benchmarks

cushon:
This commit introduces two representative regex patterns :
- `fruitSearchQuery`: A search-heavy, wildcard-backtracking pattern.
- `fruitMarkupTag`: A short, anchored tags matching pattern.

These patterns are used to evaluate the comparative performance between SafeRE (iterative BitState) and native C++ RE2 (FFM). The new benchmarks resolve subgroups, forcing the use of `BitState`.

Comparative Performance Metrics (ns/op, lower is better):
- fruitSearchQuery (1000 chars, match=true):
  * SafeRE:           33,209 ns
  * RE2_FFM (C++):    13,711 ns (C++ is 2.42x faster)
- fruitMarkupTag (1000 chars, match=true):
  * SafeRE:              100 ns
  * RE2_FFM (C++):       629 ns (SafeRE is 6.26x faster due to avoiding FFM transition overhead)

https://github.com/eaftan/safere/issues/525

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -323,10 +323,12 @@
         "match": "  <Banana>  ",
         "nonMatch": "  <Orange>  ",
         "matchInput": {
-          "kind": "repeat"
+          "kind": "surroundWithSpaces",
+          "body": "<Banana>"
         },
         "nonMatchInput": {
-          "kind": "repeat"
+          "kind": "surroundWithSpaces",
+          "body": "<Orange>"
         }
       }
     ]
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -302,6 +302,32 @@
           "kind": "prefixedRepeat",
           "prefix": "12345 "
         }
+      },
+      {
+        "name": "fruitSearchQuery",
+        "pattern": "(?is).*\\b(apples?.*for banana|cherry.*dates?|elderberry lists?|fig lists?|grape lists?)\\b.*",
+        "op": "replaceAllGroup1",
+        "match": "Here are the apples to show for banana in this session.",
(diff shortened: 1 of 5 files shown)


safere-benchmarks/benchmark-data.json
```diff
@@ -302,6 +302,32 @@
...
+        },
+        "nonMatchInput": {
+          "kind": "repeat"
+        }
+      },
+      {
+        "name": "fruitMarkupTag",
+        "pattern": "^\\s*<(\\QApple\\E|\\QBanana\\E|\\QCherry\\E)>\\s*$",
+        "op": "replaceAllGroup1",
+        "match": "  <Banana>  ",
```
eaftan, line 323: I think with `matchInput.kind` set to `"repeat"`, this won’t match because the pattern is anchored at the start and end. Would it better match the intent of this benchmark case to use a single tag padded with spaces to the target input size?

For example, this case could use a new input kind like:

```json
"matchInput": {
  "kind": "surroundWithSpaces",
  "body": "<Banana>"
}
```

where the generated input is `body` padded with leading/trailing spaces until it reaches `inputSize`. That would still benchmark a 1000/10000-char anchored input, but the `match=true` case would actually match and exercise the group-1 replacement.

Optionally, the non-match input could use the same shape for symmetry:

```json
"nonMatchInput": {
  "kind": "surroundWithSpaces",
  "body": "<Orange>"
}
```

eaftan (changes requested):

safere-benchmarks/benchmark-data.json
cushon: Thanks for the catch, yes, I think that's right. I have added `surroundWithSpaces` support to the benchmarks and updated these cases to use it.

eaftan (approved):
Fixed a small issue with the C++ benchmarks.  LGTM

Issue #528, opened 2026-06-26
DFA loop unrolling and nested nullable loops


cushon:
I have been investigating `customProtocolLink` performance for `input_size=1000`, `match=false`, which appears to be significantly worse for SafeRE than RE2-FFM.

I think this difference comes down to SafeRE's handling of nested nullable loops (e.g. `(?:a*)*`). SafeRE allocates compile-time loop registers to prevent infinite loops in its NFA engine, which prevents the use of the DFA for these patterns. In C++ RE2, the NFA engine prevents infinite loops during epsilon transitions by tracking a bitmask of visited states dynamically at runtime.

Proposal: unroll greedy nullable loops in the AST prior to compilation. This eliminates the need for loop registers in the DFA program, allowing SafeRE to achieve the same DFA-clean representation as C++ RE2.

Alternatives:

* C++ RE2's approach of a bitmask of visited NFA states: This requires tracking visited states in the NFA and BitState engines. This is hard to implement as efficiently on the JVM due to GC and memory access overheads, and would regress NFA/BitState execution performance for all regexes to optimize for this case.
* Unrolling nongreedy loops: The proposal only unrolls greedy loops, since nongreedy loops (e.g. `a*?`) are harder to implement correctly with leftmost-first matching in this scheme. Simplifying a nongreedy loop to `a+?` deletes its inner bypass path, which moves the choice to the greedy outer loop and inverts the priority of matching empty. One way to preserve leftmost-first semantics would be to label NFA states with priorities, which would add complexity and overhead, for a case that is much rarer in practice than greedy nullable loops.

cushon:
#529 has been merged. The loop unrolling approach created correctness problems (https://github.com/eaftan/safere/pull/529#pullrequestreview-4597252506), the PR pivoted to a sound over approximation that uses the DFA for a fast-reject pass but doesn't trust the positions it reports (https://github.com/eaftan/safere/pull/529#issuecomment-4844089682)

PR #529, opened 2026-06-26
Enable DFA matching for greedy nested nullable loops via AST unrolling

cushon:
See https://github.com/eaftan/safere/issues/528

This specifically helps cases like `customProtocolLink` with nested nullable loops, and allows using the DFA.

For negative matches this is ~17x faster for 100 chars and ~45x faster for 1000 chars.

---

```
./run-java-benchmarks.sh RealWorldRegexBenchmark.runBenchmark -- \
  -p patternName=customProtocolLink \
  -p engine=SafeRE,RE2_FFM
```

Before:

```
Benchmark                             (engine)  (inputSize)  (match)       (patternName)  Mode  Cnt        Score      Error  Units
RealWorldRegexBenchmark.runBenchmark    SafeRE         1000     true  customProtocolLink  avgt   10      600.532 ±   10.966  ns/op
RealWorldRegexBenchmark.runBenchmark    SafeRE         1000    false  customProtocolLink  avgt   10   307536.841 ± 4499.326  ns/op
RealWorldRegexBenchmark.runBenchmark    SafeRE        10000     true  customProtocolLink  avgt   10     7316.213 ±  254.207  ns/op
RealWorldRegexBenchmark.runBenchmark    SafeRE        10000    false  customProtocolLink  avgt   10  1196376.338 ± 9472.251  ns/op
RealWorldRegexBenchmark.runBenchmark   RE2_FFM         1000     true  customProtocolLink  avgt   10      927.537 ±   57.289  ns/op
RealWorldRegexBenchmark.runBenchmark   RE2_FFM         1000    false  customProtocolLink  avgt   10     1883.303 ±   62.131  ns/op
RealWorldRegexBenchmark.runBenchmark   RE2_FFM        10000     true  customProtocolLink  avgt   10     1998.314 ±  119.532  ns/op
RealWorldRegexBenchmark.runBenchmark   RE2_FFM        10000    false  customProtocolLink  avgt   10    17941.803 ±  550.960  ns/op
```

After:

```
Benchmark                             (engine)  (inputSize)  (match)       (patternName)  Mode  Cnt      Score     Error  Units
RealWorldRegexBenchmark.runBenchmark    SafeRE         1000     true  customProtocolLink  avgt   10    353.138 ±  18.972  ns/op
RealWorldRegexBenchmark.runBenchmark    SafeRE         1000    false  customProtocolLink  avgt   10   6854.579 ±  27.686  ns/op
RealWorldRegexBenchmark.runBenchmark    SafeRE        10000     true  customProtocolLink  avgt   10    341.841 ±  12.931  ns/op
RealWorldRegexBenchmark.runBenchmark    SafeRE        10000    false  customProtocolLink  avgt   10  69344.533 ± 725.002  ns/op
RealWorldRegexBenchmark.runBenchmark   RE2_FFM         1000     true  customProtocolLink  avgt   10    903.762 ±  20.215  ns/op
RealWorldRegexBenchmark.runBenchmark   RE2_FFM         1000    false  customProtocolLink  avgt   10   1883.994 ±  71.952  ns/op
RealWorldRegexBenchmark.runBenchmark   RE2_FFM        10000     true  customProtocolLink  avgt   10   1913.986 ±  87.076  ns/op
RealWorldRegexBenchmark.runBenchmark   RE2_FFM        10000    false  customProtocolLink  avgt   10  18098.031 ± 698.583  ns/op
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,13 @@ jobs:
           -Dtest.parallelism=4
           --batch-mode --no-transfer-progress
 
+      - name: Run work-counter scaling tests
+        run: >
+          mvn clean test -pl safere
+          -Pwork-counters
+          -Dtest.parallelism=4
+          --batch-mode --no-transfer-progress
+
   fuzz-regression:
     needs: changes
     if: needs.changes.outputs.code == 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,32 @@ jobs:
           -Dtest.parallelism=4
           --batch-mode --no-transfer-progress
 
+  fuzz-regression:
+    needs: changes
+    if: needs.changes.outputs.code == 'true'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up JDK 25
+        uses: actions/setup-java@v5
+        with:
+          java-version: '25'
+          distribution: 'temurin'
+          cache: maven
+
+      - name: Build SafeRE for fuzz tests
+        run: >
+          mvn -pl safere -am install
+          -DskipTests
+          --batch-mode --no-transfer-progress
+
+      - name: Run fuzz regression seeds
+        run: >
+          mvn -pl safere-fuzz test
+          --batch-mode --no-transfer-progress
+
   benchmark-smoke-test-java:
     needs: changes
     if: needs.changes.outputs.code == 'true'
@@ -73,10 +99,10 @@ jobs:
           cache: maven
 
       - name: Smoke test Java benchmarks
-        run: ./run-java-benchmarks.sh --smoke RegexBenchmark
+        run: ./run-java-benchmarks.sh --smoke '^org\.safere\.benchmark\.RegexBenchmark\.'
 
       - name: Smoke test Java memory benchmarks (GC profiler)
-        run: ./run-java-memory-benchmarks.sh --smoke RegexBenchmark
+        run: ./run-java-memory-benchmarks.sh --smoke '^org\.safere\.benchmark\.RegexBenchmark\.'
 
       - name: Smoke test Java memory benchmarks (pattern sizes)
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,6 +33,13 @@ jobs:
       - name: Build and test
         run: mvn -pl safere verify -Dtest.parallelism=4 --batch-mode --no-transfer-progress
 
+      - name: Run work-counter scaling tests
+        run: >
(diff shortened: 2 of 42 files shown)


eaftan (changes requested):
My agent found a case where this optimization breaks correctness.  Here's a repro:

```java
public class SafeReDfaAstBugRepro {
    String regex = "(?:a?\\b?)*X|(?:a?b)*c";
    String input = "bc";

    boolean jdk = java.util.regex.Pattern.compile(regex).matcher(input).matches();
    boolean safere = org.safere.Pattern.compile(regex).matcher(input).matches();

    System.out.printf("regex:  %s%n", regex);
    System.out.printf("input:  %s%n", input);
    System.out.printf("JDK:    %s%n", jdk);
    System.out.printf("SafeRE: %s%n", safere);
    System.out.printf("BUG:    %s%n", jdk != safere);
  }
}
```

Output: 

```
regex:  (?:a?\b?)*X|(?:a?b)*c
input:  bc
JDK:    true
SafeRE: false
BUG:    true
```

cushon:
Thanks for the catch.

I've revised this. I think the AST unrolling approach was flawed, and don't see a way to ensure the match boundaries are correct.

Instead of AST unrolling, this change now creates a DFA without progress checks and loop registers, which should be a sound over-approximation. And then it uses that DFA for a fast reject pass, and if it fails to reject the pattern still falls back to another engine instead of trusting the match boundaries, like the handling of other capture groups.

This is stacked on #506, the benchmarks show an improvement over #506 for cases like `Issue481ScalingBenchmark.schemeFindNegative_safere`.

```
./run-java-benchmarks.sh Issue481ScalingBenchmark.schemeFindNegative_safere -- \
    -p textSize=1024,10240,102400
```

| Text Size | `remove-semantic-guards` (PR 506 only) | `unrolled-dfa` (Combined Changes) | Speedup Factor |
| :--- | :--- | :--- | :--- |
| **1024** | 3.741 ± 0.063 μs/op | 0.515 ± 0.016 μs/op | **7.26x** |
| **10240** | 9.571 ± 0.119 μs/op | 1.428 ± 0.028 μs/op | **6.70x** |
| **102400** | 19.061 ± 0.319 μs/op | 10.756 ± 0.126 μs/op | **1.77x** |

eaftan (approved):
Thanks for the updated version.  I can repro the performance improvement on `Issue481ScalingBenchmark.schemeFindNegative_safere`:

| textSize | main | PR merged worktree | measured speedup |
|---:|---:|---:|---:|
| 1024 | 2.603 us/op | 0.383 us/op | 6.80x |
| 10240 | 6.613 us/op | 1.157 us/op | 5.72x |
| 102400 | 14.433 us/op | 8.782 us/op | 1.64x |

PR #530, opened 2026-06-26
Prototype: Support matching on raw byte arrays via InputScanner abstraction

cushon:
Introduces the InputScanner interface to abstract text access (backed by String or raw UTF-8 byte arrays). Unifies Dfa, Nfa, BitState, and OnePass match loops to operate on InputScanner, avoiding core loop duplication and preventing String performance regressions.

#516


cushon:
https://github.com/eaftan/safere/issues/516

PR #531, opened 2026-06-28
Remove obsolete dfaGroupZeroReliable and groupZeroResolved dead code

cushon:
This became dead code after 7889a3bd, which made `dfaGroupZeroReliable()` always return `true`. `groupZeroResolved` was used to track whether the NFA needed to be run to find `group(0)` end positions if the DFA positions weren't reliable, but that logic is now also unused.

This was included in #506, which needs more work, I'm moving it to a separate PR.


eaftan (approved):
thanks for the cleanup!

PR #532, opened 2026-06-28
Fix forward and reverse DFA priority inversion for boundary matches

cushon:
When a DFA state contains both a deferred boundary match (`FLAG_MATCH_BEFORE`) and a standard consuming match, the engine must distinguish between them to preserve leftmost-first priority. Without this, overlapping alternation branches or lazy quantifiers can suffer from priority inversion or false-negative mismatches.

This PR integrates the `FLAG_MATCH_AFTER_DEFERRED` flag into both the forward (`doSearch`/`doSearchMany`) and reverse (`doSearchReverse`) DFA loops to enforce correct leftmost-first priority.

Resolves bugs on patterns such as:

1. `\ba|a` (reverse DFA boundary match)
2. `(?:"(?:[^"\\]|\\.)*"|'(?:[^'\\]|\\.)*')|(\btype\b)`
3. `\[.*?\]\((.*?)\)|(\b\w+\.md\b)`

These correctness fixes were originally bundled inside #506 because removing the guards exposed the forward DFA bug. This extracts the fixes into a separate change to prepare for #506.

* `Dfa.java`:
  * Forward loops check `FLAG_MATCH_AFTER_DEFERRED` to record matches at the correct position when both deferred and consuming matches exist.
  * Reverse loop prioritizes `prevPos` over `pos` when both match types exist.
* `MatcherTest.java` / `DfaTest.java`: Added tests for alternation, lazy quantifier, and reverse DFA boundary correctness.
* `EnginePathEquivalenceTest.java`: Added cross-engine differential tests comparing NFA, BitState, and DFA traces on these patterns. Removed the obsolete `dfaStartReliabilityGuardHasSemanticContent` check since it asserts that the DFA start position is unreliable in the case fixed by this PR.

### DIFF
--- a/safere-fuzz/src/test/java/org/safere/fuzz/MatchFuzzer.java
+++ b/safere-fuzz/src/test/java/org/safere/fuzz/MatchFuzzer.java
@@ -142,11 +142,18 @@ final class MatchFuzzer {
 
   private static void assertDfaSandwichLeftmostStartCasesMatchJdk() {
     List<String> regexes =
-        List.of("\\B([^a])*[^a][^a]", "\\B(?:[^a])*[^a][^a]", "\\B[^a]*[^a][^a]", "\\B(?:b|bb)*bb");
+        List.of(
+            "\\B([^a])*[^a][^a]",
+            "\\B(?:[^a])*[^a][^a]",
+            "\\B[^a]*[^a][^a]",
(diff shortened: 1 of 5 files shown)


eaftan (approved):
Fixed a correctness issue.  LGTM, will merge after CI passes.

PR #533, opened 2026-06-28
Bypass DFA sandwich and run BitState directly on small inputs

cushon:


cushon:
https://github.com/eaftan/safere/issues/534#issuecomment-4836016856

PR #537, opened 2026-06-29
Fix ArrayIndexOutOfBoundsException in Dfa.computeNext after 7889a3bd2234d75f7185f59e0b9724dbcafbc4f9

cushon:
During empty-width assertion re-expansion in computeNext, we collect NFA states in priority order. When resolving empty transitions, multiple active states can expand to overlapping sets of reachable states. These transient duplicates are temporarily stored in `tempExpanded` before being deduplicated by `stableDedup`.

For complex patterns with large loops (e.g. unrolled loops of empty assertions), the number of duplicates can exceed the NFA program size, causing `tempExpanded` (which was allocated with size `prog.instructions.size()`) to overflow with an `ArrayIndexOutOfBoundsException`.

This fix dynamically resizes the `tempExpanded` array if it reaches capacity. Because the final unique state count is bounded by the program size, the array will grow only as needed to accommodate transient duplicates and is safely deduplicated immediately after.

This bug regressed in 7889a3bd2234d75f7185f59e0b9724dbcafbc4f9, which replaced the previous `mergeInsts` state deduplication logic with the pre-allocated flat array and `stableDedup` collection approach.

### DIFF
--- a/safere/src/main/java/org/safere/Dfa.java
+++ b/safere/src/main/java/org/safere/Dfa.java
@@ -909,6 +909,9 @@ final class Dfa {
           // Add non-MATCH instructions from this expansion.
           for (int x : expanded) {
             if (prog.inst(x).opCode != InstOp.OP_MATCH) {
+              if (tempCount == tempExpanded.length) {
+                tempExpanded = Arrays.copyOf(tempExpanded, tempExpanded.length * 2);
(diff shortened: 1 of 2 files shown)


eaftan (approved):
Nice find!  LGTM

PR #540, opened 2026-06-30
Aho-Corasick Literal Pre-filtering

cushon:
This adds a fast path to extract compile-time literals and implement Aho-Corasick pre-filtering/start-acceleration to optimize search performance on some patterns. See: https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm

This acts as a generalized prefix acceleration, which supports a set of prefixes using a trie. This allows SafeRE to reject non-matching inputs faster, and start the search with the full regex engine at a matching position for matches.

This change adds some benchmarks adapted from real world patterns.

| Benchmark (1000 chars, mismatch) | SafeRE Baseline | SafeRE Optimized | Speedup (vs Baseline) | JDK Engine | SafeRE Speedup vs JDK |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **`intlPhonePrefix`** (leading alternation) | `7,494 ns/op` | `3,503 ns/op` | **2.14x speedup** | `67,073 ns/op` | **19.1x faster** |
| **`emailDomain`** (required suffix scan) | `6,895 ns/op` | `3,450 ns/op` | **2.00x speedup** | `46,941 ns/op` | **13.6x faster** |
| **`stateCodes`** (anchored fast-reject) | `45.8 ns/op` | `48.3 ns/op` | ~2.5 ns overhead | `66.1 ns/op` | 1.37x faster |
| **`multiWordPrefixAlternation`** (anchored) | `45.9 ns/op` | `48.7 ns/op` | ~2.8 ns overhead | `30.6 ns/op` | 0.63x (slower) |

Implementation

* LiteralExtractor.java: walks the AST, extracting either the longest contiguous required prefix from all alternation branches, or fallback required infix substrings. It also detects and extracts prefixes from leading alternations nested inside concatenations.
* AhoCorasickSearcher.java: Compresses the lookup trie into primitive arrays using a compressed sparse row representation.
* Pattern.java / Matcher.java integration: performs a pre-filter scan at the start of `doFindCore()`, and if a match is found and the pattern starts exactly with the prefix list skips directly to the matched offset

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -1967,6 +1967,114 @@
       "shared": true
     },
     {
+      "id": "realWorldRegex.unanchoredStateCodes.match.{size}",
+      "axes": {
+        "size": [
+          1000,
+          10000,
+          100000
+        ]
+      },
+      "recipe": {
+        "kind": "delimitedRepeatToLength",
+        "value": "Shipping destination CA address verified successfully.",
+        "delimiterAlphabet": " ",
+        "seed": 42,
+        "length": "{size}"
+      },
+      "shared": true
+    },
+    {
+      "id": "realWorldRegex.unanchoredStateCodes.noMatch.{size}",
+      "axes": {
+        "size": [
+          1000,
+          10000,
+          100000
+        ]
+      },
+      "recipe": {
+        "kind": "delimitedRepeatToLength",
+        "value": "the quick brown fox jumps over the lazy dog and explores various places around the globe",
+        "delimiterAlphabet": " ",
+        "seed": 42,
+        "length": "{size}"
+      },
+      "shared": true
+    },
+    {
+      "id": "realWorldRegex.unanchoredLogKeywords.match.{size}",
+      "axes": {
+        "size": [
+          1000,
+          10000,
+          100000
+        ]
+      },
+      "recipe": {
+        "kind": "delimitedRepeatToLength",
+        "value": "2026-08-12 12:00:00 [ERROR] Connection timed out while reading from database socket.",
+        "delimiterAlphabet": " ",
+        "seed": 42,
+        "length": "{size}"
+      },
+      "shared": true
+    },
+    {
+      "id": "realWorldRegex.unanchoredLogKeywords.noMatch.{size}",
+      "axes": {
+        "size": [
+          1000,
+          10000,
+          100000
+        ]
+      },
+      "recipe": {
+        "kind": "delimitedRepeatToLength",
+        "value": "2026-08-12 12:00:00 [INFO] Request processed successfully in 42ms with status 200 OK.",
+        "delimiterAlphabet": " ",
+        "seed": 42,
+        "length": "{size}"
+      },
+      "shared": true
+    },
+    {
+      "id": "realWorldRegex.unanchoredFruitKeywords.match.{size}",
+      "axes": {
+        "size": [
+          1000,
+          10000,
+          100000
+        ]
+      },
+      "recipe": {
+        "kind": "delimitedRepeatToLength",
+        "value": "I would like to order a fresh banana shake with extra ice please.",
+        "delimiterAlphabet": " ",
+        "seed": 42,
+        "length": "{size}"
+      },
+      "shared": true
+    },
+    {
+      "id": "realWorldRegex.unanchoredFruitKeywords.noMatch.{size}",
+      "axes": {
+        "size": [
+          1000,
+          10000,
+          100000
+        ]
+      },
+      "recipe": {
+        "kind": "delimitedRepeatToLength",
+        "value": "The standard delivery service includes regular packaging and standard shipping times.",
+        "delimiterAlphabet": " ",
+        "seed": 42,
+        "length": "{size}"
+      },
+      "shared": true
+    },
+    {
       "id": "realWorldRegex.recitation.match.{size}",
       "axes": {
         "size": [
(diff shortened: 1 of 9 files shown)


cushon:
This was motivated by some investigation into real world regexes that showed performance gaps with RE2. So there are some real world patterns that would benefit from this, but there is also a tradeoff here with having more code and another fast path, and doing that analysis for patterns that don't benefit from it.

I think this is worth considering, from what I can tell the overhead is negligible in the cases that don't benefit from this, and it's seems to be a significant benefit for some real world patterns that the fast path does apply to.

But it's more of a tradeoff than some other smaller changes I've found like #541, #543, #544

eaftan (changes requested):
I wasn't able to reproduce the reported speedups. ☹️ I wonder if our baseline got faster in the meantime and made this optimization less impactful?

I ran the standard benchmark configuration:

```
  -p engine=SafeRE \
  -p patternName=stateCodes,multiWordPrefixAlternation,intlPhonePrefix,emailDomain \
  -p inputSize=1000 \
  -p match=false
```

Measured PR branch: e49dedc2027d300573ea9c477f84676273c3bad4
Measured PR base baseline: e87459babd99cce36630aa44073b1c83f8e4a984 plus only the benchmark files.

| Case | PR baseline SafeRE | PR optimized SafeRE | Speedup |
|---|---:|---:|---:|
| intlPhonePrefix | 2,845.887 ns/op | 2,572.106 ns/op | 1.11x |
| emailDomain | 2,722.658 ns/op | 2,568.333 ns/op | 1.06x |
| stateCodes | 30.311 ns/op | 31.674 ns/op | 0.96x, slower |
| multiWordPrefixAlternati | 30.262 ns/op | 30.479 ns/op | 0.99x, about flat/slower |

cushon:
I rebased this on top of latest changes.

I updated it to skip the fast path for start-anchored patterns, because we can quickly reject when scanning from the beginning anyways, so the fast path wasn't faster. That avoids the regression for `stateCodes` and `multiWordPrefixAlternation`.

For `emailDomain` I updated it to use `indexOf` if the analysis finds only one required literal, which is faster than the Aho-Corasick state machine, and results in a speedup for `emailDomain`.

I think this would still benefit from more analysis around tradeoffs, in general the Aho-Corasick fast path does add latency for patterns where it runs the scan and fails to reject the input, but I am still seeing some cases where it's helpful.

I'm curious if you can reproduce the results with the updated changes.

```
    -p patternName=stateCodes,multiWordPrefixAlternation,intlPhonePrefix,emailDomain \
    -p engine=SafeRE \
    -p match=true,false \
    -p inputSize=1000,10000
```

| Pattern | Input Size | Match? | Baseline (ns/op) | Optimized (ns/op) | Delta (ns) | Change | Speedup |
| :--- | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| `emailDomain` | 1,000 | false | 3,903.285 | 345.118 | -3,558.167 | **-91.2%** | **11.31x** |
| `emailDomain` | 1,000 | true | 193.252 | 200.866 | +7.614 | +3.9% | 0.96x |
| `emailDomain` | 10,000 | false | 35,480.214 | 3,265.406 | -32,214.808 | **-90.8%** | **10.87x** |
| `emailDomain` | 10,000 | true | 196.042 | 202.353 | +6.311 | +3.2% | 0.97x |
| `intlPhonePrefix` | 1,000 | false | 4,136.281 | 3,509.240 | -627.041 | **-15.2%** | **1.18x** |
| `intlPhonePrefix` | 1,000 | true | 275.882 | 220.166 | -55.716 | **-20.2%** | **1.25x** |
| `intlPhonePrefix` | 10,000 | false | 35,690.953 | 34,552.900 | -1,138.053 | -3.2% | 1.03x |
| `intlPhonePrefix` | 10,000 | true | 272.132 | 220.062 | -52.070 | **-19.1%** | **1.24x** |
| `multiWordPrefixAlternation` | 1,000 | false | 48.122 | 46.603 | -1.519 | -3.2% | 1.03x |
| `multiWordPrefixAlternation` | 1,000 | true | 166.980 | 164.428 | -2.552 | -1.5% | 1.02x |
| `multiWordPrefixAlternation` | 10,000 | false | 47.506 | 47.966 | +0.460 | +1.0% | 0.99x |
| `multiWordPrefixAlternation` | 10,000 | true | 163.759 | 164.571 | +0.812 | +0.5% | 1.00x |
| `stateCodes` | 1,000 | false | 46.935 | 48.660 | +1.725 | +3.7% | 0.96x |
| `stateCodes` | 1,000 | true | 51.255 | 52.144 | +0.889 | +1.7% | 0.98x |
| `stateCodes` | 10,000 | false | 48.079 | 48.062 | -0.017 | -0.0% | 1.00x |
| `stateCodes` | 10,000 | true | 51.166 | 50.978 | -0.188 | -0.4% | 1.00x |

eaftan (commented):
I still can't fully reproduce your numbers. Here's what I see after merging current main and applying some correctness fixes:

| Pattern | Size | Match | main ns/op | PR+fixes ns/op | PR/main | Interpretation |
|---|---:|:---:|---:|---:|---:|---|
| emailDomain | 1000 | false | 2010.425 | 290.580 | 0.145 | PR faster (6.92x) |
| emailDomain | 1000 | true | 108.111 | 114.322 | 1.057 | PR slower (1.06x) |
| emailDomain | 10000 | false | 19156.772 | 2785.726 | 0.145 | PR faster (6.88x) |
| emailDomain | 10000 | true | 109.131 | 113.917 | 1.044 | about flat |
| intlPhonePrefix | 1000 | false | 2161.764 | 2605.999 | 1.205 | PR slower (1.21x) |
| intlPhonePrefix | 1000 | true | 155.342 | 128.297 | 0.826 | PR faster (1.21x) |
| intlPhonePrefix | 10000 | false | 19548.265 | 25595.623 | 1.309 | PR slower (1.31x) |
| intlPhonePrefix | 10000 | true | 154.296 | 128.352 | 0.832 | PR faster (1.20x) |
| multiWordPrefixAlternation | 1000 | false | 32.811 | 33.943 | 1.035 | about flat |
| multiWordPrefixAlternation | 1000 | true | 94.789 | 97.161 | 1.025 | about flat |
| multiWordPrefixAlternation | 10000 | false | 33.308 | 33.735 | 1.013 | about flat |
| multiWordPrefixAlternation | 10000 | true | 94.784 | 98.823 | 1.043 | about flat |
| stateCodes | 1000 | false | 34.733 | 34.382 | 0.990 | about flat |
| stateCodes | 1000 | true | 173.047 | 167.644 | 0.969 | about flat |
| stateCodes | 10000 | false | 35.833 | 36.609 | 1.022 | about flat |
| stateCodes | 10000 | true | 167.882 | 166.602 | 0.992 | about flat |

At a high level, I see a solid win on `emailDomain` no-match, though not as large as reported. I also see the expected flat results for `multiWordPrefixAlternation` and `stateCodes`, and a win on `intlPhonePrefix` matches. The main discrepancy is `intlPhonePrefix` no-match, where I measure a regression rather than the small win you reported.

I did fix a couple of correctness issues in the prefilter. I also ran an ablation and verified that those fixes do not materially affect the benchmark results above, so I don't think they explain the discrepancy.

I think the remaining question is the tradeoff. The Aho-Corasick prefilter is clearly useful when it can reject early, as in `emailDomain` no-match. But it also adds latency when it scans and fails to reject, which seems to be what is happening for `intlPhonePrefix` no-match.

At this point I think I need to lean on your experience with real-world usage: do you think this is overall positive or negative performance-wise? I'm OK with the added complexity if you think this is an overall win.

cushon:
> I think the remaining question is the tradeoff. The Aho-Corasick prefilter is clearly useful when it can reject early, as in emailDomain no-match. But it also adds latency when it scans and fails to reject, which seems to be what is happening for intlPhonePrefix no-match.
>
> At this point I think I need to lean on your experience with real-world usage: do you think this is overall positive or negative performance-wise? I'm OK with the added complexity if you think this is an overall win.

Thanks, that all makes sense. I'm not sure about the real-world tradeoff for this one. The `intlPhonePrefix` example was adapted from a real-world regex that was an outlier in terms of the gap between SafeRE and RE2, but I don't have a sense of how often examples like that come up in practice and whether this is a net positive.

I think one way to decide that empirically would be to add some telemetry on how often this case helps, and maybe also the latency it adds when it doesn't. This is a bit related to #474, although for this I could just add a temporary patch I don't think this necessary needs permanent telemetry.

I'll try to report back when I have more data.

cushon:
I revisited this on top of the latest start acceleration changes, and I don't think Aho-Corasick is worthwhile for SafeRE.

* Since this PR was originally drafted, single required literal extraction and whole-string rejection landed separately via `RejectPrefilter` (#694, #695), and SafeRE's core Lazy DFA throughput has significantly improved.
* For multi-literal alternations with distinct leading characters (such as state codes `(?:AL|AK|...|WY)`), SafeRE's `AsciiBitmap` (`charClassPrefixAscii`) scans non-matching text faster than stepping through scalar Aho-Corasick transitions
* In unanchored regex searches, tracking match start positions and lengths across failure transitions introduces per-character array lookups that make scalar AC slower than SafeRE's Lazy DFA

I think the future of multi-literal start acceleration for SafeRE is a SIMD vector prefilter (e.g. Teddy) rather than a scalar state machine. I filed https://github.com/eaftan/safere/issues/700 for that.

The latest benchmark results (1,000-char `noMatch` inputs) were

| Benchmark | baseline | Aho-Corasick Branch | Speedup / Ratio |
|---|---|---|---|
| **`intlPhonePrefix`** | `1,943 ± 18 ns/op` | `1,943 ± 14 ns/op` | **1.00× (Parity)** |
| **`unanchoredStateCodes`** | `601 ± 11 ns/op` | `2,537 ± 8 ns/op` | **0.24× (4.2× slower)** |
| **`unanchoredLogKeywords`** | `1,922 ± 19 ns/op` | `2,542 ± 8 ns/op` | **0.76× (1.32× slower)** |
| **`unanchoredFruitKeywords`** | `1,956 ± 17 ns/op` | `2,544 ± 14 ns/op` | **0.77× (1.30× slower)** |

PR #541, opened 2026-07-01
Fix literal prefix acceleration fallback logic in Matcher

cushon:
When literal prefix acceleration is active (`literalPrefixCandidateStart = true`), `Matcher` performs a fast anchored forward DFA search at the candidate start position to verify it.

Previously, if this anchored check failed, the Matcher fell out of the DFA sandwich block completely and fell back to the backtracking `searchWithBitStateOrNfa()` engine.

This change fixes the fallback path by falling through to the reverse DFA search instead of the backtracking engine if the initial candidate verification fails.

This depends on #506 because the semantic guards removed in #506 would otherwise discard the reverse DFA start position on the motivating pattern, forcing a backtracking fallback anyway.

This new diff here compared to #506 is the handling of `literalPrefixCandidateStart` in `Matcher.java`, and the new `layoutBlock` benchmark.

```
./run-java-benchmarks.sh RealWorldRegexBenchmark.runBenchmark -- -p patternName=layoutBlock -p engine=SafeRE
```

| Input Size (Chars) | Matches Exist (`match`) | Before Score (ns/op) | After Score (ns/op) | Speedup Ratio | Status / Outcome |
| :---: | :---: | :---: | :---: | :---: | :--- |
| **1,000** | `false` (Non-matching) | 7,249 ± 66 | 7,039 ± 847 | **1.03x** | Comparable (Early DFA fail) |
| **1,000** | `true` (Sparse Match) | 19,704 ± 193 | 14,276 ± 378 | **1.38x** | **Improved** |
| **10,000** | `false` (Non-matching) | 71,482 ± 755 | 71,560 ± 799 | **1.00x** | Comparable (Early DFA fail) |
| **10,000** | `true` (Sparse Match) | 793,214 ± 43,234 | 111,828 ± 7,832 | **7.09x** | **7.1x Speedup** (DFA sandwich fallback avoided) |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -330,6 +330,18 @@
           "kind": "surroundWithSpaces",
           "body": "<Orange>"
         }
+      },
+      {
+        "name": "layoutBlock",
+        "pattern": "(?s)<block>\\n(\\s)*# (category|group)_defined:.*?\\n</block>",
+        "op": "replaceAllEmpty",
+        "match": "<block>\n  # category_defined: alpha, beta\n  entry: apple\n  entry: banana\n</block>\n",
+        "nonMatch": "<block>\n  # comment but not defined tag\n  entry: orange\n</block>\n",
+        "matchInput": {
+          "kind": "sparseMatch",
+          "nonMatchRepeats": 5,
+          "delimiterAlphabet": " "
+        }
       }
     ]
   },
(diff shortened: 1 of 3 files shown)


eaftan (approved):
LGTM.  I can repro your performance numbers:

| Input size | Match | Before ns/op | After ns/op | Speedup |
|---:|---:|---:|---:|---:|
| 1,000 | false | 2,740.428 ± 130.527 | 2,943.219 ± 192.566 | 0.93x |
| 1,000 | true | 15,569.799 ± 874.157 | 6,855.278 ± 312.981 | 2.27x |
| 10,000 | false | 27,189.795 ± 2,922.845 | 26,565.711 ± 1,552.643 | 1.02x |
| 10,000 | true | 405,977.384 ± 35,914.457 | 68,147.121 ± 3,372.391 | 5.96x |

PR #543, opened 2026-07-01
Optimize BitState capture resolution using relative indexing

cushon:
This change enables using BitState more often when resolving capture groups inside large documents.

SafeRE limits the use of the BitState engine to input texts smaller than a budget. When resolving capture groups during `resolveCaptures()`, we know the boundaries of the match, but `Matcher` was evaluating the `BitState` budget limit against the entire text length.

This change causes it to use BitState if the match region fits inside the budget, even if the entire input doesn't.

The included benchmark is ~1.46x faster at input size 100,000, which previous would have exceeded the BitState budget and used the NFA.

```
./run-java-benchmarks.sh --fastbuild RealWorldRegexBenchmark.runBenchmark -- -p patternName=jsonBlock -p engine=SafeRE
```

| Input Size | Match | Baseline (ns/op) | Optimized (ns/op) | Speedup / Diff | Engine Used |
| **1,000** | `true` | 7,252.925 | 7,235.610 | ~0% (Identical) | BitState (Both) |
| **1,000** | `false` | 131.183 | 131.804 | ~0% (Identical) | BitState (Both) |
| **10,000** | `true` | 168,945.130 | 169,283.090 | ~0% (Identical) | BitState (Both) |
| **10,000** | `false` | 1,013.979 | 1,013.143 | ~0% (Identical) | BitState (Both) |
| **100,000** | `true` | 758,974.353 | 524,129.778 | **1.45x** (**44.8% faster**) | Pike NFA $\rightarrow$ **BitState** |
| **100,000** | `false` | 9,833.918 | 9,834.113 | ~0% (Identical) | Pike NFA (Both) |


eaftan (dismissed):
I found some interesting behavior while reviewing this.

In the new benchmark, the JSON-like match region was truncated at smaller input sizes, so those cases were not actually matching even though they were intended to. I fixed that by making the match region scale with the input while preserving the intended shape of the workload.

While checking the fix, I initially could no longer reproduce the reported speedup. I did some exploration and found that the speedup appears when the input is large and the DFA-bounded match region is small-ish, but not too small. For input size 100,000, I measured:

| Region | Speedup |
|---:|---:|
| 128 | 1.06x |
| 1,024 | 1.15x |
| 5,033 | 1.61x |
| 16,384 | 2.13x |
| 50,000 | 1.02x |

This makes sense given the optimization: the DFA sandwich finds the match bounds, and then the capture-capable fallback runs inside that bounded region. This PR helps when that bounded region is small enough for BitState to be a better fallback choice, but large enough that the fallback work is meaningful. Your original benchmark had a 5,033-character match region, which explains why it showed a similar speedup.

I'm assuming this workload represents a real production case, and that the production match sizes are in the same rough range where this benefit appears.

I fixed the truncation issue and set the match region to 5% of the input size. With that, my benchmark results are close to yours:

| inputSize | match | PR ns/op | baseline ns/op | speedup |
|---:|---:|---:|---:|---:|
| 1,000 | true | 2,963 | 2,989 | 1.01x |
| 1,000 | false | 95.9 | 97.8 | 1.02x |
| 10,000 | true | 27,233 | 26,934 | 0.99x |
| 10,000 | false | 812.9 | 818.7 | 1.01x |
| 100,000 | true | 289,506 | 428,926 | 1.48x |
| 100,000 | false | 8,015 | 7,957 | 0.99x |

cushon:
Thanks for catching the issue with the benchmark!

> This PR helps when that bounded region is small enough for BitState to be a better fallback choice, but large enough that the fallback work is meaningful.
>
> I'm assuming this workload represents a real production case, and that the production match sizes are in the same rough range where this benefit appears.

Right that's my understanding, without this there's a performance cliff once the region exceeds the threshold for BitState. With this the benefit drops off gradually, for longer regions BitState is used for the suffix within the threshold, so the benefit drops off gradually the longer the region is.

The JSON regex was adapter from a real world example I was investigating because it was an outlier where SafeRE was relatively slow compared to RE2 c++. I don't know for sure that the typical inputs that regex is run on and whether this optimization is fixing the main bottleneck for this regex. But I think the fundamentals of it make sense and it doesn't require that much extra logic to pass the region bounds through, and it's nice to avoid the performance cliff.

eaftan (approved):

eaftan:
I could have sworn I approved this last night.  Thanks for the explanation, makes sense.

PR #544, opened 2026-07-01
Presize StringBuilders in Matcher replacement methods

cushon:
This updates `Matcher#replaceAll` and `replaceFirst` to presize the `StringBuilder` to match the input length.

This reduces the number of times the stringbuilder has to be resized for long inputs with many replacements. The replacements could make the output longer or shorter than the input, but starting with the input length seems like a reasonable guess.

For the included benchmark this seems to be a ~2% latency improvement but also significantly reduces allocations:

| Input Size | Matches | Baseline Alloc (B/op) | Optimized Alloc (B/op) | Memory Saved (B/op) | Allocation Reduction |
| **1,000** | 500 | 18,944.255 | 17,264.251 | **1,680 B** | **8.9%** |
| **10,000** | 5,000 | 214,130.589 | 170,266.542 | **43,864 B** (~44 KB) | **20.5%** |
| **100,000** | 50,000 | 1,990,299.004 | 1,700,288.881 | **290,010 B** (~290 KB) | **14.6%** |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -330,6 +330,19 @@
           "kind": "surroundWithSpaces",
           "body": "<Orange>"
         }
+      },
+      {
+        "name": "charReplace",
+        "pattern": "a",
+        "op": "replaceAllLiteral",
+        "match": "a",
+        "nonMatch": "b",
+        "matchInput": {
+          "kind": "repeat"
+        },
+        "nonMatchInput": {
+          "kind": "repeat"
+        }
       }
     ]
   },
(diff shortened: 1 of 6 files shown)


eaftan (approved):
Looks good, and I can repro the benchmark results.

Command used for both:

```
  -p patternName=charReplace \
  -p engine=SafeRE \
  -p match=true \
  -p inputSize=1000,10000,100000 \
  -prof gc
```

Reproduced allocation results:

| Input Size | Baseline B/op | Optimized B/op | Saved B/op | Reduction |
|---:|---:|---:|---:|---:|
| 1,000 | 18,944.137 | 17,264.134 | 1,680.003 | 8.9% |
| 10,000 | 214,129.392 | 170,265.343 | 43,864.049 | 20.5% |
| 100,000 | 1,990,285.969 | 1,700,277.507 | 290,008.462 | 14.6% |

Latency also reproduced: 2.2%, 3.4%, and 3.2% faster for the three sizes.

PR #545, opened 2026-07-01
Eagerly perform analysis in Pattern.compile

cushon:
This change eagerly initializes the DFA, OnePass, and reverse DFA in `Pattern.compile`.

Currently this initialization happens lazily when the pattern is first used. We noticed this in part comparing latency for individual regex operations between SafeRE and other engines, this approach means the first usage of a pattern is expensive, which increases the tail latency.

I think doing this initialization eagerly in `Pattern.compile` is the right thing to do for typical applications, not just because it affects metrics. For typical patterns in `static final` constants, this pays the cost during class initialization, and improves tail latency for initial requests after the application has started up.


eaftan (approved):
This seems like a reasonable tradeoff to me, and I agree with your rationale.  Let's ship it.

PR #546, opened 2026-07-02
Cache non-ASCII character class lookups

cushon:
When the DFA loop looks up character classes, it looks up classes for ASCII in `asciiClassMap`, and does a binary search on the boundaries array.

For patterns with non-ASCII character classes on inputs with non-ASCII characters, the binary search becomes a bottleneck.

#519 was another attempt to optimize this with a 128 KB `char[] bmpClassMap` for non-ASCII lookups, but that requires allocating a lot of memory, and is bad for memory locality because the entire array won't be in cache.

Instead of a lookup table for all possible BMP characters, this change creates a small cache-friendly hash table for character class lookups to reduce the amount of binary searching.

Comparing against baseline and #519, this appears to perform better than #519, likely because the hash table has a good hit rate and it's possible to keep it in cache.

```
    -p patternName=turnTitleWhitespaceCjk \
    -p engine=SafeRE \
    -p match=false \
    -p inputSize=1000
```

| Approach / Branch | JMH Score (ns/op) | Delta (ns) | Latency Reduction | Speedup | Memory Overhead |
| **Baseline**  | 17,000.285 ± 396.919 | *reference* | *reference* | *1.00x* | **0 bytes** |
| #519 | 15,452.880 ± 727.221 | -1,547.41 | -9.1% | **1.10x** | **131,072 bytes** (131 KB) |
| cjk-cache | 14,255.068 ± 1113.077 | -2,745.22 | **-16.1%** | **1.19x** | **1,536 bytes** (1.5 KB) |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -355,6 +355,19 @@
           "nonMatchRepeats": 5,
           "delimiterAlphabet": " "
         }
+      },
+      {
+        "name": "turnTitleWhitespaceCjk",
+        "pattern": "[ \\f\\r\\t\\v\\x{00a0}\\x{1680}\\x{2000}-\\x{200a}\\x{2028}\\x{2029}\\x{202f}\\x{205f}\\x{3000}\\x{feff}\\x{3164}]+",
+        "op": "replaceAllEmpty",
+        "match": "いくつもの\u3000中国語と日本語の文字\u3000がここにあります\u3000そして\u3000いくつかのスペース\u3000があります",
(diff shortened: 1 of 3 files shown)


eaftan (dismissed):
I see good performance from this, actually somewhat better than what you reported.

| Branch | Measured | Speedup vs Baseline |
|---|---:|---:|
| Baseline | 14,325.129 ± 175.094 ns/op | 1.00x |
| #519 | 11,093.653 ± 244.403 ns/op | 1.29x |
| #546 cjk-cache | 10,922.174 ± 276.876 ns/op | 1.31x |

Given that this performs slightly better than #519 and uses less memory, I'm happy to go with this one.

eaftan (approved):

PR #548, opened 2026-07-02
Preallocate the groups array in Matcher

cushon:
Preallocating a `groups` array per `Matcher` instance and then copying it as needed saves a bit of memory and slightly improves latency compared to allocating a new one every time.

I don't have super strong evidence this is an important bottleneck for real world applications, but it's also a simple change and I think the fundamentals of it make sense.

```
    -p patternName=turnTitleWhitespaceCjk \
    -p engine=SafeRE \
    -p match=false \
    -p inputSize=1000 \
    -prof gc
```

| Metric | Baseline (`cjk-baseline`) | Optimized (`preallocate-groups`) | Delta | Change (%) |
| **JMH Latency Score** | 17,656.795 ± 687.047 ns/op | 16,904.491 ± 316.517 ns/op | -752.30 ns | **-4.3%** (1.04x speedup) |
| **GC Alloc Rate (Norm)** | 7,512.248 B/op | 6,960.238 B/op | -552.01 B/op | **-7.3%** allocation reduction |
| **GC Alloc Rate (Total)** | 405.508 MB/sec | 392.251 MB/sec | -13.26 MB/sec | **-3.3%** allocation bandwidth |
| **GC Pause Count** | 10.000 counts | 10.000 counts | 0.00 | *flat* |
| **GC Pause Time** | 30.000 ms | 26.000 ms | -4.00 ms | **-13.3%** CPU GC time |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -355,6 +355,19 @@
           "nonMatchRepeats": 5,
           "delimiterAlphabet": " "
         }
+      },
+      {
+        "name": "turnTitleWhitespaceCjk",
+        "pattern": "[ \\f\\r\\t\\v\\x{00a0}\\x{1680}\\x{2000}-\\x{200a}\\x{2028}\\x{2029}\\x{202f}\\x{205f}\\x{3000}\\x{feff}\\x{3164}]+",
+        "op": "replaceAllEmpty",
+        "match": "いくつもの\u3000中国語と日本語の文字\u3000がここにあります\u3000そして\u3000いくつかのスペース\u3000があります",
+        "nonMatch": "いくつもの中国語と日本語の文字がここにありますそしていくつかのスペースがあります",
+        "matchInput": {
+          "kind": "repeat"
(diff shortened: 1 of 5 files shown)


cushon:
I also tacked on a change to have a singleton instance of `NoMatchResult` to save a few more allocations.

eaftan (approved):
I can repro the memory improvements.  Latency is basically unaffected; both results are within the confidence intervals.  I'd call this latency neutral, but a solid win for memory.  Let's ship it.

| Metric | Baseline: current main code | Optimized: merged PR branch | Delta |
|---|---:|---:|---:|
| Latency | 13,974.131 ± 1,720.663 ns/op | 14,344.858 ± 656.882 ns/op | +2.65% slower |
| Alloc norm | 7,512.196 B/op | 6,944.201 B/op | -7.56% |
| Alloc rate | 515.348 MB/sec | 461.689 MB/sec | -10.41% |
| GC count | 18 | 16 | -11.11% |
| GC time | 40 ms | 34 ms | -15.00% |

PR #547, opened 2026-07-02
Optimize replacement operations with DFA fast path

cushon:
This adds a fast path implementation of `replaceAll` and `replaceFirst` for DFA-eligible patterns, which makes a single loop over the input bypassing `find()`.

This also updates `doFindCore` to avoid an unnecessary second DFA pass for start-anchored patterns, which was noticed because the new replace loop uses the same approach.

This reduces allocations for replacement operations and bypasses some checks that are done on every `find()` call and instead does them once up-front.

This is a ~20% win for the example below. It does add some more code for the fast path and there's a complexity/performance tradeoff, but I think it's worth it in this case. We can rely on exhaustive crosscheck coverage to ensure replacement behavior doesn't diverge.

(One option to have even more test coverage would be to have a way to choose between this path and using `find()` for `replaceAll` in `EnginePathOptions`, and then have the tests compare the two implementations. But I'm not sure that's necessary.)

```
./run-java-benchmarks.sh --fastbuild RealWorldRegexBenchmark.runBenchmark -- \
    -p patternName=turnTitleWhitespace,turnTitleWhitespaceCjk \
    -p engine=SafeRE \
    -p match=false \
    -p inputSize=1000
```

| Approach / Branch | JMH Score (ns/op) | Delta (ns) | Latency Reduction | Speedup |
| :--- | :---: | :---: | :---: | :---: |
| **Baseline** | 16,966.642 ± 308.114 | *reference* | *reference* | *1.00x* |
| `replace-dfa-optimized` | 13,288.455 ± 521.461 | -3,678.19 | **-21.7%** | **1.28x** |

PR #549, opened 2026-07-02
Optimize split

cushon:
This is in the same vein as #547. It adds a separate loop to implement `split` that avoids calling `find()`, to minimize repeated work in `find()` that can be done once in the new loop, and also optimizes allocations.

This is a 20% latency win for `split` on the long inputs in the benchmark below.

```
./run-java-benchmarks.sh --fastbuild Issue481ScalingBenchmark.splitWords_safere
```

| Input Text Size (chars) | Baseline Latency (us/op) | Optimized Latency (us/op) | Delta (us) | Latency Reduction | Speedup |
| **128** | 2.996 ± 0.117 | 2.530 ± 0.075 | -0.466 | **-15.6%** | **1.18x** |
| **1,024** | 22.730 ± 1.743 | 20.754 ± 0.773 | -1.976 | **-8.7%** | **1.10x** |
| **10,240** | 231.492 ± 13.021 | 226.054 ± 14.337 | -5.438 | **-2.3%** | **1.02x** |
| **102,400** | 2,480.095 ± 95.703 | 1,927.739 ± 45.506 | -552.356 | **-22.3%** | **1.29x** |
| **1,048,576** (1 MB) | 23,969.911 ± 834.730 | 19,770.121 ± 480.684 | -4,199.790 | **-17.5%** | **1.21x** |

### DIFF
--- a/safere/src/main/java/org/safere/Matcher.java
+++ b/safere/src/main/java/org/safere/Matcher.java
@@ -969,6 +969,18 @@ public final class Matcher implements MatchResult {
     return prog != null && !prog.hasGraphemeSemantics() && prog.numLoopRegs() == 0;
   }
 
+  private boolean canUseForwardDfa() {
+    return enginePathOptions().dfa() && dfaSupportsProgram(parentPattern.flatDfaProg());
+  }
+
+  private boolean canUseReverseDfa() {
+    return enginePathOptions().dfa()
+        && enginePathOptions().reverseDfa()
+        && dfaSupportsProgram(parentPattern.flatReverseDfaProg())
+        && !parentPattern.prog().anchorStart()
+        && reverseDfa() != null;
+  }
+
   /**
    * Resets this matcher and then attempts to find the next subsequence of the input that matches
    * the pattern, starting at the specified index.
@@ -1238,11 +1250,7 @@ public final class Matcher implements MatchResult {
     int[] requiredRanges = parentPattern.requiredMatchClassRanges();
     boolean hasAcceleratedSearchPath =
         (parentPattern.prefix() != null)
-            || (options.reverseDfa()
-                && dfaSupportsProgram(parentPattern.flatReverseDfaProg())
-                && prog.anchorEnd()
-                && !prog.anchorStart()
-                && text.length() >= MIN_REVERSE_FIRST_LEN);
+            || (canUseReverseDfa() && prog.anchorEnd() && text.length() >= MIN_REVERSE_FIRST_LEN);
     if (options.charClassMatchFastPaths()
         && requiredRanges != null
         && !hasAcceleratedSearchPath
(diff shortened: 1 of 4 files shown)


safere/src/main/java/org/safere/Matcher.java
```diff
@@ -2801,4 +2806,173 @@ private void checkMatch() {
...
     }
   }
+
+  static final class SplitBuffer {
+    int[] array = new int[32];
+    int size = 0;
+
+    void add(int start, int end) {
+      if (size + 2 > array.length) {
+        int[] newArray = new int[array.length * 2];
```
eamonnmcmanus: I think you could use `Arrays.copyOf` here?

safere/src/main/java/org/safere/Matcher.java
```diff
@@ -2801,4 +2806,173 @@ private void checkMatch() {
...
+    String literal = parentPattern.literalMatch();
+    if (options.literalFastPaths() && literal != null && parentPattern.numGroups() == 0) {
+      int idx =
+          parentPattern.prefixFoldCase()
+              ? indexOfIgnoreCase(text, literal, fromIndex)
+              : text.indexOf(literal, fromIndex);
+      if (idx < 0) {
+        return -1L;
+      }
+      return ((long) idx << 32) | ((idx + literal.length()) & 0xFFFFFFFFL);
```
eamonnmcmanus: This pattern `((long) foo << 32) | (bar & 0xFFFFFFFFL)` occurs often enough that it might be worth introducing a helper method. I assume the JIT would inline it.

safere/src/main/java/org/safere/Matcher.java
```diff
@@ -2801,4 +2806,173 @@ private void checkMatch() {
       }
     }
   }
+
+  static final class SplitBuffer {
+    int[] array = new int[32];
```
eamonnmcmanus, line 2803: Since this is referenced from another class, I think it might be worth adding a doc comment saying that it consists of alternating (start, end) values.

eamonnmcmanus (commented):
A few nits.

safere/src/main/java/org/safere/Matcher.java
cushon: Done

safere/src/main/java/org/safere/Matcher.java
cushon: Done

safere/src/main/java/org/safere/Matcher.java
cushon: Done

eaftan (approved):
I made a few fixes, and I'm able to reproduce the performance improvement.

Baseline: current main at a1f0e9c
Experiment: PR branch at 6fcf249

Command for both:

```bash
./run-java-benchmarks.sh --fastbuild Issue481ScalingBenchmark.splitWords_safere
```

| Text size | main us/op | PR + fixes us/op | Reduction | Speedup |
|---:|---:|---:|---:|---:|
| 128 | 2.222 ± 0.122 | 1.842 ± 0.088 | 17.1% | 1.21x |
| 1,024 | 16.331 ± 1.997 | 12.951 ± 0.460 | 20.7% | 1.26x |
| 10,240 | 156.285 ± 11.716 | 130.010 ± 4.588 | 16.8% | 1.20x |
| 102,400 | 1721.022 ± 198.464 | 1321.197 ± 49.322 | 23.2% | 1.30x |
| 1,048,576 | 17592.544 ± 902.607 | 14484.759 ± 618.292 | 17.7% | 1.21x |

PR #550, opened 2026-07-02
Generalize `charClassPrefixAscii` to support non-ASCII character class prefixes

cushon:
This generalizes `charClassPrefixAscii` to support non-ASCII character class prefixes.

For patterns with non-ASCII character class prefixes, this allows the fast path to find the first potential match in the input and then skip over the non-matching prefix before starting to scan with the DFA, instead of having to run the DFA over the non-matching prefix. This is a latency win for patterns that start with a non-ASCII character class, and preserves existing performance for the ASCII character class fast path.

```
    -p patternName=unicodeClassPrefix \
    -p engine=SafeRE,JDK \
    -p match=true \
    -p inputSize=100000
```

| Engine | Baseline Latency (ns/op) | Optimized Latency (ns/op) | Delta (ns) | Latency Reduction | Speedup |
| **SafeRE** | 3,562.752 ± 34.566 | 2,356.671 ± 8.078 | -1,206.081 | **-33.9%** | **1.51x** |
| **JDK** | 794.870 ± 6.744 | 798.454 ± 9.126 | +3.584 | *neutral* | *1.00x* |

### DIFF
--- a/.agents/skills/safere-pr-review-scout/SKILL.md
+++ b/.agents/skills/safere-pr-review-scout/SKILL.md
@@ -1,0 +1,419 @@
+---
+name: safere-pr-review-scout
+description: "Run a serialized background sweep of open SafeRE GitHub PRs: skip drafts, track reviewed PR head SHAs and discussion changes, assess PR intent against implementation, run the review-fix-loop skill for P2+ findings, reproduce optimization benchmark claims against current main, and write durable reports and artifacts without pushing or commenting."
+---
+
+# SafeRE PR Review Scout
+
+## Goal
+
+Prepare the data needed for a human SafeRE PR review while the reviewer is away:
+
+- which open non-draft PRs need attention;
+- whether each PR's idea makes sense and matches its implementation;
+- P2+ code-review findings fixed locally with `$review-fix-loop`;
+- benchmark reproduction for optimization PRs;
+- durable reports and artifacts that can be inspected later.
+
+Do not push branches, post PR comments, close issues, or publish review text unless the user
+explicitly asks.
+
+## Required Inputs And Defaults
+
+Default repository: `/home/eaftan/safere`.
+
+Default storage root: `$HOME/.codex/safere-pr-review`.
+
+Default base branch: `origin/main`, refreshed before each sweep.
+
+Default review threshold: P2 or higher.
+
+Trusted PR authors are hardcoded to GitHub logins:
+
+- `cushon`
+- `eamonnmcmanus`
+- `kluever`
+
+Ignore PRs from other authors. Do not read their PR body, comments, reviews, linked issues, diffs,
+or code, and do not check out their branches. Record their PR number, URL, author login, and a note
+that the author is not on the allowlist in the output report so a human can decide whether to add
+the contributor. The helper script enforces this allowlist in `discover-prs`; use that command for
+PR discovery instead of calling `gh pr list` directly.
+
+Use current PR head SHA as the primary freshness key. Review a PR again when its head SHA changed,
+its `updatedAt` is newer than the last processed value, the recorded base SHA is older than current
+`origin/main`, or the user explicitly asks for a forced review. The base-SHA check is required for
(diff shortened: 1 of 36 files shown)


eaftan (commented):
Pushed a small fix to this PR branch.

I can't repro the performance improvement on this one. I'm seeing essentially flat performance on the benchmark you shared.  Would you mind re-running against current main and seeing if you still see the speedup?

Command: 

```bash
./run-java-benchmarks.sh '^org\.safere\.benchmark\.RealWorldRegexBenchmark\.runBenchmark$' -- -p patternName=unicodeClassPrefix -p engine=SafeRE,JDK -p match=true -p inputSize=100000
```

| Engine | main ns/op | PR+fixes ns/op | PR/main | Interpretation |
|---|---:|---:|---:|---|
| SafeRE | 2030.209 | 2056.535 | 1.013 | about flat / PR slightly slower |

cushon:
I can't reproduce a speedup for those benchmark cases anymore. There's some noise in the results and it's possible I was mislead by that, I think the other thing that happened is that the DFA has gotten faster so for DFA-eligible patterns there's not much benefit to this fast path.

There are cases where it could apply to DFA-ineligible patterns, and I do still see a speedup for those cases, I added a `unicodeClassPrefixNfaFallback` benchmark that shows an improvement.

I'm not sure how often this fast path will apply at this point. I think it's cheap enough if it doesn't apply that it still might make sense, but it's another thing that I could try to get more data on.

```
    -p patternName=unicodeClassPrefix,turnTitleWhitespaceCjk \
    -p engine=SafeRE \
    -p match=true,false \
    -p inputSize=100000
```

| PatternName | Match | Baseline (ns/op) | Optimized (ns/op) | Ratio (Exp/Base) |
| turnTitleWhitespaceCjk | false | 695,302.5 | 691,996.9 | 1.00x |
| turnTitleWhitespaceCjk | true | 934,675.9 | 920,318.1 | **0.98x** |
| unicodeClassPrefix | false | 57,125.4 | 58,998.9 | 1.03x |
| unicodeClassPrefix | true | 2,674.4 | 2,649.7 | **0.99x** |

```
    -p patternName=unicodeClassPrefixNfaFallback \
    -p engine=SafeRE \
    -p inputSize=100000

```

| Match | Baseline (`unicode-prefix-baseline`) | Optimized (`first-class-acceleration`) | Speedup (vs Baseline) |
| :---: | :---: | :---: | :---: |
| **`false`** | 5,701,564.8 ns/op (~5.7 ms) | 522,451.6 ns/op (~0.52 ms) | **10.91x** |
| **`true`** | 11,453,782.5 ns/op (~11.5 ms) | 11,674,791.8 ns/op (~11.7 ms) | **0.98x** (neutral) |

eaftan (approved):
I reran the benchmark cases from your comments after my last review.

Benchmark results:

| Benchmark | main library ns/op | PR ns/op | Speedup | Interpretation |
|---|---:|---:|---:|---|
| unicodeClassPrefix match=true | 1733.132 | 2307.047 | 0.75x | PR slower |
| unicodeClassPrefix match=false | 49463.000 | 48576.978 | 1.02x | about flat |
| turnTitleWhitespaceCjk match=true | 756095.224 | 722609.962 | 1.05x | about flat |
| turnTitleWhitespaceCjk match=false | 623909.339 | 614872.513 | 1.01x | about flat |
| unicodeClassPrefixNfaFallback match=true | 9134716.060 | 9365773.647 | 0.98x | about flat |
| unicodeClassPrefixNfaFallback match=false | 4220785.264 | 254454.148 | 16.59x | PR faster |

The new `unicodeClassPrefixNfaFallback` no-match result reproduces strongly for me, even larger than the ~10.9x you reported. The match=true fallback row is flat, and `turnTitleWhitespaceCjk` is also flat. There's a moderate regression in `unicodeClassPrefix match=true`, but that may just be a small-benchmark sensitivity issue.

I think this is good to merge given the large win on `unicodeClassPrefixNfaFallback match=false` and essentially flat performance on the other benchmarks

Issue #551, opened 2026-07-02
ThreadLocals and strategy for pooling heavyweight objects


cushon:
SafeRE uses `ThreadLocal` for pooling objects like the DFA cache.

With Virtual Threads, `ThreadLocal`s can be expensive if there is a large number of threads. (There is some related discussion of ThreadLocals in https://openjdk.org/jeps/444 and https://openjdk.org/jeps/506.)

One alternative could be to have a global pool, maybe based on `ConcurrentLinkedQueue`, to allow sharing the DFA cache across physical and virtual threads. That would add some overhead to retrieving a DFA cache from the pool compared to the ThreadLocal, but I think the cost would be negligible compared to the cost of typical regex operations, and might be the right approach especially in the context of Virtual Threads.

(I don't think I've run into the current use of ThreadLocals being an issue in practice, I noticed this investigating a different issue that I think is related to classes using a `Pattern` constant that are getting reloaded in new classloaders, which cases the pattern to get recompiled and wipes out the DFA cache. Having a global cache for compiled `Pattern`s might be useful in that situation. I'm not sure if it's worth it. It also isn't something that necessarily belongs in SafeRE, vs. something clients could implement at a higher level if they wanted it. That also allows clients to choose their own caching library like caffeine without having to add a dependency to SafeRE.)

cushon:
https://github.com/eaftan/safere/pull/552 attempted to migrate to lock-free bounded object pools based on `AtomicReferenceArray`, but had worse performance than `ThreadLocal`.

Capturing a comment from that PR for posterity:

> If there ends up being a performance tradeoff for single-threaded use-cases, one option would be to have both implementations and a flag to choose which cache to use. But that can probably wait until someone wants to actually evaluate that flag.
>
> Also IIUC the JDK has an internal per-carrier thread cache (discussed in [FasterXML/jackson-core#919](https://redirect.github.com/FasterXML/jackson-core/issues/919) which is about preparing a similar use of Thread Locals to prepare for Virtual Threads). The per-carrier thread cache isn't a public API, I think in part because it'd often be the wrong choice compared to scoped values or something else, but I think this is a case where it's exactly what we'd want.

The linked FasterXML/jackson-core issue might be a good example to follow here in general.

PR #552, opened 2026-07-02
Replace ThreadLocal regex engine caches with lock-free bounded ArrayPool caches

cushon:
Replaces the thread-local caches for BitState, Nfa, and Dfa instances in Pattern with lock-free bounded object pools based on AtomicReferenceArray.

https://github.com/eaftan/safere/issues/551


cushon:
There's no rush for this, I haven't run into issues in practice with SafeRE and VTs yet, but I think something like this will be desirable eventually to support using SafeRE with Virtual Threads.

eaftan (commented):
The virtual-thread motivation here makes sense, but I found several engine-lifecycle issues: one UTF-8 path no longer compiled, and ordinary find, split, optimized replacement, and deferred-capture paths could retain borrowed engines. The pools also eagerly constructed engines for every new Pattern, and disabled-DFA UTF-8 diagnostics reported a DFA search that never happened.  I've pushed commits that fix those issues and add regression coverage. 

There are some significant regressions in certain benchmarks:

| Benchmark | main (ns/op) | PR + fixes (ns/op) | Change |
|---|---:|---:|---:|
| `MatcherApiBenchmark.lookingAt@safere-string` | 91.460 | 89.821 | neutral |
| `MatcherApiBenchmark.lookingAt@safere-utf8` | 80.370 | 80.444 | neutral |
| `MatcherApiBenchmark.regionFind@safere-string` | 229.404 | 257.860 | 12.4% slower |
| `MatcherApiBenchmark.regionFind@safere-utf8` | 262.273 | 287.009 | 9.4% slower |
| `MatcherApiBenchmark.resetAndFind@safere-string` | 220.327 | 300.120 | 36.2% slower |
| `MatcherApiBenchmark.resetAndFind@safere-utf8` | 248.674 | 333.442 | 34.1% slower |

I checked to make sure that the correctness fixes are not responsible for the regressions.  Profiling identifies the `AtomicReferenceArray` acquire/release CAS operations as a substantial part of the affected path. 

Since this is not urgent and there is not yet a concurrent or virtual-thread benchmark quantifying the intended benefit, I think we should wait to merge until we either reduce the single-thread overhead or have representative virtual-thread evidence that justifies the tradeoff.

cushon:
Thanks for evaluating this, sorry it was half baked. Shelving this until there's more performance experience with uses of SafeRE and Virtual Threads sounds good to me.

If there ends up being a performance tradeoff for single-threaded use-cases, one option would be to have both implementations and a flag to choose which cache to use. But that can probably wait until someone wants to actually evaluate that flag.

Also IIUC the JDK has an internal per-carrier thread cache (discussed in [FasterXML/jackson-core#919](https://redirect.github.com/FasterXML/jackson-core/issues/919) which is about preparing a similar use of Thread Locals to prepare for Virtual Threads). The per-carrier thread cache isn't a public API, I think in part because it'd often be the wrong choice compared to scoped values or something else, but I think this is a case where it's exactly what we'd want.

cushon:
I summarized some of this back to #551, I'm going to close this out since this PR is now very out of date and the approach has issues.

My current best idea here is still to keep something like this approach, but keep the `ThreadLocal` approach and put it behind a flag, so anything using VTs who runs into issues with `ThreadLocal` could evaluate the new approach without making it the default in cases that don't need it and where it's slower.

PR #554, opened 2026-07-03
Avoid temporary String allocations for group references in Matcher.appendReplacement

cushon:
This avoids temporary `String` allocations for group references in `Matcher.appendReplacement`. Instead of creating a substring with `group(N)` and appending it via `StringBuilder#append(String)`, it retrieves the group bounds directly and uses `StringBuilder#append(CharSequence, int, int)` to write from the input string directly to the `StringBuilder`.

It also adds a fast-path for simple replacements (without `$` or `\`), bypassing the character-by-character parsing loop entirely.

This avoids some allocations, and also improves latency:

```
./run-java-benchmarks.sh --fastbuild 'ReplaceBenchmark\.manualReplaceAll_safere'
```

| Benchmark | Baseline  | `string-ops-optimization` | Latency Reduction | Speedup |
| `ReplaceBenchmark.manualReplaceAll_safere` | `3911.197 ± 331.893` ns/op | `3481.292 ± 60.404` ns/op | **-11.0%** (save ~430 ns) | **1.12x** |

### DIFF
--- a/safere-benchmarks/src/main/java/org/safere/benchmark/ReplaceBenchmark.java
+++ b/safere-benchmarks/src/main/java/org/safere/benchmark/ReplaceBenchmark.java
@@ -267,4 +267,39 @@ public class ReplaceBenchmark {
   public String emptyReplaceAll_re2ffm() {
     return re2ffmEmpty.matcher(emptyText).replaceAll(emptyReplacement);
   }
+
+  // ===== Manual appendReplacement loop (for group reference performance) =====
+
+  @Benchmark
+  public String manualReplaceAll_safere() {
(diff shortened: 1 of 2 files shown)


eaftan (approved):
This seems reasonable, and I can reproduce the performance improvement.  I fixed a small issue as well.

Command:

```bash
./run-java-benchmarks.sh '^org\.safere\.benchmark\.ReplaceBenchmark\.manualReplaceAll_safere$'
```

| Benchmark | main ns/op | PR+fixes ns/op | PR/main | Interpretation |
|---|---:|---:|---:|---|
| ReplaceBenchmark.manualReplaceAll_safere | 2622.872 | 2435.508 | 0.929 | PR faster (1.08x) |

Issue #490, opened 2026-06-20
Consider optimizing safere-ffm-re2 with critical calls


cushon:
The TL;DR here is that it might be interesting to include a version of RE2-FFM in the benchmarks that uses critical calls to avoid copying the input data from the managed heap to native, but also doing that for production workloads is probably a bad idea and so I don't think the current benchmarks are sandbagging by not doing it.

I think the current safere-ffm-re2 wrapper around RE2 is a pretty well optimized and fair usage of FFM to call RE2, and is similar to what a production implementation would look like.

The main performance challenge with calling RE2 from FFM is the ['Find-in-Text Scaling' issue discussed in BENCHMARKS.md](https://github.com/eaftan/safere/blob/main/BENCHMARKS.md#find-in-text-scaling-%C2%B5sop-lower-is-better), that for repeated matches in a loop the FFM shim copies the entire input data to native every time.

There are some ways to optimize this, but they have drawbacks:

1. The implementation of `Matcher` could copy the data to the native heap once, but then that native resource has to be managed somehow.
    * `Matcher` could use cleaners/finalizers, but then there's an expensive native allocation that may not get cleaned up eagerly, so if there are a lot of allocations it runs into the usual problems managing expensive native resources with cleaners/finalizers.
    * Another option would be to require explicit resource management for `Matcher` by having it implement `AutoCloseable` and having the client close matchers, but that complicates the API and makes migration from `java.util.regex` difficult.
2. The FFM shim could use 'critical' calls with `Linker.Option.critical(true)` which allow it to pass arrays from the Java heap to native using `MemorySegment.ofArray(inputUtf8)` without a copy, which is cheap, but means that the managed heap has to be pinned while the native call happens. In JDK 22 and newer G1 supports pinning regions instead of the entire heap ([JEP 423](https://openjdk.org/jeps/423)), but pinning is still expensive and if there are a of critical calls or long-running critical calls it will affect application performance.

So using either of those approaches would achieve some benchmark wins for the RE2-FFM benchmarks and might be interesting as a point of comparison. But also there are significant drawbacks to using either approach for production workloads, and a major benefit of SafeRE is that it doesn't require balancing any of these tradeoffs because all of the memory can stay on the managed heap.

[...]

PR #557, opened 2026-07-04
Improve the performance for the RE2 FFM wrapper.

cushon:
Improve the performance for the RE2 FFM wrapper.

* Replace flat byte-to-char mapping array (which required O(N) scans for reverse lookups) with mapping array that indexes from characters to bytes, and does binary search for byte-to-character lookups, eliminating a bottleneck on long strings.
* Add `re2_find_all()` to collect all non-overlapping match boundaries in a single linear pass, and use it to implement `split()`.

This is sort of related to #490, but unlike critical calls these optimizations don't come up any real tradeoffs, and make the comparison to RE2-FFM a bit closer to a production solution.

The following benchmark shows performance of `split()` on a long string, SafeRE is still coming out ahead here but this makes the RE2-FFM performance less pathological. (The SafeRE numbers here were also stacked on top of other changes including #549 to may not reproduce exactly against `main`.)

```
./run-java-benchmarks.sh --fastbuild '^org\.safere\.benchmark\.Issue481ScalingBenchmark\.split.*_(safere|re2ffm)$' -- -p textSize=102400,1048576
```

| Input Size | SafeRE (us/op) | Baseline FFM (us/op) | Optimized FFM (us/op) | FFM Optimization Speedup | Optimized FFM vs SafeRE |
| **100KB** (`102400`) | 1,928.49 | 42,887.64 | 4,679.21 | **9.17x** | 2.42x slower |
| **1MB** (`1048576`) | 19,019.42 | 5,186,341.35 | 43,453.90 | **119.35x** | 2.28x slower |

### DIFF
--- a/safere-ffm-re2/pom.xml
+++ b/safere-ffm-re2/pom.xml
@@ -21,6 +21,15 @@
     <maven.deploy.skip>true</maven.deploy.skip>
   </properties>
 
+  <dependencies>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13.2</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
   <build>
     <plugins>
       <!-- Build native shim library via cmake -->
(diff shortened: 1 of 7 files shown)


eaftan (approved):
This seems reasonable to me.  I made a few fixes and pushed them to this PR branch:

- Align the new FFM tests with project test style and add the required license header.
- Advance native zero-width `find_all` matches from the match position, not the previous search lower bound.
- Cap split match-buffer sizing by input size and account for the leading empty match that Java split ignores when `limit > 0`.
- Keep the UTF-8/UTF-16 offset map lazy so APIs that do not need offset conversion do not pay an extra full-input pass.

PR #560, opened 2026-07-04
Optimize replacement operations with OnePass fast path

cushon:
This adds a fast path implementation of `replaceAll` and `replaceFirst` for `OnePass`-eligible patterns, which makes a single loop over the input bypassing `find()`.

This PR is currently stacked on top of #547, it is a complementary optimization. See also the comment in https://github.com/eaftan/safere/pull/547#issuecomment-4883237869.

This is a tradeoff, `OnePass` is significantly faster if there are any replacements, but if there are no matches it is slower than the DFA to reject inputs. I think this is the same tradeoff SafeRE is making for methods like `find()` and `match()`, and this optimization uses the same `ONEPASS_ANCHORED_TEXT_LIMIT` threshold to skip `OnePass` for longer inputs.

```
  -p patternName=fruitMarkupTag \
  -p engine=SafeRE
```

| Input Size | Match? | `replace-dfa-optimized` (ns/op) | `onepass-replace` (ns/op) | Impact / Speedup |
| **1,000** | `true` (Match) | 17,822.982 ± 117.086 | 5,816.535 ± 61.866 | **3.06x faster** |
| **1,000** | `false` (Reject) | 1,518.945 ± 8.419 | 1,747.930 ± 9.225 | 1.15x slower |
| **10,000** | `true` (Match) | 171,449.961 ± 855.264 | 56,662.342 ± 558.625 | **3.03x faster** |
| **10,000** | `false` (Reject) | 14,505.678 ± 140.164 | 16,659.245 ± 163.770 | 1.15x slower |
| **100,000** | `true` (Match) | 1,729,575.841 ± 12,071.998 | 566,436.737 ± 2,722.482 | **3.05x faster** |
| **100,000** | `false` (Reject) | 140,874.367 ± 652.474 | 168,235.415 ± 529.088 | 1.19x slower |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -424,6 +424,19 @@
         "nonMatchInput": {
           "kind": "repeat"
         }
+      },
+      {
+        "name": "onePassReplace",
+        "pattern": "http://[a-z]+/([a-z]+)",
+        "op": "replaceAllGroup1",
+        "match": "http://foo/bar",
+        "nonMatch": "hello world",
+        "matchInput": {
+          "kind": "repeat"
+        },
+        "nonMatchInput": {
+          "kind": "repeat"
+        }
       }
     ]
   },
--- a/safere-benchmarks/src/main/java/org/safere/benchmark/RealWorldRegexBenchmark.java
+++ b/safere-benchmarks/src/main/java/org/safere/benchmark/RealWorldRegexBenchmark.java
@@ -136,6 +136,7 @@ public class RealWorldRegexBenchmark {
     "jsonBlock",
     "charReplace",
     "greedyOnePass",
+    "onePassReplace",
     "layoutBlock",
     "turnTitleWhitespaceCjk",
     "cjkSearch",
--- a/safere/src/main/java/org/safere/Matcher.java
+++ b/safere/src/main/java/org/safere/Matcher.java
@@ -3268,6 +3268,7 @@ public final class Matcher implements MatchResult {
     private final String replacement;
     private final int maxGroup;
     private ReplacementSegment[] value;
+    private Boolean needsCaptures;
 
     LazyTemplate(String replacement, int maxGroup) {
       this.replacement = replacement;
@@ -3275,6 +3276,13 @@ public final class Matcher implements MatchResult {
     }
 
     boolean needsCaptures() {
+      if (needsCaptures == null) {
+        needsCaptures = computeNeedsCaptures();
+      }
+      return needsCaptures;
+    }
+
+    private boolean computeNeedsCaptures() {
       if (replacement == null) {
         return false;
       }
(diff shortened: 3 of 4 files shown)


eaftan (dismissed):
I pushed fixes for the replacement state/validation issues I found.

I can repro the performance improvements, but not to the same degree.  You reported the matching cases improving by 3x, but I'm seeing around 1.8x.  

| Benchmark | main ns/op | PR+fixes ns/op | PR/main | Interpretation |
|---|---:|---:|---:|---|
| fruitMarkupTag inputSize=1000 match=true | 8565.674 | 4758.375 | 0.556 | PR faster (1.80x) |
| fruitMarkupTag inputSize=1000 match=false | 1001.006 | 1197.933 | 1.197 | PR slower (1.20x) |
| fruitMarkupTag inputSize=10000 match=true | 83930.855 | 45540.722 | 0.543 | PR faster (1.84x) |
| fruitMarkupTag inputSize=10000 match=false | 9565.384 | 11731.589 | 1.226 | PR slower (1.23x) |
| fruitMarkupTag inputSize=100000 match=true | 853227.403 | 461289.597 | 0.541 | PR faster (1.85x) |
| fruitMarkupTag inputSize=100000 match=false | 95435.982 | 115123.875 | 1.206 | PR slower (1.21x) |

The direction matches your report: matching cases are significantly faster, and no-match cases are slower by about the amount you called out.

I need to lean on your judgment about whether this is a performance win overall based on real-world usage. I'm approving on the assumption that replacement workloads with captures are match-heavy enough for this to be worthwhile; if you agree, feel free to merge.

eaftan (dismissed):

cushon:
> I think this is the same tradeoff SafeRE is making for methods like `find()` and `match()`

One thing I missed earlier is that `find()` and `match()` do this for start-anchored patterns and those methods just find a single match and then stop. For `replaceAll` we have to scan the entire input to find all matches. For longer inputs with dense matches OnePass might still be faster, but for sparse matches OnePass is slower than using the DFA to find the match ranges. So there is a tradeoff here that doesn't exist for the use of OnePass in `find()` and `match()`, and I'm not sure the OnePass fast path for replacement methods is a good idea.

> I need to lean on your judgment about whether this is a performance win overall based on real-world usage.

Given my under understanding of the tradeoffs here I'm not sure this is a good idea, I'll try to get more data and revisit.

The data I have shows that SafeRE performance is relatively worse for `replaceAll` and `replaceFirst` compared to RE2 than other methods, but I want to revisit with the other recent optimizations and the DFA fast path for replacement methods in #547 and see where things stand.

If there's still a gap, I'll try to get telemetry on whether a OnePass fast path would be a net win.

cushon:
I realized some of the benefits of this can be incorporated into the DFA fast path in https://github.com/eaftan/safere/pull/547#issuecomment-4886656742, and in the cases where doing OnePass first is better I don't think this is worth the tradeoff. I'm closing this out.

PR #563, opened 2026-07-06
Avoid unnecessary third DFA pass for start-anchored patterns

cushon:
This split outs a small performance fix that was included in #547 that wasn't related to the DFA replacement fast path.

### DIFF
--- a/safere/src/main/java/org/safere/Matcher.java
+++ b/safere/src/main/java/org/safere/Matcher.java
@@ -1489,18 +1489,14 @@ public final class Matcher implements MatchResult {
       int earlyEnd = fwdResult.pos();
 
       if (prog.anchorStart()) {
-        // Anchored: match start is effectiveStart. Run forward DFA (anchored, first-match) to find
-        // actual end, then defer inner captures until requested.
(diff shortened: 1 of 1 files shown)


eaftan (approved):
I tested start-anchored `find`/`replaceAll` workloads where this should matter, plus HTTP as a small-input sanity check.  This makes sense and I see solid speedups on `fruitMarkupTag match=true`:

| Benchmark | Baseline ns/op | PR #563 ns/op | Speedup | Interpretation |
|---|---:|---:|---:|---|
| fruitMarkupTag inputSize=10000 match=true | 77,615.393 | 59,132.341 | 1.31x | PR faster |
| fruitMarkupTag inputSize=10000 match=false | 6,615.309 | 6,551.342 | 1.01x | flat |
| fruitMarkupTag inputSize=100000 match=true | 784,626.970 | 639,357.680 | 1.23x | PR faster |
| fruitMarkupTag inputSize=100000 match=false | 65,089.794 | 65,090.135 | 1.00x | flat |
| HttpBenchmark.httpFull_safere | 375.635 | 366.700 | 1.02x | flat / tiny faster |
| HttpBenchmark.httpExtract_safere | 364.922 | 366.500 | 1.00x | flat |

PR #567, opened 2026-07-12
Add static and runtime regex diagnostics

### LINKED ISSUE #474: Add lightweight SafeRE diagnostics listener for engine selection

eaftan:
## Summary

- add immutable static pattern analysis through `Pattern.analysis()`;
- add a process-wide synchronous runtime diagnostics listener with compilation and operation events;
- report authoritative boundary strategy, capture strategy/mode, auxiliary participation, routing decisions, input length, and replacement match counts without exposing regex or input text;
- keep diagnostics disabled by default with no diagnostic allocations;
- use a synchronized mutable call site and guarded fast paths to eliminate measurable disabled-mode overhead;
- add selection, conformance, fallback, lifecycle, concurrency, privacy, and replacement regression coverage;
- add standard JMH coverage and user documentation for both static and runtime APIs.

Fixes #474

## API overview

Static analysis exposes stable pattern features, possible engine capabilities, limitations, program size, and capture count. Runtime diagnostics emit one bounded immutable event per normally completed public operation. Replacement methods emit one aggregate event rather than nested `find()` events.

The listener is process-wide, synchronous, visible across threads, and snapshotted once per operation. Listener exceptions propagate after matcher state is finalized. `SafeReMatchDiagnostics.NONE` is installed by default.

See `DIAGNOSTICS.md` for usage examples and `design/ENGINE_DIAGNOSTICS_DECISION_TABLE.md` for routing semantics.

## Performance

All JMH invocations ran serially through `./run-java-benchmarks.sh`. Routine evidence used the standard configuration: 2 forks, 2 × 500 ms warmup iterations, and 5 × 500 ms measurement iterations. Close or surprising cases were confirmed with `--long`.

The disabled matrix compared base commit `c8f51f3fa99976566ac9a3260df371caedec929e` with this branch using identical benchmark source. Branch/base geometric mean was `0.994`, and no slower result had a statistically convincing material regression.

| Workload | Baseline ns/op | Branch ns/op | Branch / baseline |
|---|---:|---:|---:|
| Tiny literal full match | 14.141 | 14.157 | 1.001 |
| OnePass match | 59.551 | 59.468 | 0.999 |
| BitState-sized ambiguous match | 89.412 | 89.369 | 1.000 |
| Long DFA find | 579.022 | 605.815 | 1.046 |
| Long NFA match | 13,336.972 | 13,449.810 | 1.008 |
| No-match character-class replacement | 20.690 | 20.449 | 0.988 |
| Sparse DFA replacement | 1,596.580 | 1,638.242 | 1.026 |

Long-mode confirmation of the initially sensitive cases had overlapping confidence intervals:

| Workload | Baseline ns/op | Branch ns/op |
|---|---:|---:|
| No-match character-class replacement | 20.312 ± 0.248 | 20.716 ± 0.338 |
| OnePass match | 59.697 ± 0.990 | 60.904 ± 1.687 |

A JFR allocation probe found no diagnostic event or bookkeeping allocations with diagnostics disabled.

Enabled diagnostics add approximately 150–200 ns of fixed work per public operation. This is significant for tiny operations and proportionally small for longer work; for example, a no-op listener measured 162.858 ns/op for a tiny literal full match, 800.373 ns/op for a long DFA find, and 13,731.093 ns/op for a long NFA match. The complete disabled, no-op, and `LongAdder` tables are in `design/ISSUE_474_PHASE_7_PERFORMANCE.md`.

Cached `Pattern.analysis()` retrieval measured 0.528 ± 0.020 ns/op. Compile-only and compile-plus-first-analysis confidence intervals overlapped broadly.

## Verification

Passed before the final documentation-only commit:

```text
mvn -pl safere spotless:check -q
mvn -pl safere -Dtest=DiagnosticsTest test -q
mvn -pl safere test -q
mvn verify -pl safere,safere-crosscheck -am -Pcrosscheck-public-api-tests -q
```

Additional verification:

- JFR disabled-allocation probe: no diagnostic event or bookkeeping allocations found.
- Codex review of the implementation against `main`: no actionable correctness findings.
- Codex review of the Phase 7 diff against `6e427ee`: no actionable correctness findings.
- `git diff --check` and documentation link checks passed after the final documentation commit.

Not rerun after the final documentation-only commit:

- full SafeRE test suite;
- public API crosscheck;
- JMH benchmarks.

No external-project validation was run for this change.

### DIFF
--- a/DIAGNOSTICS.md
+++ b/DIAGNOSTICS.md
@@ -1,0 +1,257 @@
+# SafeRE Diagnostics
+
+SafeRE provides two complementary diagnostics APIs:
+
+- **Static analysis** describes a compiled pattern's features, possible execution strategies, and
+  known limitations without matching an input.
+- **Runtime diagnostics** reports which strategies actually participated in a public matching or
+  replacement operation.
+
+Static analysis answers “what can this pattern do?” Runtime diagnostics answers “what happened for
+this operation and input?” Engine selection can depend on the operation, input length, match result,
+regions and bounds, captures, and runtime engine budgets, so static capabilities do not predict one
+specific strategy for every call.
+
+## Static pattern analysis
+
+Call `Pattern.analysis()` on a compiled pattern:
+
+```java
+import org.safere.Pattern;
+import org.safere.PatternAnalysis;
+import org.safere.PatternCapability;
+import org.safere.PatternFeature;
+import org.safere.PatternLimitation;
+
+Pattern pattern = Pattern.compile("^(?:(a)?)*$");
+PatternAnalysis analysis = pattern.analysis();
+
+if (analysis.features().contains(PatternFeature.CAPTURES_IN_QUANTIFIER)) {
+  System.out.println("The pattern captures inside a quantified expression");
+}
+
+if (analysis.capabilities().contains(PatternCapability.DFA_REJECT_PREFILTER)) {
+  System.out.println("A DFA may reject non-matches before an exact engine runs");
+}
+
+if (analysis.limitations().contains(PatternLimitation.NULLABLE_LOOP_REQUIRES_EXACT_ENGINE)) {
+  System.out.println("Successful matches require exact nullable-loop semantics");
+}
+
+System.out.printf(
+    "program instructions=%d, capturing groups=%d%n",
+    analysis.programSize(), analysis.captureCount());
+```
+
+`PatternAnalysis` is immutable and contains:
+
+| Field | Meaning |
+|---|---|
+| `features()` | Semantic and structural properties such as literals, captures, alternation, anchors, nullable expressions, and grapheme operations. |
+| `capabilities()` | Strategies that can participate in at least one operation or input for this pattern. |
+| `limitations()` | Stable reasons certain optimized strategies cannot be authoritative. |
+| `programSize()` | Size of the compiled SafeRE instruction program. |
+| `captureCount()` | Number of user capturing groups, excluding group 0. |
+
+The returned sets and analysis object are safe to retain and share between threads. Complete
+analysis is computed lazily when explicitly requested or when an enabled listener needs a
+compilation event, then cached on the `Pattern`. Repeated `analysis()` calls return the cached
+immutable result.
+
+Capabilities are possibilities, not execution promises. For example, a pattern can support both
+DFA boundary search and NFA matching. A long search might use a DFA to discover group 0 boundaries,
+while a capture access or a semantically ambiguous case can require an exact engine.
+
+## Runtime diagnostics
+
+Runtime diagnostics use one process-wide synchronous listener. Subclass `SafeReMatchDiagnostics`,
+install it with `Pattern.setDiagnostics`, and restore the previous listener when the observation
+scope ends:
+
+```java
+import org.safere.OperationDiagnostics;
+import org.safere.Pattern;
+import org.safere.PatternCompiledEvent;
+import org.safere.SafeReMatchDiagnostics;
+
+SafeReMatchDiagnostics listener =
+    new SafeReMatchDiagnostics() {
+      @Override
+      public void onPatternCompiled(PatternCompiledEvent event) {
+        System.out.printf(
+            "compiled patternId=%d capabilities=%s%n",
+            event.pattern().patternId(),
+            event.pattern().analysis().capabilities());
+      }
+
+      @Override
+      public void onOperationCompleted(OperationDiagnostics event) {
+        System.out.printf(
+            "%s outcome=%s boundary=%s captures=%s/%s matches=%d%n",
+            event.operation(),
+            event.outcome(),
+            event.boundaryStrategy(),
+            event.captureStrategy(),
+            event.captureMode(),
+            event.matchCount());
+      }
+    };
+
+SafeReMatchDiagnostics previous = Pattern.diagnostics();
+Pattern.setDiagnostics(listener);
+try {
+  Pattern pattern = Pattern.compile("(a+)b");
+  pattern.matcher("xxaaab").find();
+} finally {
+  Pattern.setDiagnostics(previous);
+}
+```
+
+Use `SafeReMatchDiagnostics.NONE` to disable runtime diagnostics explicitly:
+
+```java
+Pattern.setDiagnostics(SafeReMatchDiagnostics.NONE);
+```
+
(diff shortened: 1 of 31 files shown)


eaftan:
@cushon WDYT of this to address issue #474?  You probably want to start by reading the new `DIAGNOSTICS.md` that's part of this PR to get a sense for what the API looks like.  I added both a static analysis and a dynamic listener API as you suggested in the issue.

cushon:
cc @eamonnmcmanus 

Thanks @eaftan! I won’t have a chance to look at this in depth for a couple of weeks.

I’m still interested in using this to get runtime metrics on strategy selection, and maybe static analysis for patterns using expensive features.

One more thought (and sorry if this is already there) is that this could be useful as a way to add golden strategy selection tests for patterns like the ones in the benchmarks, and as regression tests for performance bug fixes. It could help ensure those patterns doesn’t start requiring more dfa passes than they should, and dfa eligibility doesn’t change, etc.

eaftan:
> One more thought (and sorry if this is already there) is that this could be useful as a way to add golden strategy selection tests for patterns like the ones in the benchmarks, and as regression tests for performance bug fixes. It could help ensure those patterns doesn’t start requiring more dfa passes than they should, and dfa eligibility doesn’t change, etc.

Good feedback.  Some of this was there, some I just added.

Before the latest change, golden tests could already assert:
- Whether a pattern is DFA-eligible through `Pattern.analysis().capabilities()`.
- Which strategy actually supplied match boundaries through `boundaryStrategy()`.
- Which other strategies participated or were bypassed.

But there wasn't a way to assert on a specific number of DFA searches attempted by an operation.  I just added that, now you can do it like so:

```java
assertThat(event.forwardDfaSearchCount()).isEqualTo(1);
assertThat(event.reverseDfaSearchCount()).isEqualTo(1);
```

eaftan:
I'm pretty confident something like this is needed, and I'm seeing use cases for it myself.  So I'm going to merge, and if we discover additional stuff later that we want to add, we can do that as a follow up.

PR #580, opened 2026-07-21
Accelerate long-input literal search with a SWAR candidate filter

### LINKED ISSUE #579: Multi-byte literal fast path is scalar and gives no benefit over the general engine
## Summary

The literal fast path in `Pattern.find(Utf8Input)` is very effective for single-byte literals, but for multi-byte literals it gives no measurable benefit over the general engine — and is ~2-3.5x slower than a straightforward JVM substring search over the same bytes.

The asymmetry looks structural: the single-byte path is SWAR (8 bytes per step), while the multi-byte path is scalar (1 byte per step).

Context: we're evaluating SafeRE as a regex backend for [Trino](https://github.com/trinodb/trino), matching directly over UTF-8 via the new `Utf8Input` API in 0.9.0. That API is a great fit for us — we can borrow a `Slice`'s backing array with zero copying — so thanks for adding it. This is the one spot where we still trail the other engines.

## Measurements

Input: 32,768 bytes of random `[a-z]`. Needle never occurs, so every run is a full scan (worst case). safere 0.9.0, JDK 25 (Temurin), macOS aarch64.

**Is the fast path engaged?** Comparing `Pattern.find(Utf8Input)` (has the literal fast path) against `Pattern.matcher(Utf8Input).find()` (general engine):

| pattern | `find()` | `matcher().find()` | ratio |
|---|---|---|---|
| `x` (1 byte) | **81 ns** | 355 ns | **4.4x** |
| `literal` (7 bytes) | 17,483 ns | 17,845 ns | 1.0x |
| `coolfunctionname` (16 bytes) | 8,909 ns | 8,978 ns | 1.0x |
| `l[it]teral` (not a literal) | 51,650 ns | 54,052 ns | 1.0x |

So the fast path is clearly working for 1 byte (81 ns for 32 KB is memchr-class), but multi-byte literals get no benefit over the general engine.

**How does it compare to a plain substring search?**

| pattern | safere `find()` | `Slice.indexOf` | `String.indexOf` |
|---|---|---|---|
| `literal` | 17,669 ns | **5,011 ns (3.5x)** | 6,949 ns (2.5x) |
| `coolfunctionname` | 8,954 ns | **4,952 ns (1.8x)** | 6,463 ns (1.4x) |

(`Slice.indexOf` is airlift's byte-array substring search; `String.indexOf` is the JDK intrinsic. Neither is doing anything clever that SafeRE couldn't.)

## Likely cause

In `Utf8InputScanner`:

- `indexOfByte(byte, int)` — SWAR: loads a `long`, XORs against the repeated target byte, and uses the `(x - 0x0101...) & ~x & 0x8080...` zero-byte test to advance **8 bytes per step**.
- `indexOf(byte[] literal, int[] failure, int[] shifts, int start)` — for `literal.length == 1` it delegates to `indexOfByte`. Otherwise it runs `boundedBoyerMooreHorspool`, falling back to a KMP loop. Both advance **1 byte per step** in the scanning loop.

For a short needle over a large haystack the scan is dominated by locating candidate positions, so the byte-at-a-time outer loop dominates regardless of how good the shift table is.

## Suggestion

Use the SWAR `indexOfByte` that already exists as a candidate prefilter for the multi-byte case: scan for the literal's first byte (or, better, its statistically rarest byte) 8 bytes at a time, then verify the full literal at each candidate and fall back to the current shift-table logic to advance. That's the memchr-and-verify strategy glibc and Hyperscan use, and it reuses code already in the file.

A `Vector` API implementation would go further, but SWAR alone should close most of the gap and keeps the no-dependency, no-incubator-module property.

## Repro

```java
import org.safere.*;

public class LiteralFastPath {
    static long bench(java.util.function.BooleanSupplier f, int iters) {
        long t0 = System.nanoTime();
        boolean sink = false;
        for (int i = 0; i < iters; i++) sink ^= f.getAsBoolean();
        if (sink) System.out.print("");
        return (System.nanoTime() - t0) / iters;
    }

        byte[] src = new byte[32768];
        java.util.Random r = new java.util.Random(42);
        for (int i = 0; i < src.length; i++) src[i] = (byte) ('a' + r.nextInt(26));
        Utf8Input in = Utf8Input.trusted(src);

        for (String pat : new String[] {"x", "literal", "coolfunctionname"}) {
            Pattern p = Pattern.compile(pat);
            bench(() -> p.find(in), 500);                 // warmup
            bench(() -> p.matcher(in).find(), 500);
            long fast = bench(() -> p.find(in), 3000);
            long general = bench(() -> p.matcher(in).find(), 3000);
            System.out.printf("%-18s find()=%7d ns   matcher().find()=%7d ns   ratio=%.1fx%n",
                    "'" + pat + "'", fast, general, general / (double) fast);
        }
    }
}
```

## Caveat

These are wall-clock loops rather than JMH, taken on a machine that was under load, so treat the absolute numbers as indicative. The ratios were stable across repeated warmup rounds and are what the report rests on — in particular the 1.0x for multi-byte literals versus 4.4x for the single-byte case.

wendigo:
Accelerates multi-byte literal search over UTF-8 input with a SWAR candidate
filter, keeping Boyer-Moore-Horspool for short inputs.

Fixes #579

## The reported diagnosis was not the cause

The issue attributes the problem to a scalar 1-byte-per-step KMP loop. That is
not what runs. Instrumenting `boundedBoyerMooreHorspool` on the exact repro
input shows the skip loop is engaged, completes, and never falls back to KMP:

| literal | outer iterations | avg advance | work used | work limit | fell back? |
|---|---|---|---|---|---|
| `literal` | 5,301 | 6.18 B/step | 5,502 | 65,536 | no |
| `coolfunctionname` | 2,645 | 12.39 B/step | 2,767 | 65,536 | no |

The cost is per step, not per byte: ~3.6 ns per iteration, because each step is
a serialized load, bad-character table lookup, and add. There is no
instruction-level parallelism to exploit, and a wider scan — not a smarter skip
— is what helps.

## Change

Locate candidates with a SWAR filter on the literal's first *and* last byte,
then verify each candidate in full. Two words are loaded per step and XOR-ed
against the broadcast bytes, so the standard zero-byte test identifies
positions where both bytes agree. This examines eight positions per step with
no data-dependent branching.

Requiring both bytes rather than the first alone is what keeps this cheap: on
random `[a-z]` it drops the candidate rate from 1/26 to 1/676, so verification
cost stays negligible.

Two correctness points:

- **The zero-byte test has false positives.** Exhaustive check over every
  (lane, value, filler) combination: 7 false-positive lanes, 0 false negatives.
  It is therefore sound as a filter, but a set lane bit cannot be read as "first
  and last byte matched" — every candidate is verified against the whole
  literal, including those two bytes.
- **The linear-time guarantee is preserved.** Verification is O(literal length)
  per candidate, so an adversarial input (haystack `aaaa…`, literal `abbbbbba`)
  can drive this quadratic. The existing work budget and its `-2` sentinel are
  retained: the budget is exhausted and the search falls back to linear-time
  KMP. Verified that this path fires and still returns correct results.

**Boyer-Moore-Horspool is kept for short inputs.** Measurement across literal
lengths shows the two algorithms win over different ranges — the skip loop
finishes a short input in a handful of steps, while the filter must amortize
its wider setup and a linear tail. The crossover scales with the literal
length, since a longer literal lets the skip loop advance further per step, so
the filter is selected at `max(64, 40 * literal length)` remaining bytes.

## Benchmarks

`run-java-benchmarks.sh` standard mode (2 forks, 2×500ms warmup, 5×500ms
measurement), JDK 25 Temurin, macOS aarch64. Baseline is `main` plus the new
benchmark, so both sides measure identical workloads. Times ns/op; ratio is
after/before, so lower is faster.

**Noise floor first.** 17 rows in this suite cannot be affected by this change
— they use `String.indexOf` or the JDK engine and never enter
`Utf8InputScanner`. Their spread bounds what is resolvable:

| control row | before | after | ratio |
|---|---|---|---|
| captureFreeDecode [multibyteEarly] | 27.6 | 44.5 | 1.61x |
| captureFreeDecode [multibyteLate] | 31.4 | 47.3 | 1.51x |
| captureFreeDecode [asciiNoMatch] | 21.8 | 26.9 | 1.23x |

Median control deviation 3%, maximum 61%. Nothing below ~1.2x in this run is
meaningful.

**New `literalScan` benchmark** — 32KB of random `[a-z]`, literal absent, so
every invocation is a full scan. This is the issue's scenario, which the
existing suite did not cover (its literal cases are ~20-byte texts that never
enter the vectorized loop):

| literal | before | after | speedup |
|---|---|---|---|
| `qjxv` (4 B) | 31,990 | 5,177 | **6.2x** |
| `literal` (7 B) | 19,082 | 5,189 | **3.7x** |
| `coolfunctionname` (16 B) | 9,480 | 4,231 | **2.2x** |

Long-mode confirmation of the same three: 4,144 ± 51, 4,045 ± 27, 4,450 ±
1,351. For reference, the issue measured `Slice.indexOf` at ~5,000 ns and
`String.indexOf` at ~6,500 ns for `literal`.

**Scaling** (`hardFailure`, required literal absent):

| size | before | after | ratio |
|---|---|---|---|
| 10,240 | 867 | 667 | 0.77x |
| 102,400 | 8,564 | 6,639 | 0.78x |
| 1,048,576 | 85,181 | 57,608 | **0.68x** |
| 1,024 | 59.9 | 72.2 | 1.20x — see below |

All other UTF-8-path rows fall between 0.87x and 1.04x, inside the noise floor.

## Open questions for review

Two things I could not settle on this hardware, both worth a look before merge:

1. **`hardFailure[1024]` measures ~1.2x slower.** Long mode gives 69.8 ± 1.9 vs
   59.9 ± 1.5, which looks real. But with the filter disabled entirely it
   measures 61.6, matching baseline — and at that size the threshold should
   already be routing this pattern's 26-byte literal to the skip loop, i.e. the
   same path as "disabled". Either I have misread which literal this pattern
   extracts, or it is noise. I am not claiming it is resolved.
2. **`FILTER_LENGTH_FACTOR = 40` is fitted to these measurements.** The shape is
   principled — the filter always advances 8 bytes per step while the skip loop
   can advance up to the literal length — but the constant itself is soft and
   deserves a re-run on a quiet machine.

## Verification

Ran:

- Full `safere` test suite (`mvn -pl safere test`): 14,828 tests. Green on two
  runs; a third had one failure in
  `ParserTest$CharacterClasses#manyPlainCharacterClassesParseWithLinearScaling`,
  a wall-clock ratio test in the parser that passes 3/3 in isolation and does
  not touch this code path. Consistent with the machine noise documented above.
- `Utf8MatchingBenchmark` before/after, plus long-mode confirmation of
  `hardFailure` and `literalScan`.
- Mutation check that the new tests actually cover the filter: introducing an
  off-by-one in the word step fails 3 of them. This mattered — the threshold
  means small-input tests exercise only the skip loop, so the new tests run each
  scenario both below and above it.

Not run: crosscheck, exhaustive, and fuzz modules; benchmark suites other than
`Utf8MatchingBenchmark`.

## Note on the shared benchmark script

`run-java-benchmarks.sh` fails on macOS with `JMH_EXTRA_ARGS[*]: unbound
variable`. Its shebang is `#!/bin/bash`, which is bash 3.2 on macOS, where
expanding an empty array under `set -u` is an error. These runs used Homebrew's
bash 5.3. Not fixed here as it is unrelated; happy to file it separately.

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -51,6 +51,16 @@
   },
 
   "utf8Matching": {
+    "literalScan": {
+      "seed": 20260721,
+      "textSize": 32768,
+      "alphabet": "abcdefghijklmnopqrstuvwxyz",
+      "patterns": {
+        "fourByte": "qjxv",
+        "shortWord": "literal",
+        "longWord": "coolfunctionname"
+      }
+    },
     "captureFreeFind": {
       "asciiEarly": { "pattern": "ERROR", "text": "ERROR request completed" },
       "asciiLate": { "pattern": "ERROR", "text": "request completed with ERROR" },
--- a/safere-benchmarks/src/main/java/org/safere/benchmark/Utf8MatchingBenchmark.java
+++ b/safere-benchmarks/src/main/java/org/safere/benchmark/Utf8MatchingBenchmark.java
@@ -111,6 +111,44 @@ public class Utf8MatchingBenchmark extends ByteMatchingBenchmark {
     return state.pattern.matcher(state.text).find();
   }
 
+  /** Measures a full literal scan over a large input where the literal is absent. */
+  @Benchmark
+  public boolean literalScan(LiteralScanState state) {
+    return state.pattern.find(state.input);
+  }
+
+  @State(Scope.Thread)
+  public static class LiteralScanState {
+    @Param({"fourByte", "shortWord", "longWord"})
+    public String name;
+
+    private org.safere.Pattern pattern;
+    private org.safere.Utf8Input input;
+
+    /**
+     * Builds a frozen pseudo-random single-byte-alphabet input that does not contain the literal,
+     * so every invocation scans the whole input.
+     */
+    @Setup
+    public void setup() {
+      BenchmarkData data = BenchmarkData.get();
+      String prefix = "utf8Matching.literalScan.";
+      String alphabet = data.getString(prefix + "alphabet");
+      int size = data.getInt(prefix + "textSize");
+      java.util.Random random = new java.util.Random(data.getInt(prefix + "seed"));
+      StringBuilder text = new StringBuilder(size);
+      for (int index = 0; index < size; index++) {
+        text.append(alphabet.charAt(random.nextInt(alphabet.length())));
+      }
+      String literal = data.getString(prefix + "patterns." + name);
+      if (text.indexOf(literal) >= 0) {
+        throw new IllegalStateException("literal " + literal + " must be absent from the input");
+      }
+      pattern = org.safere.Pattern.compile(literal);
+      input = org.safere.Utf8Input.trusted(text.toString().getBytes(StandardCharsets.UTF_8));
+    }
+  }
+
   @State(Scope.Thread)
   public static class CaptureFreeState {
     @Param({
--- a/safere/src/main/java/org/safere/Utf8InputScanner.java
+++ b/safere/src/main/java/org/safere/Utf8InputScanner.java
@@ -15,6 +15,16 @@ final class Utf8InputScanner implements InputScanner {
   private static final int REPLACEMENT_CHARACTER = 0xFFFD;
   private static final long BYTE_ONES = 0x0101_0101_0101_0101L;
   private static final long BYTE_HIGH_BITS = 0x8080_8080_8080_8080L;
+
+  /**
+   * Input sizes at which the SWAR candidate filter overtakes the skip loop. The filter always
+   * advances eight positions per step, while the skip loop can advance as far as the literal
+   * length, so the filter needs an input large enough to outweigh that per-step advantage. Both
+   * bounds were measured; see {@link #indexOfFiltered}.
+   */
+  private static final int MIN_FILTER_LENGTH = 64;
+
+  private static final int FILTER_LENGTH_FACTOR = 40;
   private static final VarHandle LONG_VIEW = byteArrayViewVarHandle(long[].class, nativeOrder());
 
   private final byte[] bytes;
@@ -132,15 +142,37 @@ final class Utf8InputScanner implements InputScanner {
     if (literal.length == 0) {
       return start;
     }
-    if (literal.length == 1 && !WorkCounterConfig.ENABLED) {
-      return indexOfByte(literal[0], start);
-    }
-    if (!WorkCounterConfig.ENABLED && shifts != null) {
-      int result = boundedBoyerMooreHorspool(literal, shifts, start);
-      if (result >= -1) {
-        return result;
+    if (!WorkCounterConfig.ENABLED) {
+      if (literal.length == 1) {
+        return indexOfByte(literal[0], start);
+      }
+      if (shifts != null) {
+        // Both searches beat the linear scan, but they win over different ranges. The skip loop
+        // reaches the end of a short input in a handful of steps, while the candidate filter has
+        // to pay for its wider setup and finish with a scalar tail. Once the input is long enough
+        // for the filter's eight-positions-per-step throughput to dominate that fixed cost, it
+        // wins by a growing margin. The crossover scales with the literal length, since a longer
+        // literal lets the skip loop advance further per step.
+        int result =
+            remaining(start) >= Math.max(MIN_FILTER_LENGTH, literal.length * FILTER_LENGTH_FACTOR)
+                ? indexOfFiltered(literal, failure, start)
+                : boundedBoyerMooreHorspool(literal, shifts, start);
+        // A match index or a trusted -1; only the -2 "work budget exhausted" sentinel falls
+        // through to the linear-time scan below.
+        if (result >= -1) {
+          return result;
+        }
       }
     }
+    return indexOfLinear(literal, failure, start);
+  }
+
+  private int remaining(int start) {
+    return length - start;
+  }
+
+  /** Knuth-Morris-Pratt scan, linear in the input length regardless of the literal. */
+  private int indexOfLinear(byte[] literal, int[] failure, int start) {
     int matched = 0;
     for (int position = start; position < length; position++) {
       if (WorkCounterConfig.ENABLED) {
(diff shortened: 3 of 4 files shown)


eaftan:
Thank you for the contribution, @wendigo! This looks good. I fixed an issue related to a potential integer overflow, and I reproduced your benchmark results. 

I’m digging into your questions about the `hardFailure[1024]` regression and the filter threshold. I’ll post back once I’ve investigated further. 

On the bug about the failing benchmark script on MacOS, would you mind filing a bug?

eaftan (approved):
LGTM. I reproduced clear speedups, although they were somewhat smaller than the ones reported, likely due to architectural differences between ARM and x86 (my machine).

Long-mode results on Linux/x86-64 with OpenJDK 25:

| Literal | main | This PR | Speedup |
|---|---:|---:|---:|
| `qjxv` 4 bytes | 26,147 ns/op | 6,045 ns/op | 4.33× |
| `literal` 7 bytes | 15,558 ns/op | 6,023 ns/op | 2.58× |
| `coolfunctionname` 16 bytes | 7,732 ns/op | 6,027 ns/op | 1.28× |

I pushed a follow-up commit fixing two potential `int` overflows in the filter threshold and work-budget accounting for very large inputs.

I also looked into your two open questions:

`hardFailure[1024]` regression -- This is real.   The added filter path changes C2’s optimization of the BMH loop even though the filter is not selected. I’ve restored the original performance in the stacked PR #583.

`FILTER_LENGTH_FACTOR` -- I reran the filter/BMH crossover measurements on a quiet x86-64 machine. `FILTER_LENGTH_FACTOR = 40` looks like a good conservative choice:

| Literal length | Observed filter crossover | Current threshold |
|---:|---:|---:|
| 4 | ≤64 bytes | 160 bytes |
| 7 | ≤64 bytes | 280 bytes |
| 16 | 256–384 bytes | 640 bytes |
| 26 | around 1,152 bytes | 1,040 bytes |

The exact crossover depends on the input distribution, but factor 40 avoids switching prematurely when BMH skips especially well. It leaves some performance available for short literals, while landing near the crossover for the longer, more workload-sensitive case. I think we should keep it.

wendigo:
@eaftan when do you plan next release? :)

eaftan:
> @eaftan when do you plan next release? :)

Done!  https://github.com/eaftan/safere/releases/tag/v0.10.0

PR #602, opened 2026-07-27
Latin1 character-class prefix acceleration

cushon:
This PR extends SafeRE's start-position character class acceleration from ASCII-only ranges to the full 256-bit Latin-1 character set.

When a pattern starts with a character class containing non-ASCII Latin-1 characters, `Matcher` now scans ahead using a 256-bit bitmask to skip text positions that cannot begin a match, avoiding un-accelerated DFA scanning over large non-matching text regions.

```
./run-java-benchmarks.sh --fastbuild 'JavaCharacterClassBenchmark.latin1Find_safere'
```

Before

```
JavaCharacterClassBenchmark.latin1Find_safere  avgt   10  37.058 ± 0.372  us/op
```

After

```
JavaCharacterClassBenchmark.latin1Find_safere  avgt   10  26.432 ± 0.281  us/op
```


eaftan (commented):
I don’t think we should merge this optimization. Representative long-scan benchmarks improved by about 16–18%. These are favorable 10 KB inputs with sparse candidates, so the practical benefit is likely smaller on shorter inputs or text containing frequent candidates.

To estimate prevalence, I searched Sourcegraph’s public GitHub index for relevant Java `Pattern.compile` calls and manually inspected the results. I found only 12 unrelated production uses likely to benefit.

The optimization is real, but the affected workload appears uncommon, the measured gain is modest, and the additional metadata and matcher path increase implementation and maintenance complexity. I don’t think the expected benefit justifies that cost.

cushon:
Thanks for taking a look, dropping this SGTM. I got here by searching for ideas from resharp, I don't have evidence this is helpful for production workloads.

Issue #636, opened 2026-07-28
safere 0.10.0: prefilter scanner selection leaves 10x on the table


wendigo:
Four proposals, ordered by (measured payoff × confidence). The first two are small, local, and
well-evidenced. The third is a design change with a large payoff on realistic data. The fourth is a
lead, not a conclusion.

## How the numbers were produced

`org.safere:safere:0.10.0`, JDK 25 (Temurin), Apple aarch64. Hand-rolled measurement loops — best of
3 runs, 3k warmup + 30k measured iterations, single JVM, **not JMH**. Treat ≥2x differences as real
and anything under ~1.3x as noise. Strategy attribution comes from `Pattern.setDiagnostics` via the
`CharSequence` matcher (the UTF-8 matcher NPEs — see the companion bug report), timings come from the
UTF-8 API over 32 KiB inputs.

Context: these came out of benchmarking safere as a regexp backend for Trino
(`regexp_like` / `regexp_count` / `regexp_replace`), against Joni and RE2/J. safere wins nearly
everywhere — 8-18x over Joni on `regexp_like` for dot-star patterns. These are the cases where it
doesn't.

---

## 1. A one-code-point ASCII class is rejected with a scalar scan instead of the SWAR byte scan

**Payoff: ~11.8x on the reject path. Confidence: high — strategy-attributed, DFA provably not
involved.**

Two patterns that must do the identical amount of real work — "is there an `x` in this input" — over
32 KiB of `'a'`:

| pattern | strategy (from diagnostics) | fwdDfa | time |
|---|---|---|---|
| `.*x.*` | `CHARACTER_CLASS / REJECT_PREFILTER` | 0 | **8983 ns** |
| `x.*` | `LITERAL / START_ACCELERATION` | 0 | **762 ns** |

`fwdDfa=0` in both cases: the DFA never runs, the prefilter answers alone. The entire 11.8x is which
scanner the prefilter calls.

### Root cause

`Utf8InputScanner.indexOfCodePointClass` (`Utf8InputScanner.java:86`) decodes and range-tests one
code point at a time whenever the class contains any ASCII:

```java
public int indexOfCodePointClass(int[] ranges, long bitmap0, long bitmap1, int start) {
  int position = Math.max(0, start);
  if (!WorkCounterConfig.ENABLED && bitmap0 == 0 && bitmap1 == 0) {
    return indexOfNonAsciiCodePointClass(ranges, position);   // SWAR-skips 8 ASCII bytes at a time
  }
[...]

PR #642, opened 2026-07-29
Accelerate UTF-8 ASCII character-class scans

eaftan:
## Summary

- route one- and two-code-point ASCII classes through the existing SWAR byte scanners
- add an eight-byte common-prefix filter for contiguous ASCII ranges, with exact scalar candidate verification
- preserve scalar scanning when work counting is enabled
- add exhaustive range/alignment coverage and declarative JMH workloads for the optimized paths

This implements proposals 1 and 2 from #636. The fixed-offset/selectivity and repeated-`find()` DFA proposals remain out of scope.

Refs #636

## Profiling

JFR on the contiguous-digit-range workload attributed 100% of sampled Java execution time before the change to `Utf8InputScanner.indexOfAsciiClassScalar`.

## Benchmarks

Standard Java benchmark configuration on OpenJDK 26.0.1:

|---|---:|---:|---:|
| single ASCII required-class rejection (`.*x.*`) | 9,171.986 ns/op | 951.358 ns/op | 9.64x faster |
| contiguous ASCII range rejection (`\d{3}/\d{3}/\d{4}`) | 10,165.548 ns/op | 1,154.255 ns/op | 8.81x faster |
| literal rejection control (`x.*`) | 958.001 ns/op | 950.862 ns/op | no material change |

Command:

```bash
./run-java-benchmarks.sh --fastbuild \
  '^org\.safere\.benchmark\.CrossEngineBenchmark\.run$' -- \
  -p 'crossEngineTrial=Utf8MatchingBenchmark.requiredSingleAsciiNonmatchBytes@safere-utf8,Utf8MatchingBenchmark.asciiRangePrefixNonmatchBytes@safere-utf8,Utf8MatchingBenchmark.literalPrefixNonmatchBytes@safere-utf8'
```

## Verification

Run:

- `mvn -pl safere test -q`
- `mvn -pl safere,safere-benchmarks spotless:check -q`
- `git diff --check`
- standard two-fork JMH before/after comparison shown above
- JFR profile of the contiguous ASCII range workload

Not run:

- public API crosscheck profile, because this is an internal behavior-preserving scanner optimization
- full benchmark collection or external OpenJDK-derived benchmarks, because the change is limited to preexisting UTF-8 character-class scanning

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -264,6 +264,15 @@
       "shared": true
     },
     {
+      "id": "utf8Matching.asciiClassNonmatch.text",
+      "recipe": {
+        "kind": "repeatToLength",
+        "unit": "a",
+        "length": 32768
+      },
+      "shared": true
+    },
+    {
       "id": "utf8Matching.requiredClassNonmatch.text",
       "recipe": {
         "kind": "repeatToLength",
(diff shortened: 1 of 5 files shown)


eaftan:
LGTM, will merge after CI.  Also will look at proposals 3 and 4 from #636 after this.

PR #643, opened 2026-07-29
Accelerate fixed-offset literal searches

eaftan:
## Summary

- choose the smallest mandatory character class for negative prefiltering
- detect case-sensitive ASCII literals at an exact offset from the match start
- scan those literals and reject candidates whose derived start cannot match the leading ASCII class
- add Java-string and preexisting-UTF-8 no-slash, unrelated-slash, sparse-match, and
  dense-control workloads

This builds on issue proposals 1 and 2 in #642 and implements issue proposal 3 from #636.

The measured acceleration is specific to the preexisting-UTF-8 path. Before this PR, the
Java-string matcher already searched required literals with the JDK's optimized
`String.indexOf()`. The fixed-offset String path uses that same primitive, so it does not introduce
a new source of acceleration; it preserves the existing fast behavior when the new selectivity
choice changes which prefilter wins. The preexisting-UTF-8 path lacked an equivalent
start-candidate route, which is why it receives the large speedups below. Long before/after
measurements show no statistically resolved String speedup.

Part of #636

### Preexisting UTF-8

Standard Java benchmark configuration: 2 forks, 2 x 500 ms warmup iterations, and 5 x 500 ms
measurement iterations. Times are ns/op; lower is better.

| --- | ---: | ---: | ---: |
| find, digits but no slash | 50,733 | 1,027 | 49.4x faster |
| find, unrelated slash noise | 50,076 | 4,888 | 10.2x faster |
| find, sparse 511 | 33,677 | 651 | 51.7x faster |
| find, sparse 63 | 4,106 | 104 | 39.3x faster |
| find, sparse 7 | 401 | 25 | 15.7x faster |
| find, sparse 1 | 61 | 18 | 3.3x faster |
| find-all, sparse 511 | 39,859 | 999 | 39.9x faster |
| find-all, sparse 63 | 51,823 | 1,548 | 33.5x faster |
| find-all, sparse 7 | 48,343 | 6,036 | 8.0x faster |
| find-all, sparse 1 | 54,633 | 30,194 | 1.8x faster |
| find-all, dense digits/slashes | 57,927 | 54,697 | 5.9% faster |

### Java String

Long confirmation configuration: 2 forks, 3 x 1 s warmup iterations, and 5 x 1 s measurement
iterations.

| Workload | Before | After |
| --- | ---: | ---: |
| find, digits but no slash | 1,029 | 1,077 |
| find, unrelated slash noise | 8,897 | 8,978 |
| find, sparse 511 | 718 | 695 |
| find, sparse 63 | 141 | 144 |
| find, sparse 7 | 54 | 56 |
| find, sparse 1 | 50 | 48 |
| find-all, sparse 511 | 1,194 | 1,175 |
| find-all, sparse 63 | 1,675 | 1,695 |
| find-all, sparse 7 | 5,947 | 6,798 |
| find-all, sparse 1 | 30,024 | 33,754 |
| find-all, dense digits/slashes | 54,585 | 54,727 |

All String before/after 99.9% confidence intervals overlap. The two larger point-estimate
differences are the noisy sparse-7 and sparse-1 find-all cases; their branch-side error intervals
were ±1,129 ns/op and ±5,014 ns/op, respectively. The remaining nine workloads have a 1.003
geometric-mean after/before ratio (0.3% slower), so the String result is effectively unchanged.

## Verification

Ran:

- `mvn test -q`
- `mvn -pl safere -Pwork-counters -Dtest=Utf8LinearTimeTest test -q`
- `mvn -pl safere-benchmarks spotless:check -q`
- focused standard JMH before/after runs for all UTF-8 workloads in the table
- focused standard and long JMH before/after runs for all Java-string workloads in the table

Not run:

- the full benchmark collection
- the external OpenJDK-derived benchmark suite

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -273,6 +273,56 @@
       "shared": true
     },
     {
+      "id": "phoneSearch.digitsNoSlash.text",
+      "recipe": {
+        "kind": "randomChars",
+        "alphabet": "abcdefghijklmnopqrstuvwxyz0123456789 ",
+        "length": 32768,
+        "seed": 20260729
+      },
+      "shared": true
+    },
+    {
+      "id": "phoneSearch.slashNoise.text",
+      "recipe": {
+        "kind": "repeatToLength",
+        "unit": "service 42 completed path /api/v1/items with 7 records ",
+        "length": 32768
+      },
+      "shared": true
+    },
+    {
+      "id": "phoneSearch.denseDigitsAndSlashes.text",
+      "recipe": {
+        "kind": "randomChars",
+        "alphabet": "/0123456789",
+        "length": 32768,
+        "seed": 20260729
+      },
+      "shared": true
+    },
+    {
+      "id": "phoneSearch.sparseMatches.{nonMatchRepeats}",
+      "axes": {
+        "nonMatchRepeats": [
+          511,
+          63,
+          7,
+          1
+        ]
+      },
+      "recipe": {
+        "kind": "sparseMatchToLength",
+        "match": "123/456/7890",
+        "nonMatch": "ordinary prose event 42 completed normally",
+        "nonMatchRepeats": "{nonMatchRepeats}",
+        "delimiterAlphabet": " .,;:-",
+        "seed": 20260729,
+        "length": 32768
+      },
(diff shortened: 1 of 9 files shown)


eaftan:
LGTM, will merge after CI

PR #644, opened 2026-07-29
Explicitly ignore an unused return value

cushon:
This suppresses a https://errorprone.info/bugpattern/CheckReturnValue error


eaftan (approved):
Thanks @cushon!  I'll enable Error Prone for more submodules as you pointed out in #645.

PR #646, opened 2026-07-29
Enable Error Prone checks across all modules

### LINKED ISSUE #645: Enable Error Prone checks for more submodules
Currently Error Prone is enabled for safere, but not for other submodules.

At least some of the checks could be worth enabling generally to prevent things like #644.

eaftan:
Fixes #645

## Summary

- Centralize the existing `safere` Error Prone policy and shared compiler configuration in the
  parent POM.
- Enable the same policy for every Java module, preserving module-specific Java versions, warning
  options, and annotation processors.
- Fix existing findings exposed in crosscheck, exhaustive-validation, fuzz, and benchmark code.
- Document and use the approved localized suppression policy for intentional internal array record
  components.
- Explicitly exclude checks that are unrelated to the current policy.

## Verification

- `mvn clean test -DskipTests -q`
- `mvn -pl safere-crosscheck -am -Pcrosscheck-public-api-tests test -DskipTests -q`
- `mvn -pl safere test -q`
- Focused `CompactDivergenceLogsTest` run in `safere-exhaustive`
- Effective POM inspection for `safere`, `safere-crosscheck`, `safere-ffm-re2`, and
  `safere-benchmarks`
- `mvn spotless:check -q`
- `git diff --check`

Not run:

- Full test suites for `safere-crosscheck`, `safere-fuzz`, or `safere-exhaustive`
- Benchmarks

### DIFF
--- a/safere-benchmarks/src/main/java/org/safere/benchmark/CrossEngineWorkload.java
+++ b/safere-benchmarks/src/main/java/org/safere/benchmark/CrossEngineWorkload.java
@@ -9,51 +9,33 @@ import java.util.List;
 import java.util.Objects;
 
 /** One regex workload executed consistently across multiple engines. */
-final class CrossEngineWorkload {
-  private final String id;
-  private final BenchmarkOperation operation;
-  private final List<String> patterns;
-  private final List<String> inputKeys;
-  private final int[] groups;
-  private final String replacement;
-  private final Object expected;
-  private final int limit;
-  private final DeclarativeBenchmarkPlan.MatcherLifecycle lifecycle;
-  private final String flagSet;
-  private final int seed;
-  private final int count;
-  private final DeclarativeBenchmarkPlan.Measurement measurement;
-  private final TimingGroup timingGroup;
+// groups is internal benchmark metadata; callers do not use record value semantics.
+@SuppressWarnings("ArrayRecordComponent")
+record CrossEngineWorkload(
+    String id,
+    BenchmarkOperation operation,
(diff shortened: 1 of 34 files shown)


safere-benchmarks/src/main/java/org/safere/benchmark/CrossEngineWorkload.java
```diff
@@ -9,31 +9,51 @@
...
-  CrossEngineWorkload {
-    Objects.requireNonNull(id);
-    Objects.requireNonNull(operation);
-    patterns = List.copyOf(patterns);
-    inputKeys = List.copyOf(inputKeys);
-    groups = groups.clone();
-    Objects.requireNonNull(lifecycle);
-    Objects.requireNonNull(measurement);
-    Objects.requireNonNull(timingGroup);
+final class CrossEngineWorkload {
```
cushon: If you wanted to just disable the check that's complaining about these records as not valuable for SafeRE I wouldn't blame you, but either way is fine

cushon:
Thanks!

safere-benchmarks/src/main/java/org/safere/benchmark/CrossEngineWorkload.java
eaftan: I've considered that, but I do think there's value to the check.  I've added some guidance to my AGENTS.md that my agent is allowed to suppress this warning when it's intentional.  Also I've backed out the changes to the records in this PR and suppressed the warning.

PR #653, opened 2026-07-31
Preclassify UTF-8 ASCII prefix scans

eaftan:
## Summary

- preclassify ASCII character-class prefix metadata once during `Pattern` compilation
- route UTF-8 prefix searches through the existing singleton, pair, range, and bitmap scanners
- preserve the existing `boolean[128]` path for Java `String` matching
- add systematic metadata, UTF-8 state-machine, diagnostics, and declarative benchmark coverage

The compact ranges and two 64-bit membership masks are derived from the same immutable ASCII
prefix bitmap. Tests verify exact membership equivalence for missing and empty classes, singleton,
adjacent and nonadjacent pairs, contiguous ranges, discontiguous classes, and the 63/64 bitmap
boundary. End-to-end tests cover direct `Pattern.find(Utf8Input)`, repeated `Utf8Matcher.find()`,
regions, reset, diagnostics, byte offsets, and the unchanged String path.

## Benchmark results

Environment: Apple M5 Max (`aarch64`), OpenJDK 26.0.1. Standard configuration is 2 forks,
2 x 500 ms warmup, and 5 x 500 ms measurement. Before is `5487784`; after is `de6f65a`.

| Workload | Before | After | Improvement |
|---|---:|---:|---:|
| Pair, first, 256 B | 34.17 ns | 12.42 ns | 2.75x |
| Range, first, 256 B | 36.07 ns | 12.71 ns | 2.84x |
| Pair, absent, 256 B | 35.93 ns | 15.37 ns | 2.34x |
| Range, absent, 256 B | 34.94 ns | 11.86 ns | 2.95x |
| Pair, scan-all, sparse, 32 KiB | 25.03 us | 17.63 us | 1.42x |
| Range, scan-all, sparse, 32 KiB | 24.19 us | 16.98 us | 1.42x |
| Pair, scan-all, dense, 32 KiB | 1.259 ms | 0.902 ms | 1.39x |
| Range, scan-all, dense, 32 KiB | 0.658 ms | 0.473 ms | 1.39x |

The 32 KiB absent and late controls changed by 0.6-2.2%, where input traversal dominates the
fixed classification cost.

Long confirmation (2 forks, 3 x 1 s warmup, 5 x 1 s measurement):

| Workload | Before | After | Improvement |
|---|---:|---:|---:|
| Pair, scan-all, dense, 32 KiB | 1.366 ms | 0.949 ms | 1.44x |
[...]

Issue #504, opened 2026-06-22
Proposal: Vectorized Scanning & Substring Precheck for Unanchored Matches


cushon:
There's a draft in https://github.com/eaftan/safere/pull/503

This is still a bit of a research project, but I think it addresses one of the remaining performance gaps compared to RE2-FFM for some of the benchmark cases in https://github.com/eaftan/safere/pull/500. Having this be a separate optional dependency that falls back to alternative implementations (if the vector preview APIs aren't enabled, and/or if reflection can't be used to access `String` internals) might make it possible for some SafeRE users to enable this as an optional feature if they want.

---

This issue proposes a performance-oriented enhancement to the SafeRE regular expression engine to optimize unanchored matches. Production profiles show that a significant portion of execution time is spent on unanchored patterns (e.g., searches for URLs, wildcards, or protocols) that do not result in a match, causing the engine to scan the entire input text character-by-character.

We introduce a two-phase optimization cascade that accelerates unanchored searches:
1. **Required Substring Precheck (Case-Sensitive)**: During pattern compilation, we extract a mandatory literal substring (if present in the AST) that must appear in the input text for any match to be possible. At matching start, we execute a fast `indexOf` check. If absent, the search terminates instantly with no match, avoiding running the DFA state machine entirely.
2. **Vectorized SIMD Character Scanning**: When an unanchored pattern begins with a single character class (and no literal prefix is available), we scan the input text using the **Java Vector API** (`jdk.incubator.vector`). This compiles down to raw hardware SIMD instructions (AVX/Neon) at runtime, comparing 16 characters (UTF-16) or 32/64 characters (Latin-1) in parallel in a single CPU clock cycle.

---

## Modularity & Portable Runtime Fallback

To support this in open-source without forcing all downstream users to execute their JVMs with preview flags (`--enable-preview --add-modules jdk.incubator.vector`), the design is strictly decoupled:
* **`safere` (Core Module)**: Defines `VectorScannerBridge` and a reflection-based `VectorScannerLoader` class initializer.
* **`safere-vector` (Optional Module)**: Implements the `VectorScannerImpl` targeting JDK 26 using FFM (`MemorySegment`) and the incubator Vector API.
* **Graceful Fallback**: If the optional `safere-vector` JAR is missing from the classpath, or if JVM preview flags are omitted, the loader catches the `Throwable` and falls back transparently to a high-performance inline scalar loop.

### Zero-Copy Compact Strings Optimization

[...]

PR #654, opened 2026-07-31
Add experimental JDK 26 Vector provider for UTF-8 scans

eaftan:
## Summary

- package the experimental JDK 26 Vector scanner inside the normal SafeRE artifact as a
  multi-release JAR
- keep the Vector provider contract and implementation package-private; this PR does not add a
  public SPI
- name the provider independently of UTF-8 so future String scanning experiments can use the same
  internal JDK-versioned framework
- select the JDK 26 implementation immutably at startup through an explicit experimental property
- add declarative JMH workloads for singleton, pair, and range scans over UTF-8 inputs
- compare native SWAR and Vector execution in separate JVMs through complete
  `Pattern.find(Utf8Input)` calls
- document opt-in usage and run full SafeRE, public API crosscheck, exhaustive, work-counter,
  packaging, and runtime smoke coverage with Vector enabled in CI

The base artifact continues to target Java 21. All references to `jdk.incubator.vector`, together
with a versioned module descriptor, live under `META-INF/versions/26`. No additional dependency or
class-path assembly is required.

On JDK 26, applications opt in with:

```text
--add-modules=jdk.incubator.vector
-Dorg.safere.experimental.vectorScanProvider=vector
```

Without the property, SafeRE continues to use SWAR. Requesting Vector without the incubator module
flag fails with a configuration error. Java 21 through 25 ignore the versioned implementation.

The activation property, implementation, supported scans, and tuning thresholds are experimental
and may change incompatibly or be removed in any SafeRE release.

## Draft status

This is a draft so the implementation can be tried in real applications while the provider shape
and broader opportunities for vectorized UTF-8 and String scanning are still under investigation.

This work builds on the UTF-8 ASCII-prefix classification and SWAR scanners merged in #653.
Possible support for the experimental Vector scanner on JDK 21 through 25 is tracked in #658.

## Benchmark results

Standard Java benchmark configuration on OpenJDK 26.0.1: 2 forks, 2 x 500 ms warmup, and
5 x 500 ms measurement iterations. Times are complete `Pattern.find(Utf8Input)` operations.

| Scan workload | SWAR | Vector | Ratio |
| --- | ---: | ---: | ---: |
| Singleton absent, 1 KiB | 31.890 ns | 17.684 ns | Vector 1.80x faster |
| Pair absent, 1 KiB | 52.306 ns | 20.904 ns | Vector 2.50x faster |
| Range absent, 1 KiB | 37.924 ns | 22.065 ns | Vector 1.72x faster |
| Singleton dense, 32 KiB | 11.699 ns | 11.575 ns | effectively neutral |
| Pair dense, 32 KiB | 10.480 ns | 9.944 ns | Vector 1.05x faster |
| Range dense, 32 KiB | 10.736 ns | 10.395 ns | Vector 1.03x faster |

The provider uses a 1 KiB minimum remaining-input threshold and a four-byte scalar prologue. The
threshold avoids regressions on short inputs; the prologue protects immediate dense matches from
the setup cost of a wide Vector scan.

## Verification

Completed:

- `mvn -pl safere test -q`
- `mvn clean test -pl safere -Pvector-tests -Dtest.parallelism=4 -q`
- standalone generated public API crosscheck against the installed MR-JAR with Vector enabled
- standalone exhaustive validation tests against the installed MR-JAR with Vector enabled
- `mvn clean test -pl safere -Pwork-counters,vector-tests -Dtest.parallelism=4 -q`
- `mvn clean package -Prelease -DskipTests -Dpmd.skip=true -q -f safere/pom.xml`
- `mvn verify -DskipTests -Dpmd.skip=true -q -f safere-vector-benchmarks/pom.xml`
- `jar --validate` on the packaged multi-release SafeRE JAR
- inspection of the JDK 26 classes and versioned module descriptor in the packaged JAR
- standard end-to-end SWAR/Vector benchmark matrix shown above
- deterministic packaged SWAR and enabled-Vector runtime smoke checks
- focused JMH smoke benchmark through `run-vector-benchmarks.sh`
- class-path and module-path Vector activation with only the documented flags
- ordinary execution without Vector flags
- clear failures for a missing incubator module flag and an unavailable versioned implementation

Refs #504


cushon:
The module name `safere-vector-jdk26-experimental` leaves room for new modules for new versions of the Vector APIs if there are breaking changes, which could happen. Are you thinking this could potentially support different versions of the Vector API for different JDK versions in the future? If so maybe the provider name should be `vector-jdk26` or something instead of just `vector`? It might also be possible to ship multiple versions and reflectively load the right one based on `Runtime.version().feature()` or something, or use MR-JARs. (I also think it'd be reasonably to not worry too much about different versions of the Vector API for now.)

Also relaying an agentic code review note: have you considered adding substring / literal search to the SPI? A Vector-based substring search could be faster than Boyer-Moore-Horspool for long inputs. i.e. something like

```java
  /**
   * Finds the first occurrence of a literal byte sequence.
   *
   * @return position of the match, or -1 if not found, or -2 if this provider 
   *         does not support literal search (signaling fallback to scalar).
   */
  default int indexOfLiteral(
      byte[] bytes,
      int offset,
      int length,
      byte[] literal,
      int[] failure,
      int[] shifts,
      int start) {
    return -2; 
  }
  ```

and then in `Utf8InputScanner.java`

```java
  int indexOf(byte[] literal, int[] failure, int[] shifts, int start) {
    if (scanProvider != null) {
      int idx = scanProvider.indexOfLiteral(bytes, offset, length, literal, failure, shifts, start);
      if (idx != -2) {
        return idx;
      }
    }
    // Fallback to scalar Boyer-Moore-Horspool
    ...
  }
```

safere/src/main/java/org/safere/Utf8InputScanner.java
```diff
@@ -126,6 +129,9 @@ public int indexOfCodePointClass(int[] ranges, long bitmap0, long bitmap1, int s
    */
   private int indexOfAsciiRanges(int[] ranges, int start) {
     if (ranges.length == 2 && ranges[0] >= 0 && ranges[1] < 0x80) {
```
cushon, line 130: More agentic code review feedback:

These checks prevent the Vector SPI from handling classes with 3 or more ranges (e.g., alphanumeric [a-zA-Z0-9] which has 3 ranges, ranges.length == 6). A Vector processor can check 3 or 4 ranges in parallel by loading range bounds into vector registers and performing parallel comparisons.

Consider generalizing the check in Utf8InputScanner.java to pass any ASCII-bounded range array to the provider:

```java
  private int indexOfAsciiRanges(CharClassScanInfo scanInfo, int start) {
    // Check if the entire range array is ASCII-bounded in O(1)
    if (scanInfo.isAscii) {
      if (scanProvider != null && length - start >= scanProvider.minimumInputLength()) {
        // Pass ranges and the pre-calculated isAscii metadata down to the provider
        int idx = scanProvider.indexOfAsciiClass(bytes, offset, length, scanInfo.ranges, true, start);
        if (idx != -2) { // Allow provider to reject if it can't handle the range size
          return idx;
        }
      }
    }
    // Fallback to scalar matchers
    ...
  }
```

This allows the Vector provider to leverage its register size to match larger sets of classes, fallback selectively by returning -2, and keeps Utf8InputScanner's dispatch logic clean.

**edited**: switched to `scanInfo.isAscii` bit, where that is computed in `CharClassScanInfo` to avoid repeating that work, as a micro optimization

eaftan:
@cushon you’re fast 😁 I pushed a revised version that removes the SPI and bundles the implementation into a multi-release JAR. People can now try it using the normal SafeRE artifact by passing `--add-modules=jdk.incubator.vector -Dorg.safere.experimental.utf8ScanProvider=vector` to Java. This also lets us ship JDK-specific implementations under `META-INF/versions/<jdk>` if the Vector API changes between releases.

Totally agreed that supporting ASCII classes with three or more ranges is worth exploring, perhaps as a follow-up alongside Vector-based literal scanning.

Part of what I’m trying to accomplish with this PR is to create a framework where we can continue experimenting with Vector optimizations without destabilizing the rest of SafeRE or committing to a public SPI prematurely.

WDYT of this version?

safere/src/main/java-jdk26/org/safere/IncubatorVectorScanProvider.java
```diff
@@ -0,0 +1,72 @@
...
+  private static final int SCALAR_PROLOGUE_LENGTH = Integer.BYTES;
+  private static final VectorSpecies<Byte> SPECIES = ByteVector.SPECIES_PREFERRED;
+
+  @Override
+  public int minimumInputLength() {
+    return MINIMUM_INPUT_LENGTH;
+  }
+
+  @Override
+  public int indexOfAsciiClass(byte[] bytes, int offset, int length, int[] ranges, int start) {
```
cushon, line 27: When investigating the Vector API I found `ByteVector` scans were slightly slower than `ShortVector` on long inputs due to nested loop overhead.

Consider adding a loop-unrolled fast-path for single-range classes (`ranges.length == 2`, which matches `[0-9]+` or `\d+`). It allows the JIT to bind the low/high bounds directly to CPU registers, avoiding range array accesses and nested loop counters inside the main SIMD loop:

```java
if (ranges.length == 2) {
  ByteVector low = ByteVector.broadcast(SPECIES, (byte) ranges[0]);
  ByteVector high = ByteVector.broadcast(SPECIES, (byte) ranges[1]);
  for (; pos <= limit; pos += vectorLen) {
    ByteVector inputVec = ByteVector.fromArray(SPECIES, bytes, pos); // changed 'value' to 'bytes'
    VectorMask<Byte> mask = inputVec.compare(VectorOperators.GE, low)
                                    .and(inputVec.compare(VectorOperators.LE, high));
    if (mask.anyTrue()) {
      return pos + mask.firstTrue();
    }
  }
}
```

cushon:
> Part of what I’m trying to accomplish with this PR is to create a framework where we can continue experimenting with Vector optimizations without destabilizing the rest of SafeRE or committing to a public SPI prematurely.
>
> WDYT of this version?

Thanks, this all looks great.

I looked a bit more at ways to use the Vector APIs for String APIs, including without relying on String internals (I'll share a draft PR). I think everything here could work as a basis for that as well, possibly with another method in the Vector scan provider API or two.

If the optimizations for the String API work out perhaps this could be renamed to be less UTF-8 centric (e.g. 'Utf8ScanProvider' to 'VectorScanProvider'), but that could happen later.

eaftan:
> Thanks, this all looks great.
> 
> I looked a bit more at ways to use the Vector APIs for String APIs, including without relying on String internals (I'll share a draft PR). I think everything here could work as a basis for that as well, possibly with another method in the Vector scan provider API or two.

That would be great.  I saw you posted a couple draft PRs along these lines.  Let me know when they're ready for feedback.

> If the optimizations for the String API work out perhaps this could be renamed to be less UTF-8 centric (e.g. 'Utf8ScanProvider' to 'VectorScanProvider'), but that could happen later.

I went ahead and renamed to `VectorScanProvider` as you suggested, I didn't intend to limit this to UTF-8.

safere/src/main/java-jdk26/org/safere/IncubatorVectorScanProvider.java
eaftan: I wasn't able to reproduce the gains from this approach.  Right now I'm working on a Mac Silicon machine, and maybe Vector operations work differently on this device than on Intel?  I filed issue #659 to follow up on this.

safere/src/main/java/org/safere/Utf8InputScanner.java
eaftan: I could have sworn I filed an issue for this yesterday, but I don't see it.  Filed #660

eaftan:
@cushon I'm going to merge this after CI passes.  We can try out more Vector-based optimizations on top of this.

wendigo:
Why only JDK26+, not 25+ ?

eaftan:
> Why only JDK26+, not 25+ ?

Not for any good reason, I haven't gotten around to it yet.  I think what I did here will probably just work on JDK 21+ and so maybe none of the MR-JAR stuff is necessary.  But I need to test it first.  I filed #658 to track that work.

PR #655, opened 2026-07-31
Optimize DFA search loops when returning to start state (DFA Loop Acceleration)

cushon:
Accelerate DFA loops for unanchored searches, allowing the DFA loop to use start acceleration paths when it re-enters the start state.

To avoid regressions for cases that don't benefit from this, the approach
* Restricts the inner loop to scan in 64-character chunks, checking for start-state transitions only when exiting the chunk, to minimize branching in the core DFA loop
* If a fast-forward jump is <64 characters, the engine applies a 512-character penalty during which breakouts are disabled, preventing repeated use of the acceleration paths from degrading performance for dense candidate prefixes

```
./run-java-benchmarks.sh --fastbuild CrossEngineBenchmark.run -- -p crossEngineTrial=FixedOffsetAccelerationBenchmark.findTrulySparse.511@safere-string
```

| Description | Score (ns/op) | Speedup |
| :--- | :--- | :--- |
| Baseline | `64,502.967 ± 1858.615` | 1.00x |
| DFA loop acceleration | **`33,247.480 ± 123.805`** | **1.94x faster** |

### DIFF
--- a/safere-crosscheck/pom.xml
+++ b/safere-crosscheck/pom.xml
@@ -132,6 +132,7 @@
                         <exclude name="SafeReDiagnosticsTest.java"/>
                         <exclude name="SearchScalingRegressionTest.java"/>
                         <exclude name="SimplifierTest.java"/>
+                        <exclude name="StartAcceleratorDifferentialTest.java"/>
                         <exclude name="StartAcceleratorTest.java"/>
                         <exclude name="UnicodeJdkConsistencyTest.java"/>
                         <exclude name="UnicodeTablesTest.java"/>
--- a/safere-fuzz/src/test/java/org/safere/fuzz/Utf8InputFuzzer.java
+++ b/safere-fuzz/src/test/java/org/safere/fuzz/Utf8InputFuzzer.java
@@ -30,6 +30,7 @@ final class Utf8InputFuzzer {
     assertFixedOffsetAccelerationMatchesString(data);
     assertMultiOffsetLiteralOccurrencesMatchString(data);
     assertFinalLineTerminatorEndAnchorsMatchJdk(data);
+    assertPositionDependentStartAccelerationMatchesJdk(data);
     String repeatedLiteral =
         String.valueOf((char) data.consumeInt('A', 'Z')).repeat(data.consumeInt(2, 32));
     String suffix = new String(data.consumeBytes(data.consumeInt(0, 64)), StandardCharsets.UTF_8);
@@ -135,6 +136,24 @@ final class Utf8InputFuzzer {
     boolean utf8Found = safeRe.find(Utf8Input.validated(input.getBytes(StandardCharsets.UTF_8)));
     if (stringFound != expected || utf8Found != expected) {
       throw new AssertionError("end anchor differs from JDK for final line terminator");
+    }
+  }
+
+  private static void assertPositionDependentStartAccelerationMatchesJdk(FuzzedDataProvider data) {
+    String terminator = data.pickValue(List.of("\n", "\r", "\r\n", "\u0085", "\u2028", "\u2029"));
+    String regex = "(b|(?m:^a))c[0-9]";
+    Pattern pattern = Pattern.compile(regex);
+    for (String input :
+        List.of(
+            terminator + "acz9y",
+            " ycby ",
+            "y0c" + terminator + "1bac19c1__19x y_",
+            "bxc\rca9\né\nac9é\r0x\r9a0")) {
+      boolean expected = java.util.regex.Pattern.compile(regex).matcher(input).find();
+      Utf8Input utf8 = Utf8Input.validated(input.getBytes(StandardCharsets.UTF_8));
+      if (pattern.matcher(input).find() != expected || pattern.find(utf8) != expected) {
+        throw new AssertionError("DFA-loop start acceleration differs from JDK anchor semantics");
+      }
     }
   }
 
--- a/safere/src/main/java/org/safere/AcceleratorPolicy.java
+++ b/safere/src/main/java/org/safere/AcceleratorPolicy.java
@@ -1,0 +1,44 @@
+// This file is part of a Java port of RE2 (https://github.com/google/re2).
+// Original RE2 code is Copyright (c) 2009 The RE2 Authors.
+// Modifications and Java port Copyright (c) 2026 Eddie Aftandilian.
+// Licensed under the BSD 3-Clause License (see LICENSE file).
+
+package org.safere;
+
+/**
+ * Immutable tuning policy and diagnostic metadata for start-position and DFA accelerators.
+ *
+ * @param minProfitableSkip Minimum skip distance (in chars or bytes) required for this accelerator
+ *     to be profitable over direct scalar DFA execution.
+ * @param strikeBudget Number of consecutive sub-threshold skips tolerated before declaring adaptive
+ *     defeat.
+ * @param isExactMatchCandidate Whether this accelerator identifies an exact candidate match start
+ *     that can be validated directly.
+ * @param strategy The diagnostic strategy associated with this accelerator, or {@code null} if
+ *     none.
+ */
+record AcceleratorPolicy(
+    int minProfitableSkip,
+    int strikeBudget,
+    boolean isExactMatchCandidate,
+    MatchStrategy strategy) {
+
+  // TODO: Conduct systematic empirical micro-benchmarking across diverse CPU architectures (x86
+  // AVX-512/AVX2, ARM Neon) to precisely tune minimum profitable skip thresholds.
+  // TODO: Measure defeat sensitivity across varied payload densities and evaluate whether dynamic
+  // backoff or payload-length-proportional strike budgets outperform static constant budgets.
+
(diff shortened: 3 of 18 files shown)


cushon:
This came out of looking for opportunities to improve the `String` API as part of the discussion around RE# and the Vector API.

I think this deserves more comprehensive benchmarking to decide if it's general enough to be worth adding, and whether it could regress existing cases.

cushon:
I refreshed this on top of the start accelerating refactoring in https://github.com/eaftan/safere/commit/5e36e9eb611e4e13cdf2e4bcc85b47d4ecd42d4f, and have made some more improvements to the approach.

Currently `StartAccelerator` is used once per top-level `Matcher.find()` call.

This optimization uses start acceleration inside the DFA loop to accelerate searches if the DFA returns to the start state. To avoid performance regressions on dense matches where the overhead of the accelerators doesn't pay off, the DFA tracks consecutive short skips and disables this optimization if the start accelerator is skipping small amounts of input.

In addition to using start acceleration when the DFA returns to the start state, this also adds a self-loop optimization that looks for patterns with lazy or non-greedy repetitions (e.g. `(?s)<block>.*?</block>`, `<<tag.*?>>`) that create DFA states where almost all byte transitions (>250 out of 256) loop back to the same state, with only 1, 2, or 3 escape characters leading to the delimiter. The optimization extracts those escape characters into a `StateAccelerator` associated with the DFA state, and in the DFA loop the accelerator is used to do a fast scan instead of stepping through DFA states.

This was modeled after Rust's `regex-automata` and similar optimizations in Hyperscan. Both have self-loop optimizations and do acceleration in the DFA loop, attaching accelerators to DFA states that escape on <= 3 characters was borrowed from `regex-automata`. The current SafeRE heuristics are a little different and are simplified to try to keep the DFA loop fast for cases that don't benefit.

The exact heuristics and thresholds here would benefit from more comparative benchmarking and tuning, but this seems like a promising direction for sparse matches on large inputs.

```
SCALING_1MB_TRIALS=SearchScalingBenchmark.searchEasyFail.1048576@safere-string,SearchScalingBenchmark.searchEasyFail.1048576@safere-utf8,SearchScalingBenchmark.searchEasySuccess.1048576@safere-string,SearchScalingBenchmark.searchEasySuccess.1048576@safere-utf8,SearchScalingBenchmark.searchMediumFail.1048576@safere-string,SearchScalingBenchmark.searchMediumFail.1048576@safere-utf8,SearchScalingBenchmark.searchHardFail.1048576@safere-string,SearchScalingBenchmark.searchHardFail.1048576@safere-utf8,SearchScalingBenchmark.findIngScaled.1048576@safere-string,SearchScalingBenchmark.findIngScaled.1048576@safere-utf8,Issue481ScalingBenchmark.splitWords.1048576@safere-string,Issue481ScalingBenchmark.blockFind.1048576@safere-string,Issue481ScalingBenchmark.blockFind.1048576@safere-utf8,Issue481ScalingBenchmark.blockFindNegative.1048576@safere-string,Issue481ScalingBenchmark.blockFindNegative.1048576@safere-utf8,Issue481ScalingBenchmark.tagFind.1048576@safere-string,Issue481ScalingBenchmark.tagFind.1048576@safere-utf8,Issue481ScalingBenchmark.tagFindNegative.1048576@safere-string,Issue481ScalingBenchmark.tagFindNegative.1048576@safere-utf8,Issue481ScalingBenchmark.schemeExtract.1048576@safere-string,Issue481ScalingBenchmark.schemeFindNegative.1048576@safere-string,Issue481ScalingBenchmark.schemeFindNegative.1048576@safere-utf8,Issue481ScalingBenchmark.schemeReplace.1048576@safere-string
```

```
./run-java-benchmarks.sh --fastbuild CrossEngineScalingBenchmark -- -p crossEngineScalingTrial=SCALING_1MB_TRIALS
```

| Workload | `main` (Baseline) | `dfa-loop-acceleration` | Speedup | Category |
| :--- | :---: | :---: | :---: | :--- |
| `blockFind.1048576@safere-string` | 8,908.2 μs | **55.6 μs** | **160.3× faster** | Self-Loop Escape |
| `blockFind.1048576@safere-utf8` | 3,189.0 μs | **88.5 μs** | **36.0× faster** | Self-Loop Escape |
| `blockFindNegative.1048576@safere-string` | 1,967.3 μs | **18.0 μs** | **109.0× faster** | Start Acceleration |
| `blockFindNegative.1048576@safere-utf8` | 1,969.9 μs | **72.5 μs** | **27.2× faster** | Start Acceleration |
| `tagFind.1048576@safere-string` | 141.4 μs | **104.9 μs** | **1.35× faster** | Self-Loop Escape |
| `tagFind.1048576@safere-utf8` | 182.1 μs | **176.3 μs** | **1.03× faster** | Self-Loop Escape |
| `schemeFindNegative.1048576@safere-string` | 119.3 μs | **103.9 μs** | **1.15× faster** | Start Acceleration |
| `schemeFindNegative.1048576@safere-utf8` | 102.6 μs | **103.1 μs** | Neutral | Start Acceleration |
| `searchEasyFail.1048576@safere-utf8` | 160.7 μs | **161.4 μs** | Neutral | Start Acceleration |
| `searchEasySuccess.1048576@safere-utf8` | 163.7 μs | **163.8 μs** | Neutral | Start Acceleration |
| `searchMediumFail.1048576@safere-utf8` | 157.9 μs | **156.4 μs** | Neutral | Start Acceleration |
| `searchHardFail.1048576@safere-utf8` | 159.1 μs | **156.6 μs** | Neutral | Start Acceleration |
| `findIngScaled.1048576@safere-string` | 8,607.0 μs | **8,402.0 μs** | Neutral | Multi-match Word |
| `findIngScaled.1048576@safere-utf8` | 9,281.8 μs | **8,797.1 μs** | Neutral | Multi-match Word |
| `splitWords.1048576@safere-string` | 18,911.0 μs | **18,887.7 μs** | Neutral | Control (Dense match) |

cushon:
@eaftan I would be interested in your review of this and your thoughts on https://github.com/eaftan/safere/pull/655#issuecomment-5264561323 when you have a chance.

This needs some more benchmarking / tuning to make sure that the cases that don't benefit aren't regressing, but overall I think this is a valuable category of optimizations to add for sparse matches on long inputs.

cushon:
I added a commit (https://github.com/eaftan/safere/pull/655/changes/f67106ef873c0816bcbd22a29cc050f3e10a8451) that I think addresses #711 and also more test coverage, thanks for the catch!

eaftan (commented):
This looks like a good optimization, and your latest commit fixes the start-state reindexing problem from #711. I did find two other correctness/control boundaries: the self-loop analysis proved only ASCII transitions but could scan across non-ASCII input, which could skip valid BMP or supplementary-code-point transitions, and disabling start acceleration still installed the cached DFA accelerators used by this PR. Policy selection could also use the inactive String/UTF-8 substrate.

I've pushed a commit that fixes those issues and adds String, UTF-8, generated, and fuzz-regression coverage. 

I'm going to do some benchmarking to get data on threshold tuning.  Will get back to you on that.

eaftan (approved):
Following up on the threshold-tuning benchmarks I mentioned above, I first reproduced the target gains compared to current main:

| Benchmark | main (us/op) | PR + fixes (us/op) | Change |
|---|---:|---:|---:|
| `blockFind.1048576@safere-string` | 2,795.749 | 45.830 | 61.0x faster |
| `blockFindNegative.1048576@safere-string` | 1,305.596 | 23.313 | 56.0x faster |
| `searchMediumFail.1048576@safere-string` | 0.026 | 0.027 | neutral |

I then swept the policies using controlled 1 MiB rejected inputs at candidate gaps of 8, 16, 32, and 64 characters, both String and UTF-8 where applicable, plus the existing 128- and 1,024-character #481 workloads. This table reports geometric-mean time relative to the current setting, so lower is better:

| Policy | Values tested | Result relative to current | Recommendation |
|---|---|---|---|
| Literal minimum skip | 8 / current 16 | 0.993 / 1.000; no measurable improvement | Retain 16 |
| Character-class minimum skip | 8 / current 24 / 64 | 0.998 / 1.000 / 1.006; neutral | Retain 24 |
| Character-class strike budget | 1 / current 3 / 8 | 0.997 / 1.000 / 0.997; neutral | Retain 3 |
| Line-anchor minimum skip | 8 / current 16 / 64 | 0.983 / 1.000 / 0.992; no measurable improvement | Retain 16 |
| Line-anchor strike budget | 1 / current 3 / 8 | 1.001 / 1.000 / 0.996; neutral | Retain 3 |
| Literal/self-loop short-input cutoffs | 8 / current 16 / 32 / 64 | 0.984 / 1.000 / 0.984 / 0.967; no consistent winner | Retain 16 |
| Maximum self-loop escapes | 1 / 2 / current 3 | Cap 2 is 9.16x faster than cap 1 for two escapes; cap 3 is 6.53x faster than cap 2 for three escapes, with smaller cases neutral | Retain 3 |

My recommendation is to retain the current thresholds. The three-escape cap is directly supported by the data. The other constants are reasonable defaults, but the alternatives were neutral rather than establishing a unique optimum on this x86-64 machine, so I would also leave the cross-architecture tuning TODO in place.

LGTM

Issue #660, opened 2026-08-02
Explore Vector scanning for ASCII classes with three or more ranges


eaftan:
## Context

PR #654's experimental Vector scanner only receives ASCII classes in a few specialized shapes:

- one singleton, represented as one degenerate range;
- two singleton values, represented as two degenerate ranges; or
- one contiguous range.

The dispatch checks in `Utf8InputScanner` therefore exclude classes with three or more ranges.
For example, `[a-zA-Z0-9]` has three ranges (`ranges.length == 6`) and falls back to the existing
non-Vector scanner even though all of its values are ASCII.

A Vector implementation may be able to evaluate three or four ranges efficiently by keeping the
bounds in vector registers, performing the comparisons in parallel, and combining their masks.
This could extend Vector acceleration to common classes such as alphanumeric identifiers.

## Proposed investigation

- Generalize dispatch so the provider can be offered any class known to be entirely ASCII, while
  preserving a cheap fallback when a particular range count or shape is unsupported.
- Prototype Vector implementations for three and four ranges. Compare at least:
  - direct per-range comparisons with combined masks;
  - layouts that load or broadcast multiple low/high bounds outside the input loop; and
  - the current SWAR/general ASCII-class fallback.
- Cover representative class shapes, including `[a-zA-Z0-9]`, ranges with gaps of different
  sizes, degenerate singleton ranges mixed with proper ranges, and negated or normalized classes
  where applicable.
- Benchmark first-match and repeated scan-all behavior across input lengths, match positions, and
  candidate densities. Run Vector benchmark variants sequentially and use `--long` to confirm
  close or surprising results.
- Verify the complete SafeRE UTF-8 matching path, not only the low-level scan loop.
- Check both AArch64 and x86-64 before selecting a production implementation because Vector API
  lowering and register pressure may differ by architecture.

The provider should be allowed to decline shapes it cannot handle efficiently so unsupported or
unprofitable classes continue to use the existing fallback without changing matching behavior.

Part of #504. Follow-up to #654.

eaftan:
This was implemented in #661.

Issue #659, opened 2026-08-02
Benchmark single-range Vector scan specialization on x86-64


eaftan:
## Context

PR #654's current JDK 26 Vector scanner handles singleton, pair, and single-range ASCII
classes through one generic helper. A review suggestion proposed a dedicated fast path for
single-range classes (`ranges.length == 2`), such as `[0-9]` or `\\d`:

- read the low and high bounds once before the SIMD loop;
- broadcast them into `ByteVector`s;
- compare each input vector directly against those bounds; and
- avoid the generic shape checks and range-array accesses in the main loop.

The hypothesis is that this may reduce nested/helper-loop overhead, particularly on x86-64.

## AArch64 exploration

We compared four benchmark-only implementations on an Apple AArch64 machine using OpenJDK
26.0.1 at SafeRE commit `de9c043daf738911bc755ae25aecdbbbad63f076`:

1. the current generic production loop shape;
2. a single-range loop with scalar low/high bounds hoisted before iteration;
3. the proposed single-range loop with explicit `ByteVector.broadcast` bounds; and
4. an additional experiment that processed two vectors per loop iteration.

The standard benchmark matrix covered 1 KiB and 32 KiB inputs, aligned and offset starts,
absent and late matches, and a dense first-position match. It found no meaningful advantage
for hoisting or broadcasting. Processing two vectors per iteration was effectively tied on
32 KiB scans but regressed the 1 KiB absent scan by about 16.7%.

A long confirmation run for the exact generic-versus-broadcast comparison produced:

| Trial | Current generic | Specialized/broadcast | Difference |
| --- | ---: | ---: | ---: |
| absent, 1 KiB | 20.019 ns/op | 19.940 ns/op | statistically tied |
| absent, 32 KiB | 611.714 ns/op | 613.029 ns/op | 0.2% slower |
| late match, 32 KiB | 610.920 ns/op | 614.996 ns/op | 0.7% slower |

This suggests that JDK 26's AArch64 compiler already optimizes the current helper sufficiently.
No production change was retained from the experiment.

## Next steps

- Recreate the generic and dedicated single-range variants in the Vector benchmark harness.
- Run the same controlled matrix on an x86-64/Intel machine with JDK 26. Use
  `./run-vector-benchmarks.sh` and run variants sequentially.
- Use the standard configuration first, then `--long` to confirm any close or surprising result.
- If x86-64 shows a reproducible improvement, inspect generated code or profile the loops to
[...]

PR #661, opened 2026-08-02
Extend Vector scans to multi-range ASCII classes

eaftan:
## Summary

- offer all normalized ASCII character classes to the experimental Vector provider and let the
  provider decline unsupported shapes cheaply
- support one through four ASCII ranges with direct Vector comparisons while retaining a bounded
  bitmap scalar prologue for dense matches
- extend the Vector benchmark corpus and harness with three-range alphanumeric, mixed four-range,
  retained-bound, and repeated scan-all coverage
- add alignment, tail, and unsupported five-range fallback tests

Refs #660

## AArch64 benchmark results

JDK 26.0.1 on AArch64. Times are nanoseconds per operation; lower is better.

Long end-to-end first-hit confirmation:

| Trial | SWAR | Vector | Improvement |
| --- | ---: | ---: | ---: |
| `[0-9A-Za-z]`, absent, 32 KiB | 9,263.056 | 1,551.326 | 5.97x |
| `[0-9A-Za-z]`, late, 32 KiB, offset 7 | 9,884.072 | 1,557.448 | 6.35x |
| `[!#0-9A-Z]`, absent, 32 KiB | 9,103.980 | 1,554.491 | 5.86x |
| `[!#0-9A-Z]`, late, 32 KiB, offset 7 | 10,144.201 | 1,570.777 | 6.46x |

Standard end-to-end repeated scan-all measurements:

| Trial | SWAR | Vector | Improvement |
| --- | ---: | ---: | ---: |
| `[0-9A-Za-z]`, sparse, 32 KiB | 38,548.046 | 16,601.162 | 2.32x |
| `[0-9A-Za-z]`, dense, 32 KiB | 622,112.729 | 530,954.750 | 1.17x |
| `[!#0-9A-Z]`, sparse, 32 KiB | 45,281.047 | 21,474.850 | 2.11x |
| `[!#0-9A-Z]`, dense, 32 KiB | 736,524.573 | 647,962.667 | 1.14x |

The benchmark-only retained bound-vector layout was slower than direct scalar-bound comparisons on
this platform, so the production implementation uses direct comparisons.

## Verification

Completed:

- `mvn -pl safere test -q`
- `mvn -pl safere test -Pvector-tests -q`
- `mvn package -DskipTests -Dpmd.skip=true -Dspotless.check.skip=true -q -f safere-vector-benchmarks/pom.xml`
- `mvn spotless:check -pl safere -q`
- `mvn spotless:check -f safere-vector-benchmarks/pom.xml -q`
- `git diff --check`
- `./run-vector-benchmarks.sh --long --fastbuild --end-to-end` for the four first-hit trials above
- standard end-to-end Vector benchmarks for the four repeated scan-all trials above
- Codex branch review against `main`; no P0-P2 findings

Not run:

- x86-64 Vector benchmarks; confirm that the new three- and four-range implementations are
  profitable on that architecture
- full native/external benchmark collection, which is outside this experimental Vector-only change
- public API crosscheck suite

### DIFF
--- a/safere/src/main/java-jdk26/org/safere/IncubatorVectorScanProvider.java
+++ b/safere/src/main/java-jdk26/org/safere/IncubatorVectorScanProvider.java
@@ -15,7 +15,6 @@ import jdk.incubator.vector.VectorSpecies;
 /** Experimental scan operations implemented with the incubating Vector API. */
 final class IncubatorVectorScanProvider implements VectorScanProvider {
   private static final int MINIMUM_INPUT_LENGTH = 1024;
-  private static final int SCALAR_PROLOGUE_LENGTH = Integer.BYTES;
   private static final VectorSpecies<Byte> SPECIES = ByteVector.SPECIES_PREFERRED;
 
   @Override
@@ -25,13 +24,10 @@ final class IncubatorVectorScanProvider implements VectorScanProvider {
 
   @Override
   public int indexOfAsciiClass(byte[] bytes, int offset, int length, int[] ranges, int start) {
-    int position = Math.max(0, start);
-    int scalarLimit = Math.min(length, position + SCALAR_PROLOGUE_LENGTH);
-    for (; position < scalarLimit; position++) {
-      if (matches(bytes[offset + position], ranges)) {
-        return position;
-      }
+    if (ranges.length < 2 || ranges.length > 8 || (ranges.length & 1) != 0) {
+      return UNSUPPORTED;
     }
+    int position = Math.max(0, start);
     int limit = position + SPECIES.loopBound(length - position);
     for (; position < limit; position += SPECIES.length()) {
       ByteVector values = ByteVector.fromArray(SPECIES, bytes, offset + position);
@@ -49,11 +45,22 @@ final class IncubatorVectorScanProvider implements VectorScanProvider {
   }
 
   private static VectorMask<Byte> matches(ByteVector values, int[] ranges) {
-    if (ranges.length == 4) {
-      return values.eq((byte) ranges[0]).or(values.eq((byte) ranges[2]));
+    VectorMask<Byte> matches = matches(values, ranges[0], ranges[1]);
+    if (ranges.length >= 4) {
+      matches = matches.or(matches(values, ranges[2], ranges[3]));
     }
-    byte low = (byte) ranges[0];
-    byte high = (byte) ranges[1];
+    if (ranges.length >= 6) {
+      matches = matches.or(matches(values, ranges[4], ranges[5]));
+    }
+    if (ranges.length == 8) {
+      matches = matches.or(matches(values, ranges[6], ranges[7]));
+    }
+    return matches;
+  }
+
+  private static VectorMask<Byte> matches(ByteVector values, int lowBound, int highBound) {
+    byte low = (byte) lowBound;
+    byte high = (byte) highBound;
     if (low == high) {
       return values.eq(low);
     }
@@ -64,9 +71,11 @@ final class IncubatorVectorScanProvider implements VectorScanProvider {
   }
 
   private static boolean matches(byte value, int[] ranges) {
-    if (ranges.length == 4) {
-      return value == (byte) ranges[0] || value == (byte) ranges[2];
+    for (int index = 0; index < ranges.length; index += 2) {
+      if (value >= (byte) ranges[index] && value <= (byte) ranges[index + 1]) {
+        return true;
+      }
     }
-    return value >= (byte) ranges[0] && value <= (byte) ranges[1];
+    return false;
   }
 }
--- a/safere/src/main/java-jdk26/org/safere/IncubatorVectorScanProvider.java
+++ b/safere/src/main/java-jdk26/org/safere/IncubatorVectorScanProvider.java
@@ -5,8 +5,7 @@
 
 package org.safere;
 
-import static jdk.incubator.vector.VectorOperators.GE;
-import static jdk.incubator.vector.VectorOperators.LE;
+import static jdk.incubator.vector.VectorOperators.ULE;
 
 import jdk.incubator.vector.ByteVector;
 import jdk.incubator.vector.VectorMask;
@@ -67,7 +66,9 @@ final class IncubatorVectorScanProvider implements VectorScanProvider {
     if (high == low + 1) {
       return values.eq(low).or(values.eq(high));
     }
-    return values.compare(GE, low).and(values.compare(LE, high));
+    ByteVector lowVector = ByteVector.broadcast(SPECIES, low);
+    ByteVector rangeVector = ByteVector.broadcast(SPECIES, (byte) (high - low));
+    return values.sub(lowVector).compare(ULE, rangeVector);
   }
 
   private static boolean matches(byte value, int[] ranges) {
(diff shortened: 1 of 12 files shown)


safere/src/main/java-jdk26/org/safere/IncubatorVectorScanProvider.java
```diff
@@ -49,11 +45,22 @@ public int indexOfAsciiClass(byte[] bytes, int offset, int length, int[] ranges,
...
+    if (ranges.length >= 6) {
+      matches = matches.or(matches(values, ranges[4], ranges[5]));
+    }
+    if (ranges.length == 8) {
+      matches = matches.or(matches(values, ranges[6], ranges[7]));
+    }
+    return matches;
+  }
+
+  private static VectorMask<Byte> matches(ByteVector values, int lowBound, int highBound) {
```
cushon, line 60: Agentic code review feedback

---

We can optimize this range comparison to use a single unsigned subtraction operator (`VectorOperators.ULE`) rather than executing both `GE` and `LE` comparisons and merging them. 

For a range `[low, high]`, the logic `(value - low) <= (high - low)` can be vectorized as:

```java
private static VectorMask<Byte> matches(ByteVector values, int lowBound, int highBound) {
  byte low = (byte) lowBound;
  byte high = (byte) highBound;
  if (low == high) {
    return values.eq(low);
  }
  ByteVector lowVec = ByteVector.broadcast(SPECIES, low);
  ByteVector rangeVec = ByteVector.broadcast(SPECIES, (byte) (high - low));
  return values.sub(lowVec).compare(VectorOperators.ULE, rangeVec);
}

**Comment by eaftan** [comment]
Follow-up review and x86-64 benchmark update:

The review found that the low-level SWAR benchmark wrapper rebuilt both ASCII bitmaps inside every
timed scanner invocation. This especially biased repeated scan-all comparisons because bitmap
construction was repeated after every match. Commit 8d48a0c now precomputes the bitmaps during JMH
trial setup. The end-to-end results below use the production provider path and were not affected by
that wrapper issue.

The additional machine is an Intel Core i7-11700K running OpenJDK 26.0.2 on x86-64. Its preferred
byte species is `Species[byte, 64, S_512_BIT]`. The original AArch64 machine used
`Species[byte, 16, S_128_BIT]`.

Long end-to-end first-hit confirmation on x86-64 (nanoseconds per operation):

| Trial | SWAR | Vector | Improvement |
| --- | ---: | ---: | ---: |
| `[0-9A-Za-z]`, absent, 32 KiB | 15,703.319 | 806.967 | 19.46x |
| `[0-9A-Za-z]`, late, 32 KiB, offset 7 | 15,732.176 | 830.285 | 18.95x |
| `[!#0-9A-Z]`, absent, 32 KiB | 15,688.474 | 895.672 | 17.52x |
| `[!#0-9A-Z]`, late, 32 KiB, offset 7 | 15,757.545 | 1,187.740 | 13.27x |

Standard end-to-end repeated scan-all measurements on x86-64:

| Trial | SWAR | Vector | Change |
| --- | ---: | ---: | ---: |
| `[0-9A-Za-z]`, sparse, 32 KiB | 59,204.911 | 26,985.242 | 2.19x faster |
| `[0-9A-Za-z]`, dense, 32 KiB | 839,014.134 | 857,545.020 | 2.2% slower |
| `[!#0-9A-Z]`, sparse, 32 KiB | 65,081.519 | 33,420.942 | 1.95x faster |
| `[!#0-9A-Z]`, dense, 32 KiB | 1,109,891.207 | 1,097,138.823 | 1.2% faster |

Comparison with the original AArch64 speedups:

| Trial | AArch64, 128-bit | x86-64, 512-bit |
| --- | ---: | ---: |
| Alphanumeric absent, first hit | 5.97x | 19.46x |
| Alphanumeric late, first hit | 6.35x | 18.95x |
| Mixed four-range absent, first hit | 5.86x | 17.52x |
| Mixed four-range late, first hit | 6.46x | 13.27x |
| Alphanumeric sparse, scan all | 2.32x | 2.19x |
| Alphanumeric dense, scan all | 1.17x | 0.98x |
| Mixed four-range sparse, scan all | 2.11x | 1.95x |
| Mixed four-range dense, scan all | 1.14x | 1.01x |

The 512-bit x86 species provides much larger first-hit gains, while sparse scan-all remains close
to the AArch64 result. Dense scan-all is effectively neutral on x86-64. The dense inputs place a
match within the provider's four-byte scalar prologue, so those trials primarily compare the
enabled provider's bounded scalar fast path with the general fallback; they do not measure Vector
body throughput. The first-hit and sparse results do exercise the multi-range Vector loop and
confirm that it is profitable on both architectures.

Verification after the benchmark fix:

- reactor tests through `safere-benchmarks`
- `mvn -pl safere test -Pvector-tests -q`
- standalone JDK 26 Vector benchmark packaging
- Vector benchmark Spotless check
- `git diff --check`
- final Codex branch review; no remaining actionable findings

File: safere/src/main/java-jdk26/org/safere/IncubatorVectorScanProvider.java (hunk 1 above)
Comment (eaftan) [reply safere/src/main/java-jdk26/org/safere/IncubatorVectorScanProvider.java:60]: Thanks, good suggestion. I tried the unsigned-subtraction formulation and adopted it.

On x86-64/OpenJDK 26.0.2 with 512-bit byte vectors, long end-to-end first-hit benchmarks improved as follows:

| Trial | Previous | ULE | Change |
|---|---:|---:|---:|
| Alphanumeric absent | 806.967 ns/op | 659.068 ns/op | 18.3% faster |
| Alphanumeric late | 830.285 ns/op | 686.108 ns/op | 17.4% faster |
| Mixed four-range absent | 895.672 ns/op | 724.251 ns/op | 19.1% faster |
| Mixed four-range late | 1,187.740 ns/op | 807.891 ns/op | 32.0% faster |

File: safere/src/main/java-jdk26/org/safere/IncubatorVectorScanProvider.java [hunk 2]
```diff
@@ -25,13 +24,10 @@ public int minimumInputLength() {
 
   @Override
   public int indexOfAsciiClass(byte[] bytes, int offset, int length, int[] ranges, int start) {
```
cushon, line 26: More agentic feedback (I meant to post this at the same time as the other comment)

---

If `ranges.length == 2` (single-range search like `[0-9]`), we could add a specialized 4-way unrolled loop fast path. Loading and matching 4 consecutive vectors in parallel allows the JIT to better distribute load operations across execution ports, resulting in a ~1.5x speedup on long inputs.

Regarding the regression on small inputs (1 KiB absent scans) mentioned in Issue #659 when unrolling: we resolved that regression in the String vector scanners by combining:
1. A short scalar prologue (centralized in `Utf8InputScanner`, which you have already cleanly relocated in this PR!).
2. Gating entry to the vector path with a minimum input length threshold (e.g. `minimumInputLength() == 1024` or higher for unrolled paths).

With these two safeguards, we can safely unroll the single-range loop to gain peak throughput on long strings without regressing short queries.

eaftan: Thanks for the feedback.  I think this belong as part of #659.  I've added a [comment](https://github.com/eaftan/safere/issues/659#issuecomment-5187212454) there with a reference back to this comment.

Issue #664, opened 2026-08-04
Add missing UTF-8 keyword-alternation fast path


eaftan:
## Summary

SafeRE's Java-string matcher has a specialized keyword-alternation path for patterns such as:

```regex
(?is).*\b(you|your)\b.*
```

The equivalent pre-existing UTF-8 path does not use this optimization. On successful searches,
the String path finds a keyword near the end by scanning backward and remains effectively
constant-time for this workload, while the UTF-8 path runs the general engine across the input
and scales linearly with its length.

This is a performance bug in UTF-8 fast-path coverage, not an encoding-conversion cost: the
benchmark supplies materialized `Utf8Input` before timing.

## Evidence

Standard benchmark collection at SafeRE commit
`516a1ff76141bc83c3910426332c7824acdf007c` measured
`RealWorldRegexBenchmark.runBenchmark.wildcardSearch.match` as follows:

| Input size | SafeRE String | SafeRE UTF-8 | String advantage |
|---:|---:|---:|---:|
| 1K | 55.642 ns/op | 1,802.157 ns/op | 32.4× |
| 10K | 56.651 ns/op | 17,915.469 ns/op | 316× |
| 100K | 51.942 ns/op | 181,041.370 ns/op | 3,486× |

Across the 24 successful-match rows shared by the two representations, String appears 1.58×
faster by geometric mean. Removing these three `wildcardSearch` rows reverses the result: UTF-8
is 1.36× faster. This one missing fast path therefore materially changes the aggregate.

The String path is selected by `Matcher.doFindCore()` through `Pattern.KeywordAlternation` and
`findGreedyWholeInputKeywordAlternation()`. Its current guard requires `text != null`, excluding
`Utf8InputScanner`. `Pattern.find(Utf8Input)` likewise has no equivalent keyword-alternation path.

## Expected work

- Add an engine-native UTF-8 keyword-alternation path for the supported pattern class, including
  greedy whole-input patterns.
- Keep the implementation linear-time and operate directly on pre-existing UTF-8 bytes.
- Preserve Java-compatible ASCII case-insensitive matching, word-boundary behavior, leftmost
  semantics, byte offsets, captures, regions, and matcher state where those APIs expose them.
- Add systematic correctness tests covering successful and failed searches, keyword alternatives,
  input-size scaling, non-ASCII text around boundaries, captures, and region starts.
- Add or retain benchmark coverage demonstrating that successful `wildcardSearch` no longer
[...]

PR #665, opened 2026-08-04
Publish full cross-engine benchmark report and artifacts

eaftan:
## Summary

- replace the benchmark report with the latest standard cross-engine collection and make the
  150-row real-world matrix the primary headline category
- document the Real-world, Core, and Application category membership, weighting, coverage, and
  cross-runtime limitations
- publish the complete raw collection under the benchmarked SafeRE commit, with provenance,
  normalized JSON Lines, and SHA-256 checksums
- teach the collection pipeline to emit a combined `normalized-results.jsonl` artifact for future
  analysis and independent validation

## Results

The report is based on SafeRE commit
`516a1ff76141bc83c3910426332c7824acdf007c` and the command:

```bash
./collect-benchmark-results.sh --cross-language --skip-openjdk-regex
```

Across the complete 150-row real-world matrix, the SafeRE String path is:

- 2.22× faster than the JDK; 1.88× after excluding the two most extreme JDK backtracking rows
- 12.5× faster than RE2/J
- 3.06× faster than RE2-FFM
- 13% slower than native C++ RE2
- 6.56× faster than Go `regexp`
- 2.03× slower than Rust `regex`

PCRE2 JIT and .NET non-backtracking support smaller subsets, which are reported separately with
their actual row counts. The external OpenJDK-derived suite was deliberately skipped and stale
results from that suite were removed rather than carried forward.

The complete 11.9 MB publication contains original JMH and native harness output, engine-specific
JSONL, the resolved declarative plan, 2,949 normalized measurements, generated tables, provenance,
and checksums. Raw and generated artifacts are marked non-diffable so their authenticated bytes are
retained without overwhelming source diffs.

Refs #664

## Verification

Ran:

- `python3 -m unittest discover -s safere-benchmarks/scripts -p 'test_*.py'` — 9 tests passed
- `bash -n collect-benchmark-results.sh`
- JSON/JSONL parsing with `jq`
- regeneration of `normalized-results.jsonl` from the checked-in raw inputs; byte-identical result
- uniqueness check for all 2,949 normalized engine/benchmark identities
- `sha256sum --check SHA256SUMS` — all 17 authenticated files passed
- Markdown table-column validation for `BENCHMARKS.md`
- `git diff --check`

Not run:

- the full SafeRE test suite, because this changes documentation, benchmark collection tooling,
  parser tests, and publication data rather than matching implementation code
- a new benchmark collection after the report/tooling edits; the report and normalized artifact
  are derived from the preserved raw collection described above
- the external OpenJDK-derived benchmark suite, as explicitly noted in the report


eaftan:
@cushon FYI.  This time I included my raw benchmark data in the `benchmark-results` directory so others can poke around, do additional analyses, and look for low-hanging optimization fruit.  

SafeRE is only 13% slower than native RE2 across all 150 real-world measurements, which is closer than I expected. SafeRE takes 2.03× as long as Rust regex, so there is clearly room to improve.

.NET non-backtracking is also closer than I expected: SafeRE is 8% slower over the 114 measurements supported by both engines.

PR #667, opened 2026-08-04
Optimize safere-ffm-re2

cushon:
Various optimizations to `safere-ffm-re2`

* Minimize FFM boundary crossings by implementing bulk string replacement operations (`RE2::Replace` and `RE2::GlobalReplace`) in c++
* Eliminate C++ heap allocations during regex matching by replacing per-call `std::vector` submatch allocations in `re2_find` and `re2_find_all` with stack-allocated `absl::FixedArray` buffers.


eaftan:
I'm looking at this but won't complete my review until tomorrow.

safere-ffm-re2/src/main/java/org/safere/re2ffm/RE2FfmMatcher.java
```diff
@@ -254,25 +222,8 @@ public String replaceAll(String replacement) {
    * @return the result string
    */
   public String replaceFirst(String replacement) {
-    reset();
```
eaftan, line 257: The previous implementation called `reset()` and `find()`, leaving the matcher positioned at the first replacement match. This native delegation leaves `matched`, `matchByteOffsets`, and `searchBytePos` unchanged. If `replaceFirst()` is called after `find()`, subsequent `group()`, `start()`, `end()`, and `find()` observe stale state or continue from the old position. Could this path reset and update matcher state as before?

safere-ffm-re2/src/main/java/org/safere/re2ffm/RE2FfmMatcher.java
```diff
@@ -303,37 +254,6 @@ private static String convertReplacement(String replacement) {
...
-  private void appendRewrite(StringBuilder sb, String rewrite) {
-    for (int i = 0; i < rewrite.length(); i++) {
-      char c = rewrite.charAt(i);
-      if (c == '\\' && i + 1 < rewrite.length()) {
-        char next = rewrite.charAt(i + 1);
-        if (Character.isDigit(next)) {
-          // Parse group number.
-          int groupNum = 0;
-          i++;
-          while (i < rewrite.length() && Character.isDigit(rewrite.charAt(i))) {
```
eaftan, line 316: This changes the meaning of valid multi-digit capture references. For example:

```java
RE2FfmPattern pattern = RE2FfmPattern.compile("(a)(b)(c)(d)(e)(f)(g)(h)(i)(j)");
String result = pattern.matcher("xabcdefghijy").replaceFirst("$10");
```

Java-style `$10` refers to capture group 10, which contains `j`, so the expected result is `xjy`.

`convertReplacement()` changes `$10` to `\10`, but RE2 rewrite syntax consumes only one digit. It interprets `\10` as group 1 `(a)` followed by the literal `0`, producing `xa0y`.

The removed Java-side expansion parsed the complete group number. Could we preserve that behavior for multi-digit capture references?

eaftan (commented):
Thanks for working on this. I believe you’re the primary consumer of RE2-FFM, so I’m happy to accept the performance tradeoffs if these changes improve the workloads that matter to you.

I benchmarked the PR head 117c7a3 against main at e4798a6 using the project benchmark runner and the RE2-FFM string-conversion trials.

The clearest results were:

| Benchmark | Main-ish | PR head | Change |
|---|---:|---:|---:|
| CaptureScalingBenchmark.capture0 | 82.98 ns | 82.65 ns | effectively unchanged |
| CaptureScalingBenchmark.capture1 | 287.46 ns | 274.46 ns | 4.5% faster |
| CaptureScalingBenchmark.capture3 | 333.05 ns | 318.01 ns | 4.5% faster |
| CaptureScalingBenchmark.capture10 | 722.17 ns | 713.56 ns | no clear difference |
| ReplaceBenchmark.literalReplaceFirst | 206.80 ns | 289.47 ns | 40% slower |
| ReplaceBenchmark.literalReplaceFirstNoMatch | 459.35 ns | 770.37 ns | 68% slower |
| ReplaceBenchmark.literalReplaceAll | 669.52 ns | 675.32 ns | effectively unchanged |
| ReplaceBenchmark.digitReplaceAll | 964.65 ns | 945.71 ns | 2.0% faster |

These core results can be reproduced at each commit with:

```bash
trials='CaptureScalingBenchmark.capture0@re2-ffm-string-conversion,CaptureScalingBenchmark.capture1@re2-ffm-string-conversion,CaptureScalingBenchmark.capture3@re2-ffm-string-conversion,CaptureScalingBenchmark.capture10@re2-ffm-string-conversion,ReplaceBenchmark.literalReplaceFirst@re2-ffm-string-conversion,ReplaceBenchmark.literalReplaceFirstNoMatch@re2-ffm-string-conversion,ReplaceBenchmark.literalReplaceAll@re2-ffm-string-conversion,ReplaceBenchmark.digitReplaceAll@re2-ffm-string-conversion'

./run-java-benchmarks.sh --long 'CrossEngineBenchmark.run' -- \
  -p "crossEngineTrial=$trials"
```

My interpretation is that the FixedArray change is a demonstrated optimization for ordinary capture counts, but the replacement changes regress performance overall on these benchmarks.

cushon:
Thanks for the review, I have fixed the issue you caught and made some more tweaks, and I think I have fixed the regressions and am seeing some marginal gains on the benchmarks.

Using the same trials as the results you shared:

| Benchmark | Baseline (Before) | Experiment (After) | Change | Statistically Significant? |
| **`CaptureScaling.capture0`** | 112.8 ns | 114.1 ns | +1.1% | No (within error margin) |
| **`CaptureScaling.capture1`** | 380.5 ns | 393.9 ns | +3.5% | No (within error margin) |
| **`CaptureScaling.capture3`** | 456.1 ns | 452.0 ns | -0.9% | No (within error margin) |
| **`CaptureScaling.capture10`** | 1097.2 ns | 1092.3 ns | -0.4% | No (within error margin) |
| **`literalReplaceFirst`** | 258.9 ns | 246.8 ns | **-4.7%** (faster) | Borderline (high overlap in error) |
| **`literalReplaceFirstNoMatch`**| 642.1 ns | 634.5 ns | -1.2% | No (within error margin) |
| **`literalReplaceAll`** | 907.0 ns | 864.7 ns | **-4.7%** (faster) | **Yes** (clear separation) |
| **`digitReplaceAll`** | 1321.5 ns | 1294.4 ns | **-2.1%** (faster) | **Yes** (clear separation) |

safere-ffm-re2/src/main/java/org/safere/re2ffm/RE2FfmMatcher.java
cushon: Fixed and added test coverage, it now falls back to Java if there are multiple digits

safere-ffm-re2/src/main/java/org/safere/re2ffm/RE2FfmMatcher.java
cushon: Fixed, thanks

eaftan (approved):
Thanks for addressing the two issues from my earlier review.

On re-review, I found a few adjacent compatibility edge cases. The Java fallback did not apply Java's longest-valid-group-prefix rule for references such as `$101` and `$11`; a zero-width match before a supplementary Unicode character advanced by one UTF-8 byte; and a native `replaceAll()` ending in an empty match allowed that match to be rediscovered by a subsequent `find()`. I also removed a hard-coded Debian/x86-64 RPATH.

I've pushed a commit that fixes those issues and adds regression coverage. The full `safere-ffm-re2` test suite passes, and the final review pass found no remaining P2+ issues.

| Benchmark | main (ns/op) | PR+fixes (ns/op) | Change |
|---|---:|---:|---:|
| `capture0` | 87.827 | 87.207 | 0.7% faster |
| `capture1` | 301.332 | 288.890 | 4.1% faster |
| `capture3` | 353.633 | 329.320 | 6.9% faster |
| `capture10` | 754.671 | 733.285 | 2.8% faster |
| `literalReplaceFirst` | 213.514 | 190.252 | 10.9% faster |
| `literalReplaceFirstNoMatch` | 485.900 | 479.755 | 1.3% faster |
| `literalReplaceAll` | 691.898 | 684.127 | 1.1% faster |
| `digitReplaceAll` | 1024.080 | 980.117 | 4.3% faster |

LGTM

PR #668, opened 2026-08-04
More real world benchmarks

cushon:
This benchmark is similar to a real world case where SafeRE is slower than RE2-FFM.

I'm seeing SafeRE performance around ~2x of RE2 on some real world workloads, the performance gap seems to be largely from longer >10k character inputs where the JNI/FFM overhead is negligible and RE2 has faster scanning performance.

For this specific benchmark case I measured SafeRE as ~2.2x slower, 49ns/op vs. 22ns/op for RE2-FFM.

```
./run-java-benchmarks.sh --fastbuild CrossEngineBenchmark.run -- -p crossEngineTrial=RealWorldRegexBenchmark.runBenchmark.recitation.noMatch.10000@safere-string,RealWorldRegexBenchmark.runBenchmark.recitation.noMatch.10000@re2-ffm-string-conversion
```

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -1674,6 +1674,42 @@
       "shared": true
     },
     {
+      "id": "realWorldRegex.recitation.match.{size}",
+      "axes": {
+        "size": [
+          1000,
+          10000,
+          100000
+        ]
+      },
+      "recipe": {
+        "kind": "surroundToLength",
+        "prefix": "[",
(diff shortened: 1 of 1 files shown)


cushon:
Actually it looks like this specific benchmark is fixable: https://github.com/eaftan/safere/pull/669

eaftan:
Thanks @cushon!  It looks like this was included as part of #669, which I've approved and set to auto-merge.  So I'm going to close this PR.

PR #669, opened 2026-08-04
Optimize DFA fast-path by bypassing position-dependent checks for anchorless patterns

cushon:
Previously, the DFA search loop fell back to the slower general loop for the last character of the input (due to posDepThreshold boundaries), and the loop boundary calculations were too complex, preventing JIT from optimizing the fast-path loop.

This CL introduces a check for `hasPositionDependentTransitions` (determined at DFA compile time). For patterns without position-dependent transitions (no anchors, no grapheme semantics):
- The fast-path inner loop limit is set to the full text length.
- Position-dependent boundary checks are bypassed.
- This allows JIT to prune unused branches and optimize the search loop into a single tight loop that runs for the entire string.

This achieves a ~2x speedup on the recitation benchmark from https://github.com/eaftan/safere/pull/668 (from ~50us to ~25us), making SafeRE competitive with (and slightly faster than) FFM RE2 with string conversion.

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -1763,9 +1763,10 @@
         ]
       },
       "recipe": {
-        "kind": "suffixToLength",
-        "prefixUnit": "ordinary prose without a recitation ",
-        "suffix": "[1.2.3,4.5.6]",
+        "kind": "surroundToLength",
+        "prefix": "[",
+        "unit": "1.2.3,",
+        "suffix": "4.5.6]",
         "length": "{size}"
       },
       "shared": true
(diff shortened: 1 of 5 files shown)


eaftan (approved):
LGTM overall. The DFA optimization is well-scoped, and I found no correctness or linear-time concerns.

The PR now includes:
- Updated benchmark-plan and input-count invariants for the new recitation workloads.
- Declarative exclusions for the four large recitation trials that cause `StackOverflowError` in the JDK engine: matching and non-matching inputs at 10k and 100k.
- Regression coverage verifying those exclusions. The 1k JDK trials remain enabled.

The PR description reports an approximately 2× SafeRE improvement, from ~50 µs/op to ~25 µs/op. My validation of the current 10k no-match workload measured:

| Engine | Result |
|---|---:|
| SafeRE | 16.30 µs/op |
| RE2-FFM with string conversion | 19.46 µs/op |

SafeRE was approximately 1.19× faster than RE2-FFM, consistent with the PR’s relative-performance claim. I didn't rerun the pre-optimization baseline on the final workload, so I didn't independently verify the reported 2× before/after ratio, but the relative result compared to RE2-FFM is enough evidence for me.

PR #675, opened 2026-08-06
Optimize replaceAll and replaceFirst execution paths in Matcher

cushon:
Improves performance of replacement loops by introducing dedicated fast paths and lifting wrapper restrictions.

* Add replacement fast path for literal patterns
* Generalized DFA fast path to handle zero match characters, and defend against stalls on empty matches, similar to non-replacement paths
* avoid redundant second NFA execution pass when replacement templates require capturing groups by setting eager capture demand early

```
$ ./run-java-benchmarks.sh --fastbuild CrossEngineBenchmark.run -- -p crossEngineTrial=ReplaceBenchmark.literalReplaceAll@safere-string,ReplaceBenchmark.emptyReplaceAll@safere-string
```

| Benchmark | Baseline (ns/op) | Experiment (ns/op) | Change |
| :--- | :--- | :--- | :--- | 
| `literalReplaceAll` | $289.639 \pm 10.256$ | $189.717 \pm 17.709$ | **34.5% Faster** |
| `emptyReplaceAll` | $113.108 \pm 30.723$ | $95.676 \pm 0.709$ | **15.4% Faster** |

### DIFF
--- a/safere-fuzz/src/test/java/org/safere/fuzz/FuzzSupport.java
+++ b/safere-fuzz/src/test/java/org/safere/fuzz/FuzzSupport.java
@@ -418,6 +418,33 @@ final class FuzzSupport {
           () -> runJdkOracle("replaceFirst", null, () -> jdkMatcher.replaceFirst(replacement)));
     }
 
+    boolean hasReplacementMatchState() {
+      boolean safeRe;
+      try {
+        safeReMatcher.start();
+        safeRe = true;
+      } catch (IllegalStateException e) {
+        safeRe = false;
+      }
+      boolean jdk =
+          runJdkOracle(
+              "replacement match state",
+              safeRe,
+              () -> {
+                try {
+                  jdkMatcher.start();
+                  return true;
+                } catch (IllegalStateException e) {
+                  return false;
+                }
+              });
+      assertSame("replacement match state", safeRe, jdk);
+      if (safeRe) {
+        assertMatchState("replacement match state");
+      }
+      return safeRe;
+    }
+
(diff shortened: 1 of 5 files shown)


cushon:
`replaceAll` is the API where I'm still seeing RE2-FFM perform best relative to SafeRE, I'm not sure this definitely closes the gap I was seeing. I just went looking for more opportunities to optimize the replacement methods, and I think the fundamentals of these optimizations make sense.

eaftan (approved):
I found four matcher-state/compatibility regressions in the new replacement fast paths: literal `replaceFirst` lost inner captures, unmatched literal replacements validated malformed templates too early, terminal empty DFA matches could be returned again by `find()`, and diagnostics omitted the terminal empty match. I fixed these and pushed them to this PR in e747dd737b73a5de0ef462a9bb7717ea40ace99c.

I reproduced and exceeded the claimed `literalReplaceAll` improvement: 185.800 to 106.129 ns/op, or 42.9% faster. However, `emptyReplaceAll` was effectively unchanged: 60.167 ns/op on main, 59.737 ns/op on the PR before fixes, and 60.211 ns/op after the fixes. The fixes therefore do not explain the mismatch.

The `emptyReplaceAll` workload uses `a*` on `abc`, which already takes `nullableCharClassReplaceFastPath` on main, so it does not exercise this PR’s generalized DFA path. I don’t consider that benchmark mismatch blocking given the substantial, reproducible `literalReplaceAll` improvement, and I’m happy to approve the PR with the fixes now included.

PR #687, opened 2026-08-07
Generalize fixed-offset start acceleration

cushon:
Fixed-offset start acceleration previously required the prefix before an invariant literal to have a single, constant width (e.g. `\d{3}/\d{3}/\d{4}` where `/` is at exact offset 3). Any pattern containing leading alternations (such as line-start anchors `(^|[^#])(#!tag)`), optional atoms (`(?:\$|#)?literal`), or bounded repetitions was rejected by `exactAsciiWidth`, disabling literal start acceleration and falling back to byte-by-byte DFA scanning.

This change generalizes fixed-offset start acceleration to support:
1. Discrete multi-offsets: A bounded set of exact integer offsets (e.g., `{0, 1}` for `(^|[^#])` or `{2, 4}` for `(?:--|/\*)`).
2. Bounded ranges: Small repetition windows (e.g., `\s{0,8}directive`) for multi-character literals (`length >= 2`).
3. Multi-offset candidate pre-filtering: For discrete offsets, candidate start positions are checked against the leading character class (`firstAscii`) inside the literal scanning loop, skipping non-matching occurrences in SIMD time without invoking the forward DFA.

```
./run-java-benchmarks.sh --fastbuild CrossEngineScalingBenchmark.run -- \
  -p crossEngineScalingTrial=Issue488ReplaceAllBenchmark.altCapture.1000@safere-string,Issue488ReplaceAllBenchmark.altCapture.10000@safere-string,Issue488ReplaceAllBenchmark.altCapture.100000@safere-string
```

| Input Size | Base | Optimized | Delta | Speedup |
| **1,000** | `3.946 ± 0.154 µs/op` | **`1.692 ± 0.156 µs/op`** | **-57.1%** | **2.33x faster** |
| **10,000** | `33.591 ± 1.117 µs/op` | **`11.518 ± 0.225 µs/op`** | **-65.7%** | **2.92x faster** |
| **100,000** | `335.942 ± 11.443 µs/op` | **`112.518 ± 10.271 µs/op`** | **-66.5%** | **2.99x faster** |

### DIFF
--- a/safere-fuzz/src/test/java/org/safere/fuzz/Utf8InputFuzzer.java
+++ b/safere-fuzz/src/test/java/org/safere/fuzz/Utf8InputFuzzer.java
@@ -27,6 +27,8 @@ final class Utf8InputFuzzer {
     assertLiteralSearchMatchesString("XXXXXX", "..XXXXXX");
     assertBoundarySensitiveRegionCaptures();
     assertKeywordAlternationMatchesString(data);
+    assertFixedOffsetAccelerationMatchesString(data);
+    assertMultiOffsetLiteralOccurrencesMatchString(data);
     String repeatedLiteral =
         String.valueOf((char) data.consumeInt('A', 'Z')).repeat(data.consumeInt(2, 32));
     String suffix = new String(data.consumeBytes(data.consumeInt(0, 64)), StandardCharsets.UTF_8);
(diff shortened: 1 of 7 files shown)


eaftan (approved):
I noticed that the follow-up ASCII-only restriction in 44193f3c130b084e082ead7b73f141b8b273d86c fixed the unsafe UTF-8 byte-offset case, but it also disabled the optimization for this PR's `(^|[^#])...` benchmark. I also found 2 correctness issues in generalized candidate handling, along with recursive width analysis and a quadratic metadata scan.

I've pushed commits that fix these issues with conservative code-point retreat, iterative width analysis, and a linear maximal-literal scan. 

Benchmark results with the above fixes:

| Input size | Baseline (us/op) | PR + fix (us/op) | Result |
|---:|---:|---:|---|
| 1,000 | 2.833 | 1.262 | 2.24x faster |
| 10,000 | 25.664 | 10.360 | 2.48x faster |
| 100,000 | 255.385 | 95.090 | 2.69x faster |

LGTM, will merge

PR #656, opened 2026-08-01
Prototype Vector API String scanning

cushon:
This is another prototype of using the Vector API to accelerate scanning of the String APIs (see #504, #503, #654).

The main challenge with supporting the String APIs compared to the UTF-8 APIs is that the backing String array is not part of a public API. #503 used reflection to access String internals, but relying on unsupported String internals isn't appealing for production workloads.

This PR implements a few different alternatives for evaluation:

* chunked scanning without relying on String internals - this uses the standard `String.getChars()` API to copy 512 char chunks into a reused thread-local `char[]` buffer, and then do vectorized scans of the buffer.
* unsafe short vector scans - this uses reflection to access the String internal `byte[] value`, but instead of leaking the compact string implementation details it always creates a `ShortVector`, using `VectorOperators.B2S` on latin1 strings.
* unsafe byte vector hybrid - this also uses String internals, and uses the String's coder to return a `ByteVector` for latin1 strings and `ShortVector` for UTF-16 strings

Benchmark results

```
TRIALS="CharClassStartAccelerationBenchmark.lengthSweep.128@safere-string,CharClassStartAccelerationBenchmark.lengthSweep.512@safere-string,CharClassStartAccelerationBenchmark.lengthSweep.2048@safere-string"
```

e.g.

```
./run-java-benchmarks.sh --fastbuild CrossEngineBenchmark.run -- -jvmArgsAppend "-Dsafere.vector.disabled=true" -p crossEngineTrial=$TRIALS
```

| Input Length | Scalar Baseline | Copy Mode | Unsafe - Short | Unsafe-Byte - Hybrid | Speedup (Copy vs Scalar) | Speedup (Hybrid vs Scalar) | Speedup (Hybrid vs Unsafe) |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| **128** | 259.43 | 267.34 (scalar) | 187.66 (vector) | 174.36 (vector) | **1.00x** (bypassed) | **1.48x** | **1.07x** |
| **512** | 799.40 | 476.54 (vector) | 259.96 (vector) | 217.41 (vector) | **1.67x** | **3.67x** | **1.19x** |
| **2048** | 2788.29 | 1437.01 (vector) | 540.02 (vector) | 422.47 (vector) | **1.94x** | **6.60x** | **1.27x** |

Observations

[...]

PR #688, opened 2026-08-09
Add unit tests for literal and character-class prefix start acceleration

cushon:
Add PrefixAccelerationTest covering case-insensitive, case-sensitive, and character-class prefix start acceleration across String and UTF-8 inputs with boundary alignments.

This was created as part of #656, but the test coverage could be useful independent of the String vector optimization work.


eaftan (approved):
I noticed that most of the new tests used fully literal patterns, so they exited through the literal fast path without exercising prefix acceleration. I've pushed a commit that retains each literal prefix while adding a nonliteral suffix, ensuring the intended path is covered. 

LGTM with that change

PR #689, opened 2026-08-09
Consolidate singleCharClass and add ASCII fast-path for single character class matching

cushon:
This was developed as part of #656, but is independent from it.

Currently this is performance neutral, the branch for ASCII seems to have been getting predicted and the separate loop doesn't help. The `indexOfCharClass` that was extracted could be more helpful in the future if it's vectorized, so this is mostly a refactoring/cleanup that paves the way for that.


eaftan (approved):
LGTM

PR #690, opened 2026-08-09
Extract shared 8-bit ByteVectorScan and ByteSwarScan kernels from Utf8InputScanner

cushon:
This is a pure refactoring, the point is that this code can be also be for latin1 in vector and SWAR fast paths for `String` that have access to its backing storage, if the `String` is latin1 (#656).


eaftan (approved):
This is a good refactoring.  I did notice that performance regressed by on one specific benchmark, `Utf8MatchingBenchmark.requiredSingleAsciiNonmatchBytes@safere-utf8`, by about 12%.  My agent investigated and found that one part of the refactoring increased register pressure.

I've pushed a commit that fixes the regression.  With that fix, this refactoring is performance neutral.

LGTM

Issue #674, opened 2026-08-06
Prepare for Vector API becoming a preview feature / reconsider MR-JARs


cushon:
After https://github.com/eaftan/safere/commit/b75da43d5296245b009990f9f751f643d3072113 the Vector API is unconditionally compiled into the main SafeRE module, and only loaded if the Vector API preview is enabled.

I think this approach may need to be revised when the Vector API graduates from an incubating feature to a preview feature.

Preview features require compiling with `--enable-preview`, which produces class files with a special minor version number that require also passing `--enable-preview` at runtime to use.

The Vector scanner is in the main SafeRE module, and for modularized compilations all of the sources have to be passed to a single javac invocation, and `--enable-preview` shouldn't be passed to the other SafeRE classes that don't use the Vector prototype.

I may be missing ways to reconcile that and keep the Vector stuff in the main module even after it becomes a preview feature, but I wonder if it'll be necessary to move it back to an MR-JAR eventually.

Nothing needs to happen until Vector becomes a preview, which may happen relatively soon now that JEP 401 is in preview.

cushon:
Also it does look like there are some changes in the Vector API since JDK 21, e.g. some `VectorOperators`s got renamed. I ran into that experimenting with a change to the SafeRE Vector scanner, MR-JARs would provide a way to deal with changes like that.

https://github.com/openjdk/jdk/blob/eabfcf47f63542efcca4cfe7375345b09b257589/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/VectorOperators.java#L655

https://github.com/openjdk/jdk21u-dev/blob/3139a68dd765f049596b5e83bd6d0d3882ffe193/src/jdk.incubator.vector/share/classes/jdk/incubator/vector/VectorOperators.java#L643

eaftan:
Just a note that Vector is still in incubation in Java 27 and 28-ea.  I'd be happy to add MR-JAR support back in if we need it for any reason though.

PR #692, opened 2026-08-10
Add 16-bit Vector and SWAR kernels

cushon:
Following the extraction of `ByteVectorScan` and `ByteSwarScan` in #690, this adds support for 16-bit (UTF-16) and `MemorySegment` SIMD and SWAR scanning kernels.

This adds support for all combinations of:

* 1-byte (Latin-1, UTF-8) and 2-byte (UTF-16) charsets
* Heap Arrays (`byte[]`, `char[]`) and `MemorySegment`
* Vector API (`ByteVector`, `ShortVector`) and 64-bit broadword SWAR

This lays groundwork for optimized `String` APIs using vector and SWAR scanning across heap arrays or `MemorySegments` representing the backing data for a Java `String` (#656).

The FFM `MemorySegment` APIs aren't stable until JDK 22+, so this adds a separate `safere-foreign` module for `MemorySegment` based APIs. It would also be possible to include them in the main module and load them reflectively on JDK 22+, but that could cause problems for tools that scan the jar and expect the bytecode to all be JDK 21 compatible. Another option would be to re-introduce multi-release jars. Having a separate module seems OK for now, but I think the other approaches are probably workable if benefits to them are found.


eaftan (commented):
I'm happy accepting this kernel groundwork given my belief in the direction. I did find several correctness issues in the initial implementation: accepted vector ranges could be skipped, unsigned and supplementary UTF-16 ranges were mishandled, byte-backed UTF-16 was endian-dependent, large segment positions could be narrowed, invalid byte ranges and empty prefixes were not handled safely, and multi-character ignore-case searches could become O(n*m). The new module was also included unconditionally in the JDK-21 reactor.

I've pushed a commit that fixes those issues. Longer ignore-case prefixes now use linear-time KMP while one-character scans retain the SIMD/SWAR path, and `safere-foreign` is activated only on JDK 22+. 

My main feedback is about packaging. As you suggested in #674, I think we should probably revive SafeRE's MR-JAR approach rather than establish `safere-foreign` as a separate public artifact. I removed the earlier MR-JAR infrastructure because the UTF-8 Vector implementation turned out to work entirely on JDK 21, which is already SafeRE's minimum supported version. But in this PR we have genuinely version-specific functionality, and we're likely to want more going forward.  Given all that, MR-JAR seems like the most elegant way to package this.

cushon:
Thanks for the fixes!

> I think we should probably revive SafeRE's MR-JAR approach rather than establish safere-foreign as a separate public artifact.

Switching back to MR-JARs SGTM. What do you think is the best sequencing? Do you want to restore the MR-JAR build integration from https://github.com/eaftan/safere/commit/b75da43d5296245b009990f9f751f643d3072113 first and then update this PR to use it, or do that as a follow-up?

Also I have slightly mixed feelings on MR-JARs--I think they still aren't that well supported by parts of the ecosystem, but they also seem like the cleanest way to handle using FFM and Vector when they're available, and supporting Vector as it graduates from incubator to preview, and potentially benefiting from upcoming features (like using value types to flatten types in the DFA). But I think they're worth trying again.

---

Also in `ShortVectorScan` and `SegmentShortVectorScan` I think the `lanewise(XOR, ...` could still be avoided for the lower half of BMP:

```java
  if (highBound <= 0x7FFF) {
    return values.compare(VectorOperators.GE, low).and(values.compare(VectorOperators.LE, high));
  }
```

eaftan:
> Switching back to MR-JARs SGTM. What do you think is the best sequencing? Do you want to restore the MR-JAR build integration from https://github.com/eaftan/safere/commit/b75da43d5296245b009990f9f751f643d3072113 first and then update this PR to use it, or do that as a follow-up?

Yeah, that sounds good.  I'll probably put up a stacked PR on top of this one to revive the MR-JAR stuff and send it to you for review.  Then when I merge it, it will merge into this PR, then I can approve this PR.

> Also in `ShortVectorScan` and `SegmentShortVectorScan` I think the `lanewise(XOR, ...` could still be avoided for the lower half of BMP:
>
> ```java
>   if (highBound <= 0x7FFF) {
>     return values.compare(VectorOperators.GE, low).and(values.compare(VectorOperators.LE, high));
>   }
> ```

Good point.  Done in 4c3dfe5f10cbd97fba54d3678c6bbcadc7219257.

cushon:
> I'll probably put up a stacked PR on top of this one to revive the MR-JAR stuff and send it to you for review. Then when I merge it, it will merge into this PR, then I can approve this PR.

The changes in https://github.com/cushon/safere/pull/1 LGTM, thanks!

cushon:
I have reworked this again to drop the `MemorySegment` kernels of now, and keep the new `short` Vector and SWAR Kernels. It isn't clear the `MemorySegment` support is going to be useful so I think it's best to defer it.

That also means there isn't a need to switch back to MR-JARs immediately, sorry for the back and forth on that. I kept the `DEVELOPMENT.md` with the notes about MR-JARs, which should be useful when the Vector API graduates from incubator to preview and needs to be moved to a MR-JAR.

safere/src/main/java/org/safere/internal/Ascii.java
```diff
@@ -0,0 +1,53 @@
+// This file is part of a Java port of RE2 (https://github.com/google/re2).
+// Original RE2 code is Copyright (c) 2009 The RE2 Authors.
+// Modifications and Java port Copyright (c) 2026 Eddie Aftandilian.
+// Licensed under the BSD 3-Clause License (see LICENSE file).
+
+package org.safere.internal;
```
eaftan: Would it make sense to move the 3 classes in `org.safere.internal` to `org.safere` and make them package-private?  Since they're public, consumers can depend on them when the JAR is used on the traditional classpath rather than the module path.  Making them package-private would keep them implementation details.

eaftan (dismissed):
Makes sense to me.  Let me know if you do need any newer Vector APIs and I can revive the MR-JAR packaging.  

My comment is non-blocking. Feel free to address or not and then merge.

safere/src/main/java/org/safere/internal/Ascii.java
cushon: Good point, thanks. This was only in a separate package to share with the vector submodule, which doesn't exist anymore. I moved them to `org.safere` and made them package-private.

eaftan (approved):

PR #693, opened 2026-08-10
Refactor start accelerators behind a shared interface

cushon:
This is a refactoring to extract a shared abstraction for start accelerators (literal, line anchor, char class), in preparation for the changes in #655.

### DIFF
--- a/safere-crosscheck/pom.xml
+++ b/safere-crosscheck/pom.xml
@@ -128,6 +128,7 @@
                         <exclude name="SafeReDiagnosticsTest.java"/>
                         <exclude name="SearchScalingRegressionTest.java"/>
                         <exclude name="SimplifierTest.java"/>
+                        <exclude name="StartAcceleratorTest.java"/>
                         <exclude name="UnicodeJdkConsistencyTest.java"/>
                         <exclude name="UnicodeTablesTest.java"/>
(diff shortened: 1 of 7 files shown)


eaftan (approved):
This is a good maintainability refactoring: the typed hierarchy is clearer, non-discrete offsets are preserved, and start acceleration correctly avoids bypassing reject prefilters.

I did notice that the shared interface added measurable dispatch overhead on two UTF-8 hot paths. I also found that literal-prefix patterns encoded and retained the same UTF-8 prefix twice. I've pushed commits that preserve direct sealed-type dispatch at the hot `Pattern.find` site and reuse the existing encoded prefix.

| Benchmark | main (ns/op) | submitted head (ns/op) | with fixes (ns/op) | Result |
|---|---:|---:|---:|---|
| `FixedOffsetAccelerationBenchmark.findDigitsWithoutSlash@safere-utf8` | 1537.464 | 1841.309 | 1533.446 | neutral |
| `Utf8MatchingBenchmark.literalPrefixNonmatchBytes@safere-utf8` | 1363.911 | 1573.770 | 1383.525 | neutral |

The PR head was 19.8% and 15.4% slower on those workloads; with the fix it's neutral.

LGTM

Issue #594, opened 2026-07-25
Investigate RE#


eaftan:
RE# is a non-backtracking regex engine for .NET. Instead of the NFA/DFA approach that RE2-descendants use, RE# takes a derivative based approach. They claim SOTA performance on the rebar benchmark suite. 

We should investigate RE# and see there are any learnings we can integrate into SafeRE. Perhaps we should implement an engine that uses this approach and either expose it as an optional mode, or select it when it outperforms our current strategies. 

Paper: https://arxiv.org/pdf/2407.20479

cushon:
It looks like really interesting work. There’s also a rust port, which make some refinements to the approach that could also be worth considering: https://iev.ee/blog/symbolic-derivatives-and-the-rust-rewrite-of-resharp/

eaftan:
After looking more closely at the paper and its baseline results, I do not think the performance is intrinsic to derivatives. Derivatives give RE# a convenient algebraic representation for discovering optimizations, but the resulting matcher is still essentially a symbolic lazy DFA. Most of the useful ideas should transfer to SafeRE without adding a derivative engine or changing SafeRE semantics.

The main transferable idea is: **treat every automaton state as a new search problem, and attach the best available accelerator to that state.** SafeRE already performs substantial analysis and acceleration for the original pattern. RE# continues doing this after the match has partially progressed.

A possible implementation order:

1. **Per-DFA-state self-loop skipping.** For a DFA state `q`, identify character classes `C` for which `delta(q, C) == q` and consuming `C` cannot report or otherwise affect a match. Instead of taking one DFA transition per character, scan directly for the next character outside `C`. For example, after consuming `abcd` in `abcd.*efg`, the residual state can skip directly to candidate `e` characters. This can be derived from the ordinary DFA graph; residual regex objects are not required. A conservative first version could cover ASCII input, nonaccepting states, and patterns without position-dependent assertions.

2. **State-local literal runs.** When the continuation from a state has an unambiguous literal path, compare several characters at once and jump directly to the destination state. When the current state self-loops over all noncandidate characters, this can become a search-and-consume operation rather than merely a bulk comparison. This generalizes prefix acceleration from the beginning of a pattern to intermediate DFA states.

[...]

PR #694, opened 2026-08-10
Support disjoint required literal reject prefilters for alternations

cushon:
When executing unanchored searches for alternations where every branch requires a distinct literal substring (such as `.*(?:apple|banana|cherry).*`, `(foo.*|bar.*|baz.*)`, or `(?:\bfirstToken\b|\bsecondToken\b)`), this introduces a reject prefilter using `indexOf`.

The optimization only applies to a maximum of 16 alternations, at that threshold repeated calls to `indexOf` should still be faster than the DFA. There is potential future work to vectorize this approach and do it one pass, and get more benefit (see also #656 #594).

This was inspired by similar concepts in Hyperscan's RejectPrefilter / ReSharp's Noodle branch-literal extraction / Rust regex-automata's prefilter gate.

### DIFF
--- a/safere/src/main/java/org/safere/Matcher.java
+++ b/safere/src/main/java/org/safere/Matcher.java
@@ -438,6 +438,7 @@ public final class Matcher implements MatchResult {
 
   private void invalidatePatternCaches() {
     preparedMatchRunner = null;
+    disjointRequiredLiteralsChecked = false;
     cachedForwardFirstMatchDfa = null;
     cachedForwardLongestMatchDfa = null;
     cachedReverseDfa = null;
--- a/safere/src/main/java/org/safere/Matcher.java
+++ b/safere/src/main/java/org/safere/Matcher.java
@@ -21,6 +21,7 @@ import java.util.function.Function;
 import java.util.regex.MatchResult;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
+import org.safere.Pattern.DisjointRequiredLiterals;
 
 /**
  * An engine that performs match operations on a {@linkplain CharSequence character sequence} by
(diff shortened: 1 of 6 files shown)


eaftan (dismissed):
This one looks reasonable.  I understand why you'd take this `indexOf` approach as opposed to doing it in one pass given the situation with accessing String internals.

I think the central decision for this version of the optimization is how many required literals should be accepted into the repeated-`String.indexOf` path. To measure the crossover, I benchmarked `RegexBenchmark.alternationFind` on paired current-main and PR builds using a 32 KiB rejected String input with two, four, and eight disjoint required literals. Performance below is normalized to main, so higher is better:

| Required literals | PR performance relative to main | Decision |
|---:|---:|---|
| 2 | 2.50x | Keep the prefilter |
| 4 | 1.27x | Keep the prefilter |
| 8 | 0.62x | Use the existing matcher path |

This PR's original threshold was 16. Since the optimization is already slower than main at eight literals, **I went ahead and changed the threshold from 16 to four, the largest measured winning count.** (bolding so you notice this)

I also found several correctness and scaling boundaries in the submitted implementation. UTF-16 substring minimization could incorrectly prune a supplementary-code-point branch, repeated `find()` calls could become quadratic, `usePattern()` did not reset the cached state, start-anchored patterns could scan the whole input unnecessarily, and the UTF-8 path was substantially slower than the existing DFA.

I've pushed commits that fix those issues and add regression coverage. The prefilter now uses code-point KMP for minimization, runs once per String matcher search session, resets with matcher state, skips anchored and UTF-8 paths, and is capped at four literals based on the measured crossover. 

As a secondary worst-case positive-input check, I benchmarked `(?:banana|apple)\d` against a 32 KiB input consisting of `apple0 ` repeated and truncated to 32,768 characters, yielding 4,681 matches per find-all operation. Because the first required literal, `banana`, is absent, the prefilter scans the full input before finding `apple`; normal matching then starts from the beginning and enumerates every match. This deliberately unfavorable case is about 6% slower than current main. 

I think the 2.50x two-literal and 1.27x four-literal rejection gains justify that small positive-input cost.

cushon:
Thanks!

> I went ahead and changed the threshold from 16 to four, the largest measured winning count

SGTM--it might make sense to have the threshold be tuneable depending on which accelerator it's being used with, e.g. I think if this can be vectorized it won't require 16 passes for 16 literals. But that can be future work.

eaftan (approved):

PR #695, opened 2026-08-10
Refactor to encapsulate reject prefilters

cushon:
This is a similar refactoring to https://github.com/eaftan/safere/pull/693 but for reject prefilters. This will make adding additional strategies like https://github.com/eaftan/safere/pull/694 simpler, and lays groundwork for future optimizations like Teddy SIMD or aho-corasick prefilters.

### DIFF
--- a/safere-crosscheck/pom.xml
+++ b/safere-crosscheck/pom.xml
@@ -125,6 +125,7 @@
                         <exclude name="ProgTest.java"/>
                         <exclude name="RegexpOpTest.java"/>
                         <exclude name="RegexpTest.java"/>
+                        <exclude name="RejectPrefilterTest.java"/>
                         <exclude name="SafeReDiagnosticsTest.java"/>
                         <exclude name="SearchScalingRegressionTest.java"/>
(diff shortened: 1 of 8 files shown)


eaftan (dismissed):
This is a good maintainability refactoring. Encapsulating the reject-prefilter strategies gives them a clearer structure, and the disjoint-literal prefilter retains its once-per-search and String-only semantics from #694.

I did notice that composite String diagnostics could report the first filter instead of the filter that actually rejected. I also found measurable interface-dispatch overhead on the UTF-8 leaf path. I've pushed commits that fix the diagnostics attribution and preserve direct sealed-type dispatch at the hot site.

| Benchmark | main (ns/op) | submitted head (ns/op) | with fixes (ns/op) | Result |
|---|---:|---:|---:|---|
| `Utf8MatchingBenchmark.requiredSingleAsciiNonmatchBytes@safere-utf8` | 1404.232 | 1590.839 | 1377.088 | neutral |

The submitted implementation was 13.3% slower on that workload; with the fix it's neutral. 

LGTM

eaftan (approved):

PR #702, opened 2026-08-12
Tune anchored OnePass limits based on capture requirements

cushon:
Differentiate OpenPass limits (256 without captures, 65536 with captures) in Matcher.

I was curious about the current OnePass thresholds, and the benchmark below seems to show strong evidence that having a lower limit for no captures and a higher limit for captures is valuable.

I think that re2 does something similar and it was missed for SafeRE, which currently always uses a fixed threshold of 4096:

https://github.com/google/re2/blob/972a15cedd008d846f1a39b2e88ce48d7f166cbd/re2/re2.cc#L838-L839

---

For patterns without captures, Forward DFA alone is faster for texts >= 512B due to its direct transition table loop. For patterns with captures or replacements with group substitutions, OnePass is 2x-28x faster up to 64KB, avoiding the multi-pass DFA + BitState/NFA submatch extraction cliff.

This change adds a `OnePassScalingBenchmark` to evaluate different thresholds.

* Patterns with Captures / Group Substitutions (`FEW_CAPTURES`):
   * OnePass is 2.2x to 28.8x faster across all input sizes up to 64 KB.
   * At 16 KB and 64 KB, DFA submatch extraction exceeds BitState limits and falls back to NFA, jumping from microseconds to milliseconds (~1.5 ms and ~6.1 ms). OnePass avoids this cliff entirely (53 µs and 328 µs).
   * Conclusion: `ONEPASS_WITH_CAPTURE_LIMIT = 65536` avoids regressions on medium/large inputs with captures.

* Patterns without Captures / Literal Replacements (`NO_CAPTURES`):
   * DFA is 1.3x to 3.0x faster for inputs >= 512 B, and over 31x faster at 256 KB.
   * Because no submatch extraction is needed, Forward DFA processes inputs with an unrolled ASCII transition table loop.
   * Conclusion: `ONEPASS_NO_CAPTURE_LIMIT = 256` gives low startup latency for tiny strings while routing larger texts directly to the DFA.

---

`matches()`

| Workload | Input Size | ONE_PASS (ns/op) | DFA_SANDWICH (ns/op) | Ratio (DFA / OnePass) | Faster Engine |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **NO_CAPTURES** | 64 B | 243.3 | 299.3 | 1.23x | ONE_PASS |
| | 512 B | 1,658.5 | 1,350.1 | 0.81x | DFA |
| | 4 KB | 12,887.5 | 9,974.3 | 0.77x | DFA |
| | 16 KB | 51,420.9 | 39,840.2 | 0.77x | DFA |
| | 64 KB | 204,435.8 | 155,594.6 | 0.76x | DFA |
| | 256 KB | 20,206,852.7 | 621,147.3 | 0.03x | DFA (32.5x faster) |
| **FEW_CAPTURES** | 64 B | 279.4 | 750.9 | 2.69x | ONE_PASS |
| | 512 B | 1,707.5 | 4,507.7 | 2.64x | ONE_PASS |
| | 4 KB | 13,376.0 | 60,479.4 | 4.52x | ONE_PASS |
| | 16 KB | 53,215.3 | 1,534,371.1 | 28.83x | ONE_PASS |
| | 64 KB | 328,753.5 | 6,154,217.7 | 18.72x | ONE_PASS |
| | 256 KB | 23,681,828.6 | 24,119,969.6 | 1.02x | Parity |
| **MANY_CAPTURES** | 64 B | 999.9 | 1,023.8 | 1.02x | ONE_PASS |
| | 512 B | 4,641.4 | 4,643.5 | 1.00x | Parity |
| | 4 KB | 34,557.7 | 42,538.9 | 1.23x | ONE_PASS |
| | 16 KB | 1,487,248.0 | 1,499,404.6 | 1.01x | Parity (NFA fallback) |
| | 64 KB | 5,983,463.9 | 6,292,835.4 | 1.05x | Parity (NFA fallback) |
| | 256 KB | 23,328,192.0 | 23,854,209.4 | 1.02x | Parity (NFA fallback) |

`replaceAll()`

| Workload | Input Size | ONE_PASS (ns/op) | DFA_SANDWICH (ns/op) | Ratio (DFA / OnePass) | Faster Engine |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **NO_CAPTURES** (Literal replacement) | 64 B | 656.4 | 388.9 | 0.59x | DFA (1.69x faster) |
| | 512 B | 4,276.6 | 1,505.7 | 0.35x | DFA (2.84x faster) |
| | 4 KB | 31,372.3 | 10,318.6 | 0.33x | DFA (3.04x faster) |
| | 16 KB | 125,606.7 | 40,908.5 | 0.33x | DFA (3.07x faster) |
| | 64 KB | 481,824.8 | 170,452.6 | 0.35x | DFA (2.83x faster) |
| | 256 KB | 20,605,085.5 | 648,626.5 | 0.03x | DFA (31.77x faster) |
| **FEW_CAPTURES** (`$1://$2-mod:$3`) | 64 B | 625.6 | 1,404.7 | 2.25x | ONE_PASS |
| | 512 B | 2,376.9 | 5,987.3 | 2.52x | ONE_PASS |
| | 4 KB | 16,912.7 | 62,580.3 | 3.70x | ONE_PASS |
| | 16 KB | 65,471.1 | 1,548,078.0 | 23.65x | ONE_PASS |
| | 64 KB | 269,914.5 | 6,104,948.9 | 22.62x | ONE_PASS |
| | 256 KB | 23,882,177.5 | 24,629,401.7 | 1.03x | Parity |
| **MANY_CAPTURES** (8 group references) | 64 B | 1,617.4 | 1,894.2 | 1.17x | ONE_PASS |
| | 512 B | 5,488.7 | 6,391.6 | 1.16x | ONE_PASS |
| | 4 KB | 36,189.8 | 44,843.5 | 1.24x | ONE_PASS |
| | 16 KB | 1,481,257.5 | 1,511,535.1 | 1.02x | Parity (NFA fallback) |
| | 64 KB | 5,960,150.2 | 6,328,519.2 | 1.06x | Parity (NFA fallback) |
| | 256 KB | 26,708,586.4 | 25,208,788.3 | 0.94x | Parity (NFA fallback) |


safere/src/main/java/org/safere/Matcher.java
```diff
@@ -4465,9 +4477,11 @@ private PreparedMatchRunner createPreparedRunner(boolean regionActive) {
 
     // Anchored OnePass runner
     if (options.onePass()
-        && (parentPattern.canOnePassFind() || parentPattern.canOnePassPrimary())
-        && activeScanner().length() <= ONEPASS_ANCHORED_TEXT_LIMIT) {
-      return new OnePassAnchoredPreparedRunner(prog.numCaptures());
+        && (parentPattern.canOnePassFind() || parentPattern.canOnePassPrimary())) {
+      boolean requiresCaptures = parentPattern.prog().numCaptures() > 0;
```
eaftan: I believe that `Prog.numCaptures()` includes implicit group zero. If so, this condition will be true for every pattern, so every prepared matcher will select the 65,536-unit limit and the new 256-unit tier won't be reached.

I think the choice may need to depend on whether this matcher has actually needed an inner group, rather than whether the pattern declares one: one matcher might only call `matches()`, while another matcher using the same `Pattern` might later call `group(1)`. My expectation would be for that demand to survive `reset()`, reset on `usePattern()`, and cause the cached prepared runner to be reconsidered when it changes. What do you think?

safere/src/main/java/org/safere/Matcher.java
```diff
@@ -2652,10 +2663,11 @@ private String replaceAnchoredOnePass(LazyTemplate template, boolean replaceAll)
         parentPattern.prog().anchorStart()
             && (parentPattern.prog().anchorEnd() || parentPattern.prog().dollarAnchorEnd());
 
+    boolean requiresCaptures = template.needsCaptures();
```
eaftan, line 2695: Could we distinguish `$0` from references to inner groups here? `needsCaptures()` currently returns true for any `$`, but a `$0` replacement only needs the whole-match boundaries that the DFA already provides, so I think it should remain in the capture-free tier.

Numbered groups greater than zero and named groups would still select the inner-capture tier. We'd also want to preserve the existing timing of malformed-replacement validation rather than fully parsing templates earlier than before.

safere-benchmarks/src/main/java/org/safere/OnePassScalingBenchmark.java
```diff
@@ -0,0 +1,110 @@
...
+        input = prefix + "a".repeat(middleLen) + suffix;
+      }
+      default -> throw new IllegalArgumentException("Unknown workload: " + workload);
+    }
+
+    saferePattern =
+        switch (engine) {
+          case ONE_PASS ->
+              Pattern.compile(
+                  regex, 0, EnginePathOptions.builder().dfa(false).onePass(true).build());
```
eaftan: I may be missing another constraint here, but it looks like these options don't fully force the engine named by `Engine`. The `NO_CAPTURES` pattern can be handled by the character-class fast path before either OnePass or DFA is selected. When a nominal `ONE_PASS` input exceeds the production OnePass threshold, disabling DFA also appears to let it fall through to BitState/NFA.

Could the benchmark verify the strategy that actually ran, or instead measure the real automatic engine-selection policy directly? Without that, I'm not sure the rows identify the OnePass/DFA crossover they are intended to measure.

safere-benchmarks/src/main/java/org/safere/OnePassScalingBenchmark.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Param({"ONE_PASS", "DFA_SANDWICH"})
+  public Engine engine;
+
+  private String input;
+  private String replacement;
+  private Pattern saferePattern;
+
+  @Setup
+  public void setup() {
```
eaftan: Would you mind moving these workload definitions into `benchmark-data.json`, so they use the same checked-in workload source as the rest of the suite?

safere-benchmarks/src/main/java/org/safere/OnePassScalingBenchmark.java
```diff
@@ -0,0 +1,110 @@
...
+        input = prefix + "a".repeat(middleLen) + suffix;
+      }
+      case MANY_CAPTURES -> {
+        regex =
+            "^([0-9.]+) - - \\[([^\\]]+)\\] \"(GET|POST|PUT) ([^ ]+) ([^\"\\s]+)\" ([0-9]+)"
+                + " ([0-9]+) \"([^\"]+)\"$";
+        replacement = "$1 - [$2] $3 $4 $5 $6 $7 $8";
+        String prefix = "127.0.0.1 - - [10/Oct/2000:13:55:36 -0700] \"GET /";
+        String suffix = " HTTP/1.0\" 200 2326 \"http://example.com\"";
+        int middleLen = Math.max(1, size - prefix.length() - suffix.length());
```
eaftan: I also noticed that the fixed prefix and suffix for `MANY_CAPTURES` already exceed 64 characters. With `Math.max(1, ...)`, the size-64 row therefore appears to construct an input larger than its label. Could the replacement workload make each requested input size exact?

eaftan (changes requested):
I agree with the direction here: anchored OnePass should have a lower limit when the caller only needs group zero, and remain available longer when this matcher actually needs inner captures.

I ran into a few questions that I think we should resolve before merging. I've left inline comments with the details:

- implicit group zero makes the 256-unit tier unreachable;
- capture demand needs to be based on this matcher's observed API usage and integrated with prepared-runner caching and matcher lifecycle;
- `$0` replacement must be distinguished from inner numbered or named groups;
- the submitted benchmark neither isolates its named engines nor always constructs inputs of the reported size, and its workloads need to come from `benchmark-data.json`.

Could you please update the implementation and add tests covering:

- patterns with no declared groups;
- declared but unread inner groups;
- inner-group access followed by `reset()` and another match;
- two matchers using the same `Pattern` but having different capture demand;
- `$0`, numbered inner-group, and named-group replacements;
- prepared-runner invalidation when capture demand changes?

Did you consider other threshold values? If not, consider measuring the actual engine crossover at several input sizes around 256 and 65,536, separately for callers that only need group zero and callers that consume inner captures. Those workloads should be defined in `benchmark-data.json`, use exact input sizes, and verify that the intended engines actually ran. I'd be happy to help put together and run that comparison if useful. Alternatively, we could treat these as initial values and leave further threshold tuning to a follow-up.

safere/src/main/java/org/safere/Matcher.java
cushon: Done: updated `LazyTemplate` parsing to distinguish group 0, and updated the heuristics to use the DFA when a replacement references only `$0` or literal text.

safere/src/main/java/org/safere/Matcher.java
cushon: Done. Consolidated the threshold logic and used it consistently across `matches()`, `find()`, and `replaceAnchoredOnePass()`, and added more javadoc.

safere-benchmarks/src/main/java/org/safere/OnePassScalingBenchmark.java
cushon: Done.

* Added `OnePassThresholdTest.java` using `OperationDiagnostics` to explicitly verify that the expected strategy executes
* Added `OnePassScalingBenchmark` workloads to `benchmark-data.json` to measure the actual performance across input sizes

safere-benchmarks/src/main/java/org/safere/OnePassScalingBenchmark.java
cushon: Done

safere-benchmarks/src/main/java/org/safere/OnePassScalingBenchmark.java
cushon: Done. Added `OnePassThresholdTest.java` with tests verifying thresholds.

cushon:
Thanks for the review!

I've addressed the issues you've found.

* The heuristics now distinguish between `$0` and non-`$0` capturing groups to use the DFA for replacements that only use `$0`
* For patterns with capturing groups, the approach now uses OnePass up to 64k in `Matcher`, to avoid regressions with the pattern `pattern.matcher(input).matches() -> group(1)`. The benefit of using the DFA for short inputs was outweighed by the penalty of falling back if `group(...)` is called, and caused some benchmark regressions. APIs where we know whether or not capture groups are needed like `replaceAll` still make use of precise information about whether or not non-zero captures are needed.

The benchmarks show a win for `replaceAll` and neutral results for everything else.

```
./safere-benchmarks/scripts/compare-branch.sh   --baseline benchmark-baseline   --current onepass-threshold   'OnePassScalingBenchmark.*@safere-string'
```

| OnePassScalingBenchmark.noCaptures.64            |        200 ± 4.9 |       202 ± 5.1 |   0.99x |
| OnePassScalingBenchmark.noCaptures.256           |         701 ± 22 |        700 ± 11 |   1.00x |
| OnePassScalingBenchmark.noCaptures.512           |        1351 ± 30 |       1426 ± 59 |   0.95x |
| OnePassScalingBenchmark.noCaptures.4096          |      10575 ± 103 |      9999 ± 132 |   1.06x |
| OnePassScalingBenchmark.noCaptures.65536         |    158935 ± 1189 |   158917 ± 1303 |   1.00x |
| OnePassScalingBenchmark.withCaptures.64          |         296 ± 12 |       298 ± 7.2 |   0.99x |
| OnePassScalingBenchmark.withCaptures.256         |         899 ± 35 |        894 ± 13 |   1.01x |
| OnePassScalingBenchmark.withCaptures.512         |        1685 ± 32 |       1686 ± 40 |   1.00x |
| OnePassScalingBenchmark.withCaptures.4096        |       12744 ± 63 |     12874 ± 119 |   0.99x |
| OnePassScalingBenchmark.withCaptures.65536       |   420949 ± 59568 |   301194 ± 3121 |   1.40x |
| OnePassScalingBenchmark.withCaptures.262144      |  2057683 ± 36716 | 2045447 ± 30173 |   1.01x |
| OnePassScalingBenchmark.replaceAllLiteral.64     |         664 ± 32 |        660 ± 26 |   1.01x |
| OnePassScalingBenchmark.replaceAllLiteral.512    |        3701 ± 85 |       984 ± 715 |   3.76x |
| OnePassScalingBenchmark.replaceAllLiteral.4096   |      15909 ± 248 |     3962 ± 2124 |   4.01x |
| OnePassScalingBenchmark.replaceAllLiteral.65536  |    432613 ± 7941 |   64574 ± 41641 |   6.70x |
| OnePassScalingBenchmark.replaceAllCaptures.64    |         811 ± 56 |        797 ± 14 |   1.02x |
| OnePassScalingBenchmark.replaceAllCaptures.512   |        3812 ± 66 |      3975 ± 399 |   0.96x |
| OnePassScalingBenchmark.replaceAllCaptures.4096  |     16606 ± 1024 |     16175 ± 812 |   1.03x |
| OnePassScalingBenchmark.replaceAllCaptures.65536 |    439666 ± 7413 |  257606 ± 10565 |   1.71x |

eaftan (approved):
Thanks, this addresses my earlier questions. I agree with using the 65,536-character limit whenever a pattern declares inner groups: `matches()` cannot know whether the caller will request `group(1)` afterward, while replacement operations can choose based on the replacement template.

I found two smaller issues while checking the revision. Java parses a numeric replacement reference using the pattern's actual group count. For a pattern with no inner capture groups, for example, `$01` means `$0` (the entire match) followed by the literal character `1`, because group 1 does not exist. The new preliminary capture-demand check instead read `$01` as a reference to group 1. The final replacement output was still correct because the real replacement parser handled it properly, but the preliminary check could unnecessarily select the more expensive inner-capture tier. The new replacement workloads also needed explicit backslash-number syntax for C++ RE2 so that the cross-engine trials perform the same operation. I've pushed a commit that fixes both and adds regression coverage.

I compared the revision with current main using the same declarative workloads on both sides. Time is normalized as `PR/main`, so lower is better:

| Workload | Size | PR/main | Result |
|---|---:|---:|---|
| `noCaptures` | 4,096 | 0.891 | 10.9% lower |
| `withCaptures` | 65,536 | 0.610 | 1.64x faster |
| `replaceAllLiteral` | 512 | 0.273 | 3.67x faster |
| `replaceAllLiteral` | 4,096 | 0.214 | 4.68x faster |
| `replaceAllLiteral` | 65,536 | 0.147 | 6.79x faster |

The other measured sizes were neutral. LGTM

PR #703, opened 2026-08-12
Optimize onepass compilation time

cushon:
I was looking at some profiles where OnePass was the most expensive part of `Pattern.compile`.

This change skips `OnePass` compilation entirely for patterns that won't use it because they have lazy quantifiers to too many captures. It also optimizes some allocations during OnePass construction.

This is related to https://github.com/eaftan/safere/commit/2a668bf25de7129fce7cabe8a23419bb2f871a97 which made this initialization eager. I still think that's probably the right tradeoff, and if `Pattern.compile` is a bottleneck it's often because patterns are getting compiled over and over instead of stored in a constant field or cached.

The benchmarks show a small win, also variance is reduced perhaps because of fewer allocations. I think these benchmarks are spending less time in onepass compilation than the profile I was looking at, which was compiling a very large pattern.

```
./run-java-benchmarks.sh --fastbuild CrossEngineScalingBenchmark -- \
  -p crossEngineScalingTrial=CompileBenchmark.compileSimple@safere-string,CompileBenchmark.compileMedium@safere-string,CompileBenchmark.compileComplex@safere-string,CompileBenchmark.compileAlternation@safere-string
```

| Benchmark | Baseline (us/op) | Experiment (us/op) | Delta |
| :--- | :---: | :---: | :---: |
| `CompileBenchmark.compileSimple` | 8.338 ± 0.517 | 7.923 ± 0.240 | -5.0% |
| `CompileBenchmark.compileMedium` | 36.073 ± 1.791 | 36.033 ± 3.643 | -0.1% |
| `CompileBenchmark.compileComplex` | 20.862 ± 5.362 | 18.708 ± 1.241 | -10.3% |
| `CompileBenchmark.compileAlternation` | 37.545 ± 3.507 | 34.768 ± 1.003 | -7.4% |

### DIFF
--- a/safere/src/main/java/org/safere/OnePass.java
+++ b/safere/src/main/java/org/safere/OnePass.java
@@ -5,12 +5,7 @@
 
 package org.safere;
 
-import java.util.ArrayDeque;
 import java.util.Arrays;
-import java.util.Deque;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.TreeSet;
 
 /**
  * One-pass NFA execution engine. For patterns where the NFA is deterministic (at most one possible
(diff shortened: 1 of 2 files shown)


eaftan (approved):
This looks good. I don't see any correctness issues, and the change seems like a good one.

I ran the four compile workloads:

| Benchmark | main (us/op) | PR (us/op) | Change |
|---|---:|---:|---:|
| `CompileBenchmark.compileSimple` | 6.366 | 6.102 | 4.1% faster |
| `CompileBenchmark.compileMedium` | 26.429 | 25.217 | 4.6% faster |
| `CompileBenchmark.compileComplex` | 14.359 | 13.737 | 4.3% faster |
| `CompileBenchmark.compileAlternation` | 25.515 | 24.833 | 2.7% faster |

The exact gains are smaller than some of the original measurements, but the improvement is consistent across all four workloads.

LGTM

PR #705, opened 2026-08-13
Support UTF-8 start acceleration for ASCII case-insensitive literal prefixes

cushon:
This adds start acceleration support for ASCII case-insensitive literal prefixes in `Utf8StartAccelerator`, for parity with `StringStartAccelerator`.

I noticed this looking at things that were out of sync between the different accelerator implementations.

This PR also adds an `Ascii` helper to deduplicate some helper methods, which is also present in another PR. I can deal with the merge conflicts depending on which one goes in first, I just didn't want to add more copies of those helpers here.

This also adds some test coverage to try to keep the String and UTF-8 start accelerators in sync.

### DIFF
--- a/safere/src/main/java/org/safere/internal/Ascii.java
+++ b/safere/src/main/java/org/safere/internal/Ascii.java
@@ -35,65 +35,56 @@ public final class Ascii {
     return a == b || toLowerCase(a) == toLowerCase(b);
   }
 
-  /** Builds the KMP failure function for an ASCII case-insensitive pattern. */
-  public static int[] ignoreCaseFailure(String pattern) {
-    int[] failure = new int[pattern.length()];
-    int matched = 0;
-    for (int i = 1; i < pattern.length(); i++) {
-      while (matched > 0 && !equalsIgnoreCase(pattern.charAt(i), pattern.charAt(matched))) {
-        matched = failure[matched - 1];
-      }
-      if (equalsIgnoreCase(pattern.charAt(i), pattern.charAt(matched))) {
-        matched++;
-      }
-      failure[i] = matched;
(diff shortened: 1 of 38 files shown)


safere/src/main/java/org/safere/internal/Ascii.java
```diff
@@ -0,0 +1,91 @@
+// This file is part of a Java port of RE2 (https://github.com/google/re2).
+// Original RE2 code is Copyright (c) 2009 The RE2 Authors.
+// Modifications and Java port Copyright (c) 2026 Eddie Aftandilian.
+// Licensed under the BSD 3-Clause License (see LICENSE file).
+
+package org.safere.internal;
```
eaftan: As I mentioned in [a comment on #692](https://github.com/eaftan/safere/pull/692#discussion_r3781209150), consider moving this class to `org.safere` and making it package-private.

eaftan (dismissed):
Nice catch on spotting this inconsistency!  

I measured this new UTF-8 path using `(?i)keyword_to_find.*` on the existing 100 KiB benchmark inputs and saw a nice speedup:

| Workload | main (ns/op) | PR + fix (ns/op) | Change |
|---|---:|---:|---:|
| Rejected UTF-8 input | 125,707 | 29,319 | 4.29x faster |
| Positive UTF-8 input | 171,254 | 170,864 | neutral |

LGTM.  Comment not blocking

safere/src/main/java/org/safere/internal/Ascii.java
cushon: Done

eaftan (approved):

PR #709, opened 2026-08-15
Fix unnecessary fully qualified names

cushon:
And enable https://errorprone.info/bugpattern/UnnecessarilyFullyQualified for the safere module

I left is disabled for other submodules, since the readability of the main module is more important, and some of the other code is doing things like comparing different regex engines where there are multiple versions of `Matcher` and `Pattern` and using fully qualified names is more deliberate.


eaftan (approved):
LGTM.  Thanks for the cleanup!

PR #710, opened 2026-08-15
Implement MatchResult.hasMatch() on Matcher and SnapshotMatchResult

cushon:
This adds a missing implementation of https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/regex/Matcher.html#hasMatch()

### DIFF
--- a/safere-crosscheck/src/main/java/org/safere/crosscheck/Matcher.java
+++ b/safere-crosscheck/src/main/java/org/safere/crosscheck/Matcher.java
@@ -221,6 +221,15 @@ public final class Matcher implements MatchResult {
     return sr;
   }
 
+  /** Returns whether this matcher has a match. */
+  @Override
+  public boolean hasMatch() {
+    boolean sr = safereMatcher.hasMatch();
+    boolean jr = jdkMatcher.hasMatch();
+    checkBoolean("hasMatch", "", sr, jr);
+    return sr;
+  }
+
   // ---------------------------------------------------------------------------
   // Replacement
   // ---------------------------------------------------------------------------
--- a/safere-crosscheck/src/test/java/org/safere/crosscheck/CrosscheckTest.java
+++ b/safere-crosscheck/src/test/java/org/safere/crosscheck/CrosscheckTest.java
@@ -618,6 +618,19 @@ class CrosscheckTest {
     }
 
     @Test
+    @DisplayName("hasMatch() crosschecks matcher lifecycle state")
+    void hasMatch() {
+      Matcher m = Pattern.compile("abc").matcher("abc");
+      assertThat(m.hasMatch()).isFalse();
+      assertThat(m.find()).isTrue();
+      assertThat(m.hasMatch()).isTrue();
+      m.usePattern(Pattern.compile("xyz"));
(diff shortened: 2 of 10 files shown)


eaftan (approved):
Thanks for catching the missing `hasMatch()` API; I'm not sure how I missed it. I've pushed a commit that adds direct crosscheck coverage for `hasMatch()` across `usePattern()`, extends the stateful matcher fuzzer to exercise `hasMatch()`, fixes the lifecycle transition inventory, and corrects the test description.

While reviewing this, I came across some interesting questions about the state after a successful match followed by `usePattern()`. My reading of the JDK 26 documentation is:

- [`usePattern()`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/regex/Matcher.html#usePattern(java.util.regex.Pattern)) discards information about the groups of the previous match while maintaining the matcher's input position.
- The concrete [`Matcher.group()`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/regex/Matcher.html#group()) and [`Matcher.group(int)`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/regex/Matcher.html#group(int)) contracts explicitly say they return `null` after the matcher changes patterns and before another match is attempted.
- [`MatchResult.start(int)`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/regex/MatchResult.html#start(int)) and [`end(int)`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/regex/MatchResult.html#end(int)) explicitly say group zero is equivalent to the no-argument accessor.

For a previous match of `"a"` at `[0, 1)`, that gives this comparison:

| Call | Literal spec reading | JDK 26 | This PR |
|---|---:|---:|---:|
| `hasMatch()` | `true` | `true` | `true` |
| `start()` | `0` | `0` | `0` |
| `start(0)` | `0` | `-1` | `0` |
| `end()` | `1` | `1` | `1` |
| `end(0)` | `1` | `-1` | `1` |
| `group()` | `null` | `null` | `"a"` |
| `group(0)` | `null` | `null` | `"a"` |

A literal reading of the specification therefore produces a pretty strange state: `group()` returns `null`, even though `start()` and `end()` describe a valid previous match. This also conflicts with the usual documented equivalence between `group()` and `input.substring(start(), end())`.

I think this PR's behavior is more reasonable than either that literal reading or the JDK behavior. It preserves a coherent overall match and the documented group-zero equivalences while discarding the inner captures associated with the old pattern. The JDK's disagreement between `start()` and `start(0)`, and between `end()` and `end(0)`, also looks like an implementation bug.

I propose that we keep this PR's behavior and document it as an intentional divergence. Would you be interested in reporting these two issues upstream?

1. Clarify the internally inconsistent post-`usePattern()` specification.
2. Fix the JDK disagreement between `start()`/`start(0)` and `end()`/`end(0)`.

cushon:
> I propose that we keep this PR's behavior and document it as an intentional divergence. Would you be interested in reporting these two issues upstream?

Sounds good, thanks. I filed https://bugs.openjdk.org/browse/JDK-8390449 to try to cover all of this.

PR #712, opened 2026-08-15
Fix UTF-8 code-point boundary stepping in unanchored grapheme NFA search

cushon:
This fixes a crosscheck divergence caught by fuzzing.

For patterns with grapheme semantics, `doSearchEveryCharPosition` previously stepped byte-by-byte. On multi-byte UTF-8 sequences, this caused candidate match attempts to start inside code points.

This fix ensures stepping advances across code-point boundaries.


eaftan (approved):
Thanks for finding this and for the fix.  Your fix LGTM.  I pushed a commit to update the UTF-8 fuzzer to start generating this class of cases.

PR #714, opened 2026-08-15
Optimize pattern compilation and memory allocation

cushon:
This optimizes parts of `Pattern.compile` that showed up in profiles.

- Use primitive `IntArrayList` and in-place sorting for DFA and OnePass boundary extraction.
- Flatten predecessor graph and use $O(\text{visited})$ reachability resets during AST dominator analysis in `Prog`.
- Pack `PatchList` into primitive `long`s and reuse `stop` arrays in `Walker`.
- Skip `flatReverseDfaProg` and `OnePass` for patterns that don't need them (e.g. fast-path literal/keyword matches), similar to what #703 did for `OnePass`.

```
COMPILE_TRIALS="CompileBenchmark.compileSimple@safere-string,\
CompileBenchmark.compileMedium@safere-string,\
CompileBenchmark.compileComplex@safere-string,\
CompileBenchmark.compileAlternation@safere-string"
```

```
./run-java-benchmarks.sh CrossEngineScalingBenchmark.run --fastbuild -- -p crossEngineScalingTrial="$COMPILE_TRIALS"
```

| Benchmark          | baseline (us/op) | current (us/op) | Speedup |
| ------------------ | ---------------- | --------------- | ------- |
| compileSimple      |       7.9 ± 0.33 |      2.7 ± 0.39 |   2.99x |
| compileMedium      |         35 ± 2.7 |        31 ± 2.3 |   1.13x |
| compileComplex     |         20 ± 2.4 |        17 ± 1.7 |   1.16x |
| compileAlternation |         36 ± 2.0 |       28 ± 0.63 |   1.27x |

### DIFF
--- a/safere-crosscheck/pom.xml
+++ b/safere-crosscheck/pom.xml
@@ -115,6 +115,7 @@
                         <exclude name="GraphemeLinearTimeTest.java"/>
                         <exclude name="InstOpTest.java"/>
                         <exclude name="InstTest.java"/>
+                        <exclude name="IntArrayListTest.java"/>
                         <exclude name="InputScannerTest.java"/>
                         <exclude name="MatcherDeferredCaptureStateTest.java"/>
                         <exclude name="MatcherTransitionInventoryTest.java"/>
--- a/safere/src/main/java/org/safere/Compiler.java
+++ b/safere/src/main/java/org/safere/Compiler.java
@@ -657,19 +657,27 @@ final class Compiler extends Walker<Compiler.Frag> {
    * ("patched") to point to the next instruction. The list is threaded through the unused out/out1
    * fields of the instructions themselves.
    *
-   * <p>Encoding: {@code (instIndex << 1)} means patch {@code inst.out}; {@code (instIndex << 1) |
-   * 1} means patch {@code inst.out1}.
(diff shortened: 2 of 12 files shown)


eaftan (approved):
This looks like a great set of compilation and allocation optimizations. 

I noticed that treating descriptor presence as proof that a fast path could always execute permanently disabled OnePass for some operations. For example, `(foo)` cannot use the literal runner because it has a capture, and case-insensitive UTF-8 literals also fall back, so `lookingAt()` regressed from OnePass to the general engine path. I've pushed a commit that preserves the compile-time optimization by deferring OnePass while still allowing it to be built lazily when the advertised fast path is ineligible. I also added engine-selection regression coverage for both cases.

The submitted compile benchmark set reproduces strongly against current main. This table reports normalized time (`PR/main`), so lower is better:

|---|---:|---|
| `CompileBenchmark.compileSimple` | 0.318 | 3.14x faster |
| `CompileBenchmark.compileMedium` | 0.868 | 13.2% faster |
| `CompileBenchmark.compileComplex` | 0.797 | 20.3% faster |
| `CompileBenchmark.compileAlternation` | 0.753 | 24.7% faster |

LGTM

cushon:
Thanks for the review and for catching the mistake with OnePass!

PR #716, opened 2026-08-15
OnePass performance - pre-scale state offsets

cushon:
I had experimented wtih something similar on the DFA at one point, but the DFA has a much bigger transition table and it didn't make sense.

OnePass has fewer states, and it seems to be a measurable improvement for some cases without regressions.

---

In OnePass.search(), resolving the next state from the flat 1D transition table standardly required an integer multiplication on every character: `action = flatActions[state * numClasses + cls]`

This change pre-scales state IDs by numClasses at automaton build time:
1. Actions store `nextStateOffset` (nextState * numClasses) in bits 30-63, allowing the search loop to index `flatActions[stateOffset + cls]` with zero multiplications and advance directly via `(action >>> INDEX_SHIFT)`.
2. Sized `matchAction` table to `stateCount * numClasses`, allowing direct indexed lookups `(matchAction[stateOffset])` and eliminating `matchStateBits` bitmask checking.

```
./safere-benchmarks/scripts/compare-branch.sh --baseline main --current perf/onepass-prescaled-layout '(ApplicationBenchmark\.(uuidValidation\|csvFieldScan\|logLineParse)\|CaptureScalingBenchmark\.capture(3\|10)\|RegexBenchmark\.captureGroups).*@safere-string'
```

| Benchmark      | baseline (ns/op) | current (ns/op) | Speedup |
| -------------- | ---------------- | --------------- | ------- |
| uuidValidation |         715 ± 54 |        619 ± 54 |   1.16x |
| logLineParse   |       2638 ± 158 |      2600 ± 114 |   1.01x |
| csvFieldScan   |       3617 ± 156 |      3638 ± 129 |   0.99x |
| capture3       |        177 ± 3.2 |        177 ± 13 |   1.00x |
| capture10      |         432 ± 51 |        422 ± 56 |   1.02x |
| captureGroups  |        169 ± 3.3 |       166 ± 3.5 |   1.02x |

### DIFF
--- a/safere/src/main/java/org/safere/OnePass.java
+++ b/safere/src/main/java/org/safere/OnePass.java
@@ -36,8 +36,8 @@ final class OnePass {
 
   private static final int MAX_CAP_REGS = 2 * MAX_CAPTURE_GROUPS;
 
-  // OnePass is an optional accelerator. Keep its dense transition table within a fixed raw
-  // payload budget and fall back to the other engines when a pattern would exceed it.
+  // OnePass is an optional accelerator. Keep its dense action tables within a fixed raw payload
+  // budget and fall back to the other engines when a pattern would exceed it.
   private static final int MAX_STATES = 65_000;
   private static final long MAX_ACTION_BYTES = 16L * 1024L * 1024L;
   static final long MAX_ACTION_CELLS = MAX_ACTION_BYTES / Long.BYTES;
@@ -388,7 +388,8 @@ final class OnePass {
   }
 
   private static boolean exceedsActionBudget(int maxStates, int numClasses) {
-    return maxStates > MAX_ACTION_CELLS / numClasses;
+    // flatActions and matchActions have the same stateCount * numClasses layout.
+    return maxStates > MAX_ACTION_CELLS / 2 / numClasses;
   }
 
   private static int maxOnePassStates(Prog prog) {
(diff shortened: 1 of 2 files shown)


eaftan (approved):
This optimization looks worthwhile. Pre-scaling state offsets removes work from the hot transition loop, and the packed offsets are encoded and consumed consistently.

I noticed that this PR accidentally doubled OnePass’s action-table memory budget from 16 MiB to roughly 32 MiB by adding a second full-size dense array without accounting for it. I’ve pushed a commit that restores the intended 16 MiB total limit across both arrays and adds boundary regression coverage.

I reran the reported benchmarks against current main using the long configuration. This table reports normalized time (`PR/main`), so lower is better:

|---|---:|---|
| `RegexBenchmark.captureGroups` | 0.906 | 9.4% faster |
| `ApplicationBenchmark.logLineParse` | 0.996 | neutral |
| `ApplicationBenchmark.uuidValidation` | 0.914 | 8.6% faster |
| `ApplicationBenchmark.csvFieldScan` | 1.016 | neutral |
| `CaptureScalingBenchmark.capture3` | 0.880 | 12.0% faster |
| `CaptureScalingBenchmark.capture10` | 1.021 | neutral |

The results are in the same ballpark as your reported numbers, and there are 3 clear improvements and no clear regressions.

LGTM

PR #720, opened 2026-08-16
Add CharClassEscape to accelerate DFA self-loops on character classes

cushon:
I found this opportunity looking at vectorized String scans (https://github.com/eaftan/safere/pull/656), this adds the optimization to the existing DFA loop acceleration, which is beneficial for the existing accelerators.

- Add `StateAccelerator.CharClassEscape` record
- Collect escape sets with `AsciiBitmap` in `Dfa.analyzeStateAcceleration`
- Fast-forward DFA self-loop states with character class escapes via `InputScanner.indexOfCodePointClass

This also refactors `instanceof` checks on accelerators that are being used to avoid megamorphic virtual calls to use `instanceof` patterns patterns in switches. They should get optimized the same and have exhaustiveness checking, which is helpful to ensure that all of the fast paths are being used everywhere they should.

```
  --baseline main \
  --current dfa-char-class-acceleration \
  '(schemeReplace|markupEntityStrip|customProtocolLinkReplace|findDigitAbsent).*@safere-(string|utf8)'
```

| Benchmark                                |  baseline (ns/op) |   current (ns/op) | Speedup |
| ---------------------------------------- | ----------------- | ----------------- | ------- |
| customProtocolLinkReplace.match.1000     |       19200 ± 938 |       18893 ± 710 |   1.02x |
| customProtocolLinkReplace.match.10000    |     784385 ± 7214 |     792441 ± 6116 |   0.99x |
| customProtocolLinkReplace.match.100000   | 14929314 ± 647164 | 14792199 ± 236883 |   1.01x |
| customProtocolLinkReplace.noMatch.1000   |         2025 ± 30 |         2027 ± 28 |   1.00x |
| customProtocolLinkReplace.noMatch.10000  |       19121 ± 227 |       19013 ± 152 |   1.01x |
| customProtocolLinkReplace.noMatch.100000 |    364816 ± 13613 |     193113 ± 6295 |   1.89x |
| markupEntityStrip.match.1000             |        7136 ± 212 |        6759 ± 103 |   1.06x |
| markupEntityStrip.match.10000            |      68263 ± 1634 |      70546 ± 5250 |   0.97x |
| markupEntityStrip.match.100000           |     676285 ± 8013 |    678516 ± 10835 |   1.00x |
| markupEntityStrip.noMatch.1000           |         157 ± 6.8 |         159 ± 7.8 |   0.98x |
| markupEntityStrip.noMatch.10000          |         1112 ± 11 |         1116 ± 17 |   1.00x |
| markupEntityStrip.noMatch.100000         |       10689 ± 115 |       10856 ± 416 |   0.98x |
| findDigitAbsent.1024                     |         644 ± 4.6 |          653 ± 17 |   0.99x |
| findDigitAbsent.32768                    |      20684 ± 5171 |      20477 ± 5500 |   1.01x |
| findDigitAbsent.1048576                  |    606095 ± 14699 |    612179 ± 14332 |   0.99x |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -3178,6 +3178,44 @@
       "requirements": []
     },
     {
+      "id": "RealWorldRegexBenchmark.runBenchmark.caseInsensitiveKeywordFind.{matchLabel}.{size}",
+      "operation": "find",
+      "patterns": [
+        {
+          "java": "(?i)keyword_to_find",
+          "alternates": {
+            "dotnet": {
+              "unsupported": true,
+              "reason": ".NET non-backtracking mode cannot express Java's ASCII-only case-insensitive semantics"
+            }
+          }
+        }
+      ],
+      "inputs": [
+        "realWorldRegex.caseInsensitiveKeyword.{matchLabel}.{size}"
+      ],
+      "resultConsumption": "boolean",
+      "measurement": {
+        "mode": "averageTime",
+        "timingUnit": "nanoseconds",
+        "constraints": [
+          "prematerializedInput"
+        ]
+      },
+      "axes": {
+        "matchLabel": [
+          "match",
+          "noMatch"
+        ],
+        "size": [
+          1000,
+          10000,
+          100000
+        ]
+      },
+      "requirements": []
+    },
+    {
       "id": "RealWorldRegexBenchmark.runBenchmark.boundedNameMatch.{matchLabel}.{size}",
       "operation": "replaceAll",
       "patterns": [
--- a/safere-vector-benchmarks/src/main/java/org/safere/Utf8ScannerBenchmarkAccess.java
+++ b/safere-vector-benchmarks/src/main/java/org/safere/Utf8ScannerBenchmarkAccess.java
@@ -22,7 +22,7 @@ public final class Utf8ScannerBenchmarkAccess {
 
   /** Returns the first byte position in the configured code-point ranges. */
   public int indexOfCodePointClass(int start) {
-    return scanner.indexOfCodePointClass(ranges, bitmap0, bitmap1, start);
+    return scanner.indexOfCodePointClass(ranges, bitmap0, bitmap1, start, scanner.length());
   }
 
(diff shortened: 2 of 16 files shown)


eaftan (approved):
This looks good. The DFA enables the new character-class escape accelerator only after proving that every non-ASCII equivalence class remains in the current state, and the bounded String/UTF-8 scanner changes preserve the search limit. The sealed dispatch refactor is also consistent across accelerator types.

I reproduced the headline benchmark result. I also benchmarked the new `caseInsensitiveKeywordFind` workload across its matching and rejected inputs on both representations. This table reports normalized time (`PR/main`), so lower is better:

| Workload | Input | Size | String PR/main | UTF-8 PR/main | Result |
|---|---|---:|---:|---:|---|
| `customProtocolLinkReplace` | Reject | 100K | 0.478 | — | 2.09x faster |
| `markupEntityStrip` | Match | 100K | 0.952 | — | neutral / 4.8% lower |
| `findDigitAbsent` | Reject | 1 MiB | 0.963 | — | neutral / 3.7% lower |
| `caseInsensitiveKeywordFind` | Match | 1K | 1.020 | 1.000 | neutral |
| `caseInsensitiveKeywordFind` | Match | 10K | 1.016 | 1.004 | neutral |
| `caseInsensitiveKeywordFind` | Match | 100K | 0.978 | 0.982 | neutral |
| `caseInsensitiveKeywordFind` | Reject | 1K | 0.982 | 0.999 | neutral |
| `caseInsensitiveKeywordFind` | Reject | 10K | 1.000 | 1.009 | neutral |
| `caseInsensitiveKeywordFind` | Reject | 100K | 1.002 | 0.998 | neutral |

The example benchmark command does not measure UTF-8: its replacement trials resolve only to the String runner, and `findDigitAbsent` is explicitly String-only. The em dashes above indicate combinations that command did not measure.

The caseInsensitiveKeywordFind results are neutral. This workload adds UTF-8 benchmark coverage, but its rejected unanchored search remains in the DFA start state, while the new per-state accelerator runs only in non-start states. It therefore does not exercise the optimization on UTF-8. I’m fine with that; I mention it only to clarify what the benchmark demonstrates.

LGTM

PR #721, opened 2026-08-16
Vectorize multi-character case-insensitive ASCII prefix search on UTF-8

cushon:
I found this opportunity looking at vectorized String scans (https://github.com/eaftan/safere/pull/656), this adds the optimization to the existing UTF-8 `byte[]` paths.

- Add `Ascii.regionMatchesIgnoreCase` overload for `byte[]`
- Implement 32-lane candidate first-character filtering in `ByteVectorScan.indexOfIgnoreCase` for multi-character ASCII prefixes
- Enable vector acceleration for multi-character prefixes in `Utf8InputScanner.indexOfIgnoreCase`

```
  --baseline benchmark-case-insensitive-utf8 \
  --current utf8-multi-char-prefix \
  'caseInsensitiveKeywordFind.*@safere-utf8'
```

| Benchmark                                 | baseline (ns/op) | current (ns/op) | Speedup |
| ----------------------------------------- | ---------------- | --------------- | ------- |
| caseInsensitiveKeywordFind.match.1000     |         94 ± 1.4 |        95 ± 2.6 |   1.00x |
| caseInsensitiveKeywordFind.match.10000    |         99 ± 4.0 |        97 ± 5.5 |   1.02x |
| caseInsensitiveKeywordFind.match.100000   |         96 ± 2.0 |         95 ± 10 |   1.01x |
| caseInsensitiveKeywordFind.noMatch.1000   |        405 ± 2.5 |       401 ± 2.9 |   1.01x |
| caseInsensitiveKeywordFind.noMatch.10000  |        3873 ± 97 |       171 ± 3.1 |   22.7x |
| caseInsensitiveKeywordFind.noMatch.100000 |     39804 ± 1614 |       2046 ± 15 |   19.5x |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -3178,6 +3178,44 @@
       "requirements": []
     },
     {
+      "id": "RealWorldRegexBenchmark.runBenchmark.caseInsensitiveKeywordFind.{matchLabel}.{size}",
+      "operation": "find",
+      "patterns": [
+        {
+          "java": "(?i)keyword_to_find",
+          "alternates": {
+            "dotnet": {
+              "unsupported": true,
+              "reason": ".NET non-backtracking mode cannot express Java's ASCII-only case-insensitive semantics"
+            }
+          }
+        }
+      ],
+      "inputs": [
+        "realWorldRegex.caseInsensitiveKeyword.{matchLabel}.{size}"
+      ],
+      "resultConsumption": "boolean",
+      "measurement": {
+        "mode": "averageTime",
+        "timingUnit": "nanoseconds",
+        "constraints": [
+          "prematerializedInput"
+        ]
+      },
+      "axes": {
+        "matchLabel": [
+          "match",
+          "noMatch"
+        ],
+        "size": [
+          1000,
+          10000,
+          100000
+        ]
+      },
+      "requirements": []
+    },
+    {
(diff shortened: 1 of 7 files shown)


eaftan (dismissed):
This optimization makes sense. It extends the SIMD candidate-filtering approach we already use for case-insensitive ASCII prefix searches on String inputs to UTF-8 byte-array inputs. Multi-character UTF-8 prefixes previously used the scalar linear-time KMP scan; this adds a fast path that scans 32 bytes at a time for a relatively rare character from the prefix and verifies the full prefix only at those candidate positions. The existing KMP path remains available as the fallback.

I found a worst-case complexity issue in the new vector scan. If the selected character occurred throughout the input, the code could compare a long prefix at every position, making the work proportional to the input length times the prefix length rather than linear in the input length. I also found an integer-overflow bug in the bounds check for a candidate near the end of a roughly 2 GiB input, which could allow an out-of-range access. I've pushed a commit that limits candidate-verification work before falling back to KMP, makes the bounds check overflow-safe, and adds regression coverage.

I also reproduced the benchmark on current main. Time is normalized as `PR/main`, so lower is better:

| Input | Size | PR/main | Result |
|---|---:|---:|---|
| Match | 1K | 1.038 | neutral |
| Match | 10K | 1.003 | neutral |
| Match | 100K | 1.018 | neutral |
| Reject | 1K | 1.029 | neutral |
| Reject | 10K | 0.031 | 32.1x faster |
| Reject | 100K | 0.028 | 35.4x faster |

LGTM

eaftan (approved):

PR #722, opened 2026-08-16
Add a script to run benchmarks comparing two commits

cushon:
This supports comparing two branches or commits and presenting the results, e.g.

```
./safere-benchmarks/scripts/compare-branch.sh   --baseline main   --current dfa-char-class-acceleration   '(jsonBlock|templateTagMatch)\.match.*@safere-string'
```

This is a slightly cleaned up version of a workflow I'd been using to benchmark PRs. I'm not opinionated about the details here if you have suggestions for a better workflow or better ways to share benchmark results.

### DIFF
--- a/.agents/skills/daily-repo-assist/SKILL.md
+++ b/.agents/skills/daily-repo-assist/SKILL.md
@@ -130,12 +130,17 @@ For each trusted open non-draft PR:
      post-merge/pre-fix `HEAD` to final `HEAD`.
 
 5. For optimization PRs, reproduce only new benchmark claims made since the last review/comment.
-   Use `./run-java-benchmarks.sh` only. Never run benchmarks in parallel.
+   For targeted SafeRE nanosecond workloads with identical benchmark definitions at both revisions,
+   prefer `safere-benchmarks/scripts/compare-branch.sh`; otherwise use
+   `./run-java-benchmarks.sh` directly. Never run benchmarks in parallel.
    - Baseline: current `origin/main`.
    - Experiment: PR branch after merging current `origin/main`, plus local review fixes.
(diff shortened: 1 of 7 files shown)


eaftan (approved):
Thanks, this is a good workflow to standardize. 

I found several cases where the submitted script could print a plausible but incorrect comparison:

- Maven build products survive a Git checkout, so the second run could reuse classes or a benchmark JAR built from the baseline and effectively benchmark the same implementation twice.
- Relative refs such as `HEAD~1` were resolved after checking out the baseline, so they could identify a different commit from the one requested by the caller.
- String and UTF-8 execution variants were relabeled into the same baseline/current identities, allowing distinct measurements to overwrite or be compared as though they were the same workload.
- Revisions with different workload data, benchmark harnesses, runner settings, or build definitions could be compared directly, attributing an experimental change to library performance.
- Scores were divided without normalizing units or checking measurement type. For example, `1000 ns/op` and `1 us/op` could be reported as a 1000x difference even though they are equal, and latency could be compared with throughput despite their opposite directions.

I've pushed commits that fix those cases. The script now resolves both refs to immutable commits before moving `HEAD`, uses detached checkouts so branches in other worktrees are supported, cleans and rebuilds around revision changes, rejects differing workload or harness inputs, limits each run to one SafeRE execution variant, and normalizes compatible units while rejecting incompatible ones. It also supports the project's `--long` confirmation mode, preserves benchmark identities and mixed-unit labels, restores the original branch or detached commit, and removes temporary files on every exit path.

I'll start using this for targeted PR before/after measurements when both revisions share the same benchmark definitions. I've documented that workflow and updated the benchmark-report, PR-scout, and daily-report instructions to prefer this script in that situation. 

LGTM

eaftan (approved):

PR #723, opened 2026-08-16
Replace UTF8_FIND_SITE MutableCallSite with direct diagnostics dispatch

cushon:
I noticed this validating SafeRE in a test that was using proguard, which was pruning `findWithDiagnostics`, causing `Pattern`'s class initializer to fail.

I can fix the proguard config, but I also don't think the `MethodHandle` here has a performance benefit over this approach, which should get constant and optimized out.


eaftan (approved):
This simplification looks good. The direct branch keeps the existing diagnostics listener call site as the source of truth while making both UTF-8 find implementations visible to Proguard, avoiding the reflective-method retention problem.

I also checked the disabled-diagnostics hot path against current main. This table reports normalized time (`PR/main`), so lower is better:

|---|---:|---|
| `RegexBenchmark.emailFind@safere-utf8` | 0.978 | neutral |
| `mapFieldPath.match.1000@safere-utf8` | 0.997 | neutral |
| `mapFieldPath.noMatch.1000@safere-utf8` | 0.965 | neutral / 3.5% lower |

LGTM

PR #724, opened 2026-08-16
Optimize trivial matches

cushon:
This optimizes some trivial patterns on short inputs. This adds fast paths for pure literal patterns and single-character classes to bypass some of the `Matcher` setup, and also recognizes length-bounded patterns where positive matches are required to be longer than a certain length to early return on short inputs.

Also, move initialization of immutable `PreparedMatchRunner` classes into `Pattern.compile` instead of repeating it when `Matcher`s are constructed.

I noticed this taking a look at the overall benchmarks comparing SafeRE to JDK and re2-ffm, I think this might close the gap a bit with the JDK for some trivial patterns on short inputs. I'm not sure how general this is, but trivial patterns like pure literals do sometimes get used, and this doesn't add that much additional logic. Moving `PreparedMatchRunner` to `Pattern` is probably a more unambiguous win if the other fast paths seem too specialized though.

```
./safere-benchmarks/scripts/compare-branch.sh   --baseline main   --current perf/trivial-match-specialization   '(literalMatch|charClassMatch|uuidValidation|apiRouteMatch|httpSmall).*@safere-string'
```

| Benchmark      | baseline (ns/op) | current (ns/op) | Speedup |
| -------------- | ---------------- | --------------- | ------- |
| literalMatch   |         28 ± 5.7 |        26 ± 4.4 |   1.05x |
| charClassMatch |         42 ± 5.1 |        39 ± 3.5 |   1.08x |
| uuidValidation |         691 ± 57 |        579 ± 24 |   1.19x |
| apiRouteMatch  |         923 ± 27 |       595 ± 8.0 |   1.55x |
| httpSmall      |         105 ± 13 |        77 ± 2.0 |   1.36x |

### DIFF
--- a/safere/src/main/java/org/safere/MatchDescriptor.java
+++ b/safere/src/main/java/org/safere/MatchDescriptor.java
@@ -28,14 +28,16 @@ record MatchDescriptor(
     String literalMatch,
     CharClassScanInfo singleCharClass,
     KeywordAlternation keywordAlternation,
-    CharClassMatchInfo charClassMatch) {
+    CharClassMatchInfo charClassMatch,
+    int minMatchLength) {
 
-  static final MatchDescriptor NONE = new MatchDescriptor(null, null, null, null);
+  static final MatchDescriptor NONE = new MatchDescriptor(null, null, null, null, 0);
 
   boolean hasFastPath() {
     return literalMatch != null
         || singleCharClass != null
         || keywordAlternation != null
-        || charClassMatch != null;
+        || charClassMatch != null
+        || minMatchLength > 0;
   }
 }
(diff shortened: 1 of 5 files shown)


eaftan (dismissed):
This looks good. Precomputing the immutable prepared runners on `Pattern` is a clean reduction in per-matcher setup, and the literal, one-character-class, and minimum-length shortcuts preserve the normal String, UTF-8, region, bounds, and matcher-state behavior.

I reproduced the five reported workloads against current main using the longer benchmark configuration. Time is normalized as `PR/main`, so lower is better:

| Workload | PR/main | Result |
|---|---:|---|
| `literalMatch` | 0.845 | 1.18x faster |
| `charClassMatch` | 0.950 | 1.05x faster |
| `uuidValidation` | 0.845 | 1.18x faster |
| `apiRouteMatch` | 0.819 | 1.22x faster |
| `httpSmall` | 0.782 | 1.28x faster |

LGTM

eaftan (approved):

PR #730, opened 2026-08-17
Add work limits and optimize Matcher.indexOfIgnoreCase, and consolidate work limit handling

cushon:
This adds a `WorkLimit` helper to consolidate checks like the ones added in https://github.com/eaftan/safere/pull/721#pullrequestreview-4948390289 (thanks for the catch!)

- Centralizes overflow-safe candidate bounds checking (`candidateInBounds`) and work budget accounting (`forRemaining`, `isExhausted`) across `ByteVectorScan`, `ByteSwarScan`, `Utf8InputScanner`, `ShortVectorScan`, and `ShortSwarScan`.
- An audit of similar scan paths found that `Matcher.indexOfIgnoreCase` for `String` inputs was using a naive sliding window without work limits or KMP fallback. Added rarest-character anchoring, a short scalar window with exhaustion caching, and linear Knuth-Morris-Pratt (`Ascii.indexOfLinearIgnoreCase`) fallback to guarantee $O(N)$ worst-case time without regressing repeated-match throughput.
- Implemented 64-bit SWAR acceleration (`ByteSwarScan.indexOfIgnoreCase`) for multi-character case-insensitive prefix search on UTF-8 when incubator vector acceleration is unsupported or inactive.
- Added deterministic scaling tests in `SearchScalingRegressionTest` to verify linear work scaling across both `String` and `UTF-8`.

```
./safere-benchmarks/scripts/compare-branch.sh   --baseline origin/main   --current refactor/candidate-verifier   'caseInsensitiveKeyword.*@safere-string'
```

| Benchmark                                    | baseline (ns/op) | current (ns/op) | Speedup |
| -------------------------------------------- | ---------------- | --------------- | ------- |
| ApplicationBenchmark.caseInsensitiveKeywords |        477 ± 5.2 |        494 ± 28 |   0.97x |
| caseInsensitiveKeyword.match.1000            |       9743 ± 586 |     11259 ± 511 |   0.87x |
| caseInsensitiveKeyword.match.10000           |    100140 ± 4113 |  117224 ± 10253 |   0.85x |
| caseInsensitiveKeyword.match.100000          |  1028506 ± 13535 | 1148462 ± 36855 |   0.90x |
| caseInsensitiveKeyword.noMatch.1000          |         520 ± 25 |        119 ± 12 |   4.36x |
| caseInsensitiveKeyword.noMatch.10000         |       4876 ± 171 |        450 ± 40 |   10.8x |
| caseInsensitiveKeyword.noMatch.100000        |     37390 ± 1567 |       3587 ± 26 |   10.4x |
| caseInsensitiveKeywordFind.match.1000        |        160 ± 9.6 |       162 ± 8.5 |   0.98x |
| caseInsensitiveKeywordFind.match.10000       |         164 ± 12 |        165 ± 15 |   1.00x |
| caseInsensitiveKeywordFind.match.100000      |        158 ± 8.2 |        157 ± 10 |   1.01x |
| caseInsensitiveKeywordFind.noMatch.1000      |         399 ± 12 |         97 ± 12 |   4.13x |
| caseInsensitiveKeywordFind.noMatch.10000     |       3812 ± 221 |       397 ± 8.2 |   9.60x |
| caseInsensitiveKeywordFind.noMatch.100000    |     36010 ± 1063 |       3875 ± 82 |   9.29x |


eaftan (commented):
The `WorkLimit` consolidation makes sense, and the new candidate searches produce substantial rejected-input gains.

I found two cases where the submitted String path could still take quadratic time. Repeated `find()` with a one-character case-insensitive prefix searched the remaining suffix for an absent case variant after every match. With a multi-character prefix, sparse false candidates could repeatedly search for the same distant opposite-case anchor. I've pushed a commit that changes the one-character path to a single scan, retains each discovered anchor until it is consumed, and adds scaling coverage for both cases.

I compared current main with the corrected PR using the long benchmark configuration. Time is normalized as `PR/main`, so lower is better:

| Workload | Representation | 1K | 10K | 100K |
|---|---|---:|---:|---:|
| `caseInsensitiveKeyword.match` | String | 1.173 | 1.214 | 1.260 |
| `caseInsensitiveKeyword.noMatch` | String | 0.299 | 0.112 | 0.150 |
| `caseInsensitiveKeywordFind.match` | String | 1.018 | 1.034 | 1.066 |
| `caseInsensitiveKeywordFind.noMatch` | String | 0.281 | 0.138 | 0.151 |
| `caseInsensitiveKeywordFind.match` | UTF-8 | 0.963 | 0.952 | 0.925 |
| `caseInsensitiveKeywordFind.noMatch` | UTF-8 | 0.430 | 0.432 | 0.423 |

The `caseInsensitiveKeyword` workload calls `replaceAll` to remove every match; `caseInsensitiveKeywordFind` calls `find()` once. The rejected searches improve by 3.35–8.95x for String and about 2.3x for UTF-8. Successful `find()` is close to neutral for String and modestly faster for UTF-8. 

However, `caseInsensitiveKeyword.match` is 17–26% slower, with the penalty growing by input size. That's a bit larger than the regression in the PR description.

What do you think about this tradeoff? The rejection gains are meaningful, but I'm a little concerned about the regression in matching `replaceAll` workloads.  Would you be willing to take an optimization pass over those to try to get back to neutral on those workloads?

cushon:
I took some more optimization passes over this and found some places to improve.

* precompute rarity anchors and KMP failure tables on `LiteralPreparedRunner` at compile time, and extended `literalReplaceFastPath` to stream case-folded replacements directly into `StringBuilder`
* search loop - verified candidates in-place, used single-branch bitwise checks `((c | 0x20) == low)`, skipped redundant anchor re-reads, and deferred `WorkLimit` bookkeeping so it is initialized lazily
* use pre-lowercased prefixes in `Ascii.regionMatchesIgnoreCase` / `Utf16` (`(c != p && toLowerCase(c) != p)`), folded KMP characters once per text position, and switched `Ascii.isWordChar`/`isAlnum` to 64-bit word bitmaps

This fixes the `caseInsensitiveKeyword` scaling issue, it's still slower but it doesn't get worse for longer inputs.

I think the tradeoff here is probably worth it, `caseInsensitiveKeyword` is deliberately pathological, it has a lot of matches separated by a single space. The optimization has to do some additional bookkeeping to decide whether to accelerate skips. It's going to be faster to stay in the scalar loop and not accelerate skips if there is a single character between matches. But the benchmarks also show this is a lot better for sparser matches.

Let me know what you think, it'd be good to get a second opinion on the state of this PR overall with the new micro-optimizations.

| Benchmark | baseline (ns/op) | current (ns/op) | Speedup |
| :--- | :--- | :--- | :--- |
| ApplicationBenchmark.caseInsensitiveKeywords | 480 ± 26 | 482 ± 14 | 1.00x |
| caseInsensitiveKeyword.match.1000 | 9051 ± 175 | 10739 ± 577 | 0.84x |
| caseInsensitiveKeyword.match.10000 | 94586 ± 3067 | 112659 ± 5678 | 0.84x |
| caseInsensitiveKeyword.match.100000 | 937269 ± 24194 | 1059125 ± 40038 | 0.88x |
| caseInsensitiveKeyword.noMatch.10000 | 4791 ± 101 | 378 ± 13 | 12.7x |
| caseInsensitiveKeywordFind.match.10000 | 154 ± 7.1 | 155 ± 5.2 | 0.99x |
| caseInsensitiveKeywordFind.noMatch.10000 | 3615 ± 16 | 351 ± 28 | 10.3x |

eaftan (approved):
Thanks for taking another optimization pass over this. I reran the same workload family against current main using the long configuration. Time is normalized as `PR/main`, so lower is better:

| Workload | Representation | 1K | 10K | 100K |
|---|---|---:|---:|---:|
| `caseInsensitiveKeyword.match` | String | 1.122 | 1.077 | 1.158 |
| `caseInsensitiveKeyword.noMatch` | String | 0.239 | 0.100 | 0.128 |
| `caseInsensitiveKeywordFind.match` | String | 0.987 | 1.074 | 1.043 |
| `caseInsensitiveKeywordFind.noMatch` | String | 0.230 | 0.122 | 0.150 |
| `caseInsensitiveKeywordFind.match` | UTF-8 | 0.955 | 0.938 | 0.949 |
| `caseInsensitiveKeywordFind.noMatch` | UTF-8 | 0.449 | 0.425 | 0.416 |

This addresses my previous scaling concern. The dense matching `replaceAll` case is still 8–16% slower, but I agree that the workload is deliberately pathological, and the percentage no longer grows with input size.  I think this is a reasonable tradeoff. 

LGTM

PR #732, opened 2026-08-18
Bypass inner capture resolution for literal and group-zero replacements

cushon:
This change bypasses capture resolution in cases where it isn't needed, such as when only group zero (`$0`) is referenced in replacements or for parenthesized literals like `(parens)` during `split()` and literal find.

This was found by looking for other opportunities like the one caught in https://github.com/eaftan/safere/pull/702#pullrequestreview-4947955369, where e.g. `numCaptures() > 0` was being used instead of `needsCaptures()` and the use of `$0` was forcing full capture resolution to run unnecessarily.

I don't think there are any benchmark cases that exclusively use `$0` and benefit from this, I added a work counter test to verify it.


eaftan (dismissed):
This makes sense: `split()` needs only match boundaries, and literal or `$0` replacement templates need only group zero, so patterns that declare inner groups should not have to resolve those captures unless an operation actually consumes them.

I noticed one performance issue in the literal replacement path. The no-match preflight found the first literal occurrence, but the replacement loop discarded that index and searched again from the beginning. A late `replaceFirst()` match could therefore scan almost the entire prefix twice. I've pushed a commit that reuses the preflight result and adds regression coverage for that case.

LGTM

eaftan (approved):

cushon:
> The no-match preflight found the first literal occurrence, but the replacement loop discarded that index and searched again from the beginning. 

Thanks for the catch!

PR #734, opened 2026-08-18
Add more scaling regression tests

cushon:
This adds a few more `SearchScalingRegressionTest` `WorkCounter` tests inspired by recent optimizations, to guard scaling invariants across accelerators and execution modes.

* DFA self-loop accelerators: verify linear scaling on String and UTF-8 for quote (`"[^"]*"`) and newline (`[^,\n]*\n`) escape fast-forwards without DFA state thrashing
* Single character class direct scan: verify that positive (`\d`) and negated (`[^a-z]`) character class searches scale linearly and bypass DFA setup overhead
* Encoding parity scaling: assert that searching UTF-16 and native UTF-8 bytes maintains asymptotic parity with latin1
* SIMD / SWAR boundaries: sweep string lengths from 1 to 65 to verify continuous linear scaling across 16-byte (SSE) and 32-byte (AVX2) vector lane transitions

### DIFF
--- a/safere/src/main/java/org/safere/StringInputScanner.java
+++ b/safere/src/main/java/org/safere/StringInputScanner.java
@@ -29,6 +29,16 @@ final class StringInputScanner implements InputScanner {
 
   @Override
   public int indexOfAscii(int ascii, int fromIndex, int limit) {
+    if (WorkCounterConfig.ENABLED) {
+      int end = Math.min(limit, text.length());
+      for (int i = Math.max(0, fromIndex); i < end; i++) {
+        WorkCounter.record();
+        if (text.charAt(i) == ascii) {
+          return i;
+        }
+      }
+      return -1;
+    }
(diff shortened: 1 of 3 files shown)


eaftan (approved):
Thank you for contributing these tests!  This is useful coverage for performance-sensitive paths that can regress without changing matching results. 

I noticed that two of the new tests did not exercise what their names described. The UTF-8 state-accelerator test counted only the surrounding DFA work because the accelerated scan was invisible to `WorkCounter`, so it reported four units for both input sizes. The vector-boundary test used a String matcher, which took the scalar String scanner rather than a Vector or SWAR kernel.

I've pushed a commit that makes the accelerated ASCII scans visible in work-counter builds and checks the actual SWAR and Vector character-class kernels across lengths 1 through 65, for both an absent class and a match in the final position. LGTM

Issue #564, opened 2026-07-06
More analysis of generalized character class prefix optimization


cushon:
https://github.com/eaftan/safere/commit/e29b7b7ee9016b157b8c49428e12d6b777dfaeae showed wins on some benchmarks, but I think it may have regressed others.

I think it's worth checking for a larger suite of benchmarks whether it caused regressions, or just backing it out and revisiting later.

`Matcher.charClassContains` is doing a bisect which is more expensive than what was being done before for ASCII. The changes to `indexOfCharClass` make it more expensive, it will now do a bisect for for unicode characters.

cushon:
#766 restored character class acceleration for cases where it's helpful.

There's a related open PR in #803 that will improve character class optimization a little more, once that lands I think this can be closed out.

cushon:
#803 is merged

PR #735, opened 2026-08-18
Generalize character-class prefix acceleration to Unicode (#564)

cushon:
See #564. This is a re-do of https://github.com/eaftan/safere/commit/37f5e828a157c203bfd47125e1c69f91cd5316d0, but updated to integrate into the new structure for accelerators, and preserving ASCII fast paths to ensure it doesn't regress.

- Replace ASCII-only AsciiBitmap prefix representation with CharClassScanInfo in StartDescriptor, Pattern, StringStartAccelerator, and Utf8StartAccelerator.
- Extract full Unicode character-class prefix range info in Pattern.extractCharClassPrefix with selectivity guard.
- In StringStartAccelerator, use precomputed 128-boolean lookup for ASCII characters and binary range search on code points for non-ASCII.
- Preserve zero extra branch overhead in the hot ASCII scanning loop.
- Support supplementary code points (surrogate pairs) in String candidate scanning.
- Add unit and cross-check tests for Unicode and supplementary character-class prefix acceleration.

---

The optimization shows a ~3.4x speedup for rejecting non-matching inputs with non-ASCII character class prefixes.

`cjk.absent` is neutral because it was already being rejected by `RejectPrefilter`.

The slowdown for late, long matches is the expected tradeoff with the accelerator, which has to do an extra loop over the input before failing to reject the input and falling through to the DFA. Future work to vectorize accelerators will reduce that penalty (and increase the benefit of the cases where it helps).

```
./safere-benchmarks/scripts/compare-branch.sh --baseline bench/unicode-prefix-baseline 'UnicodePrefixBenchmark.*@safere-string'
```

| Benchmark                                        | baseline (ns/op) |  current (ns/op) | Speedup |
| ------------------------------------------------ | ---------------- | ---------------- | ------- |
| UnicodePrefixBenchmark.alphabetic.absent.1024    |        2085 ± 34 |         615 ± 28 |   3.39x |
| UnicodePrefixBenchmark.alphabetic.absent.32768   |      62441 ± 296 |      18361 ± 581 |   3.40x |
| UnicodePrefixBenchmark.alphabetic.absent.1048576 |  1998883 ± 23660 |   592322 ± 34292 |   3.37x |
| UnicodePrefixBenchmark.alphabetic.late.1024      |        2075 ± 65 |         741 ± 51 |   2.80x |
| UnicodePrefixBenchmark.alphabetic.late.32768     |      62341 ± 587 |     32815 ± 1979 |   1.90x |
| UnicodePrefixBenchmark.alphabetic.late.1048576   |  2157491 ± 56179 | 2832737 ± 120769 |   0.76x |
| UnicodePrefixBenchmark.cjk.absent.1024           |         639 ± 36 |        625 ± 9.2 |   1.02x |
| UnicodePrefixBenchmark.cjk.absent.32768          |      19019 ± 646 |     19371 ± 1166 |   0.98x |
| UnicodePrefixBenchmark.cjk.absent.1048576        |    988076 ± 3940 |    991033 ± 5549 |   1.00x |
| UnicodePrefixBenchmark.cjk.late.1024             |        3287 ± 73 |        2994 ± 76 |   1.10x |
| UnicodePrefixBenchmark.cjk.late.32768            |     99256 ± 4991 |    123994 ± 1532 |   0.80x |
| UnicodePrefixBenchmark.cjk.late.1048576          |  3037570 ± 97053 |  3005064 ± 47037 |   1.01x |

Results for ASCII look neutral:

```
./safere-benchmarks/scripts/compare-branch.sh --baseline origin/main '(versionList|recitation|charClassMatch|literalMatch).*@safere-string'
```

| Benchmark                     | baseline (ns/op) | current (ns/op) | Speedup |
| ----------------------------- | ---------------- | --------------- | ------- |
| RegexBenchmark.literalMatch   |         27 ± 4.9 |        28 ± 5.0 |   0.97x |
| RegexBenchmark.charClassMatch |         41 ± 4.5 |        41 ± 5.7 |   1.00x |
| versionList.match.1000        |      12627 ± 417 |     11768 ± 528 |   1.07x |
| versionList.match.10000       |    121533 ± 3764 |   119359 ± 3675 |   1.02x |
| versionList.match.100000      |  1210596 ± 31325 | 1237781 ± 57823 |   0.98x |
| versionList.noMatch.1000      |       3925 ± 136 |      3877 ± 127 |   1.01x |
| versionList.noMatch.10000     |     36867 ± 1272 |    37338 ± 2714 |   0.99x |
| versionList.noMatch.100000    |    357464 ± 3359 |  365956 ± 16294 |   0.98x |
| recitation.match.1000         |       8327 ± 228 |      8538 ± 385 |   0.98x |
| recitation.match.10000        |     76567 ± 3612 |    75454 ± 1084 |   1.01x |
| recitation.match.100000       |   786209 ± 41879 |  773831 ± 29598 |   1.02x |
| recitation.noMatch.1000       |       3877 ± 168 |      3851 ± 107 |   1.01x |
| recitation.noMatch.10000      |     36648 ± 1915 |     36050 ± 493 |   1.02x |
| recitation.noMatch.100000     |   363327 ± 13048 |   357283 ± 4494 |   1.02x |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -9763,7 +9763,15 @@
       "id": "UnicodePrefixBenchmark.alphabetic.absent.{size}",
       "operation": "find",
       "patterns": [
-        "[\\p{IsAlphabetic}]+\\d+"
+        {
+          "java": "[\\p{IsAlphabetic}]+\\d+",
+          "alternates": {
+            "dotnet": {
+              "unsupported": true,
+              "reason": ".NET does not expose Java's IsAlphabetic Unicode property"
+            }
+          }
+        }
       ],
       "inputs": [
         "unicodePrefix.digitsAbsent.{size}"
@@ -9794,7 +9802,15 @@
       "id": "UnicodePrefixBenchmark.alphabetic.late.{size}",
       "operation": "find",
       "patterns": [
-        "[\\p{IsAlphabetic}]+\\d+"
+        {
+          "java": "[\\p{IsAlphabetic}]+\\d+",
+          "alternates": {
+            "dotnet": {
+              "unsupported": true,
+              "reason": ".NET does not expose Java's IsAlphabetic Unicode property"
+            }
+          }
+        }
       ],
       "inputs": [
         "unicodePrefix.alphabeticLate.{size}"
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -2379,6 +2379,72 @@
         "suffix": ":8080",
         "length": "{size}"
       }
+    },
+    {
+      "id": "unicodePrefix.digitsAbsent.{size}",
+      "axes": {
+        "size": [
+          1024,
+          32768,
+          1048576
+        ]
+      },
+      "recipe": {
+        "kind": "repeatToLength",
+        "unit": "0123456789 9876543210 1357924680 ",
+        "length": "{size}"
+      },
+      "shared": true
+    },
+    {
+      "id": "unicodePrefix.alphabeticLate.{size}",
+      "axes": {
+        "size": [
+          1024,
+          32768,
+          1048576
+        ]
+      },
+      "recipe": {
+        "kind": "suffixToLength",
+        "prefixUnit": "0123456789 9876543210 1357924680 ",
+        "suffix": "é123",
+        "length": "{size}"
+      },
+      "shared": true
+    },
+    {
+      "id": "unicodePrefix.englishProseAbsent.{size}",
+      "axes": {
+        "size": [
+          1024,
+          32768,
+          1048576
+        ]
+      },
+      "recipe": {
+        "kind": "repeatToLength",
+        "unit": "The quick brown fox jumps over the lazy dog. Standard English prose without CJK characters. ",
+        "length": "{size}"
+      },
+      "shared": true
+    },
+    {
+      "id": "unicodePrefix.cjkLate.{size}",
+      "axes": {
+        "size": [
+          1024,
+          32768,
+          1048576
+        ]
+      },
+      "recipe": {
+        "kind": "suffixToLength",
+        "prefixUnit": "The quick brown fox jumps over the lazy dog. Standard English prose without CJK characters. ",
+        "suffix": "你好123",
+        "length": "{size}"
+      },
+      "shared": true
     }
   ],
   "workloads": [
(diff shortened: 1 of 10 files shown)


eaftan (commented):
This looks like a nice way to revive the Unicode character-class prefix optimization while preserving the existing ASCII hot path. The String scanner keeps its direct ASCII table and only uses code-point range lookup for non-ASCII input, while the UTF-8 path reuses its existing code-point class scanner.

I found one benchmark-collection issue: the new `\p{IsAlphabetic}` workloads were runnable for .NET even though .NET cannot compile that Java Unicode property. I've pushed a commit that marks both workloads unsupported for the .NET profile and verifies that the resolved execution plan excludes them.

I dug further into the regression on `UnicodePrefixBenchmark.alphabetic.late.1048576`. As far as I can tell, this does not scan the long prefix twice: after the fast scan finds the late alphabetic candidate (`é`), `doFindCore()` passes that position to the DFA as `effectiveStart`, so the DFA starts near the end. CPU profiling likewise put the time in the String character-class scanner rather than the DFA.

The issue turned out to be HotSpot's treatment of the scanner's rarely taken non-ASCII branch. After observing a long ASCII run, C2 compiled the loop on the assumption that `ch < 128` would remain true and guarded the other path with an uncommon trap. Reaching the final `é` invalidated that compilation, after which HotSpot generated a larger general version of the loop that scanned the preceding ASCII characters much more slowly.

I've pushed commit 3d27185546f295c7b7aaadc824334286c8246355, which keeps Unicode decoding and range lookup outside the dedicated ASCII scan loop. The ASCII loop now returns at the first non-ASCII position, the outer loop checks that code point, and then resumes ASCII scanning. With the long benchmark configuration:

| Version | Time (ms/op) | Relative to main |
|---|---:|---:|
| Current main | 1.307 | 1.000 |
| Submitted PR | 1.853 | 1.418 (41.8% slower) |
| PR + split scanner | 0.441 | 0.337 (2.97x faster) |

The 1 MiB all-ASCII rejection control remains neutral versus the submitted PR (0.387 ms versus 0.381 ms). I also repeated the late-match experiment with U+0100, forcing a UTF-16-backed String; the split reduced that case from 1.995 ms to 0.425 ms. That suggests this addresses the underlying ASCII/Unicode-path interaction rather than merely special-casing Latin-1.

I'd like to do additional benchmarking to make sure we retained the gains from your original PR, but I need to run that overnight. I just wanted to let you know where I am with this.

cushon:
> The issue turned out to be HotSpot's treatment of the scanner's rarely taken non-ASCII branch. After observing a long ASCII run, C2 compiled the loop on the assumption that `ch < 128` would remain true and guarded the other path with an uncommon trap. Reaching the final `é` invalidated that compilation, after which HotSpot generated a larger general version of the loop that scanned the preceding ASCII characters much more slowly.

That's very interesting! Thank you for digging into this and finding that.

eaftan:
> That's very interesting! Thank you for digging into this and finding that.

Yeah, this was an interesting one -- one of those things that makes this project fun.  :) 

I finished the larger benchmark run against current main. Time is normalized as `PR/main`, so lower is better:

| Workload | Representation | 1 KiB | 32 KiB | 1 MiB |
|---|---|---:|---:|---:|
| `alphabetic.absent` | String | 0.298 | 0.319 | 0.302 |
| `alphabetic.late` | String | 0.358 | 0.337 | 0.310 |
| `cjk.absent` | String | 1.006 | 0.999 | 1.003 |
| `cjk.late` | String | 0.567 | 0.501 | 0.531 |
| `alphabetic.absent` | UTF-8 | 0.240 | 0.356 | 0.347 |
| `alphabetic.late` | UTF-8 | 0.697 | 0.682 | 0.658 |
| `cjk.absent` | UTF-8 | 1.000 | 0.998 | 1.006 |
| `cjk.late` | UTF-8 | 0.114 | 0.064 | 0.060 |

The scanner split retains the original optimization's effectiveness across the complete workload family. Alphabetic searches are 1.43–4.17x faster, late CJK searches are 1.76–16.6x faster, and rejected CJK inputs are neutral. Long confirmation of the 1 MiB rows gives the same conclusion.

The .NET exclusion led me to check the other cross-runtime profiles. C++ RE2, Go, and PCRE2 cannot express Java's `IsAlphabetic` property exactly, and the CJK workloads need engine-specific code-point escapes and an explicit ASCII digit class. I've pushed a commit that adds the appropriate exclusions or exact alternates for C++ RE2, Go, Rust, PCRE2, and .NET, along with regression coverage for the resolved execution plan.

LGTM

eaftan (approved):

Issue #700, opened 2026-08-12
Multi-Literal Start Acceleration: Vector/SWAR (Teddy) vs. Scalar Aho-Corasick


cushon:
This issue discusses acceleration strategies for multi-literal alternations (e.g. `(?:AL|AK|AZ|...|WY)` and `(?:ERROR|WARN|INFO|FATAL)`). 

A prototype of Aho-Corasick in https://github.com/eaftan/safere/pull/540 showed that a scalar Aho-Corasick implementation in Java isn't faster than the DFA, and causes regressions compared to SafeRE's existing character-class accelerators.

* Stepping a byte-by-byte state machine in Java bytecode caps out at $\sim 0.4\text{--}0.5\text{ GB/s}$ ($2.0\text{--}2.5\text{ ns/char}$).
* For alternations with distinct leading characters, SafeRE's 128-bit `AsciiBitmap` skips non-matching text at **$1.5\text{--}3.0\text{ GB/s}$**, outperforming scalar DFA state transitions by $3\text{--}4\times$.
* In unanchored regex searches, Aho-Corasick must track leftmost start positions and match lengths across failure links, adding per-character memory loads (`matchIndices[state]`) and branch checks that make it slower than SafeRE's core DFA loop.

Instead, I think we should investigate Vector / SWAR multi-substring acceleration (the Teddy algorithm), and using ASCII bitmaps + DFA for scalar execution, similar to engines like Hyperscan and Rust's `regex-automata`. This also came up in https://github.com/eaftan/safere/issues/594#issuecomment-5093195645.

Proposal:

* SIMD Vector Shuffle
   - Split 1–3 byte pattern prefixes into 16-byte nibble lookup tables.
   - Use `ByteVector.rearrange(VectorShuffle)` / `selectFrom()` in `jdk.incubator.vector` to evaluate 16, 32, or 64 input bytes simultaneously
* Candidate Generation Only
   - Teddy should function strictly as a candidate-generating start accelerator (`StringStartAccelerator` / `Utf8StartAccelerator`).
   - When a vector mask match occurs at index $i$, verify candidates using direct prefix equality checks (`text.startsWith()`) or SafeRE's forward DFA pass, rather than maintaining a separate scalar Aho-Corasick state machine.
* Execution Gating
   - Vector API / SWAR available: use `VectorTeddyStartAccelerator` for $2 \le N \le 32$ literal patterns.
[...]

PR #736, opened 2026-08-19
Add Teddy SIMD multi-literal start accelerator for UTF-8 byte arrays

cushon:
See #700

This adds multi-literal start acceleration (of patterns like `foo|bar|baz|...`) using Hyperscan's Teddy algorithm.

This also drops `IncubatorVectorScanProvider.MINIMUM_INPUT_LENGTH` from 1024 down to 64, the conclusion I came to in #656 was that the setup overhead for the JDK Vector APIs is low enough that's a good threshold.

I experimented with some variations of this. Hyperscan's Teddy3 calculates 3 byte offsets unconditionally on every 16/32-byte block, which hotspot didn't seem to optimize well. This landed on a tiered approach with a 2-byte primary filter and a 3-byte confirmation filter.

Similar to existing vector accelerators this only handles `byte[]`, but could be generalized to `String` if it becomes possible to get a zero-copy vector view of a string's UTF-16 data (#656). When generalizing this to UTF-16, the implementation here can be used as-is for latin1. For non-latin1 UTF-16, it can use a SIMD Narrowing / Pack (load 32 bytes $\to$ pack 16 ASCII chars into 16 bytes $\to$ run 8-bit Teddy kernel).

```
./safere-benchmarks/scripts/compare-branch.sh   --baseline vector-benchmark-baseline   --vector 'SearchScalingBenchmark\.alternationScaled.*@safere-utf8'
```

| SearchScalingBenchmark.alternationScaled.1024    |        2217 ± 74 |        471 ± 33 |   4.70x |
| SearchScalingBenchmark.alternationScaled.10240   |      20159 ± 107 |       3914 ± 45 |   5.15x |
| SearchScalingBenchmark.alternationScaled.102400  |    197098 ± 1446 |     32405 ± 711 |   6.08x |
| SearchScalingBenchmark.alternationScaled.1048576 |  2028547 ± 18363 | 839162 ± 820365 |   2.42x |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -737,6 +737,8 @@
       "id": "searchScaling.prose.{size}",
       "axes": {
         "size": [
+          64,
+          256,
           1024,
           10240,
           102400,
@@ -3961,6 +3963,42 @@
       "measurement": {
         "mode": "averageTime",
         "timingUnit": "microseconds",
+        "constraints": [
+          "prematerializedInput"
+        ]
+      }
+    },
+    {
+      "id": "SearchScalingBenchmark.alternationScaled.64",
+      "operation": "find",
+      "patterns": [
+        "foo|bar|baz|qux|quux|corge|grault|garply"
+      ],
+      "inputs": [
+        "searchScaling.prose.64"
+      ],
+      "resultConsumption": "boolean",
+      "measurement": {
+        "mode": "averageTime",
+        "timingUnit": "nanoseconds",
+        "constraints": [
+          "prematerializedInput"
+        ]
+      }
+    },
+    {
+      "id": "SearchScalingBenchmark.alternationScaled.256",
+      "operation": "find",
+      "patterns": [
+        "foo|bar|baz|qux|quux|corge|grault|garply"
+      ],
+      "inputs": [
+        "searchScaling.prose.256"
+      ],
+      "resultConsumption": "boolean",
+      "measurement": {
+        "mode": "averageTime",
+        "timingUnit": "nanoseconds",
         "constraints": [
           "prematerializedInput"
         ]
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -3967,6 +3967,24 @@
       }
     },
     {
+      "id": "SearchScalingBenchmark.alternationScaled.1024",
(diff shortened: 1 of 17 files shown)


eaftan:
> Similar to existing vector accelerators this only handles byte[], but could be generalized to String if it becomes possible to get a zero-copy vector view of a string's UTF-16 data (https://github.com/eaftan/safere/pull/656). When generalizing this to UTF-16, the implementation here can be used as-is for latin1. For non-latin1 UTF-16, it can use a SIMD Narrowing / Pack (load 32 bytes → pack 16 ASCII chars into 16 bytes → run 8-bit Teddy kernel).

Side note, but I wonder if we should keep a list of these opportunities for if/when we get better support from the upstream String class.  Maybe we should file issues and use a label to track them?

eaftan (dismissed):
This is a nice use of Teddy for rejected UTF-8 alternations. I reproduced the large-input gains and measured the previously untested short side of the cutoff.

I found two related cutoff issues. The submitted 64-byte Teddy cutoff made this workload 4.45x slower at 64 bytes and 4.40x slower at 256 bytes. Lowering the shared Vector cutoff also forced the existing character-class scanners onto short inputs where Vector was 1.59–4.68x slower than the existing SWAR path. Patterns also eagerly built Teddy's repeated lookup tables when the Vector provider was disabled, even though those tables could never be used.

I've pushed commits that avoid Teddy setup unless the optional Vector provider is active, restore the existing scanners' 1 KiB cutoff, and give Teddy its own 1 KiB cutoff (the first measured winning size). Below that point the accelerator now uses the normal matcher path. I also added 64- and 256-byte workloads so both sides of the crossover remain covered.

After those changes, I compared the corrected PR with current main. Time is normalized as `(PR + fixes)/main`, so lower is better:

| Input size | PR + fixes/main | Result |
|---:|---:|---|
| 64 B | 0.988 | neutral; normal matcher path |
| 256 B | 0.973 | neutral; normal matcher path |
| 1 KiB | 0.251 | 3.98x faster |
| 10 KiB | 0.180 | 5.55x faster |
| 100 KiB | 0.109 | 9.17x faster |
| 1 MiB | 0.451 | 2.22x faster, long confirmation |

With the fixes, the short inputs are neutral and the claimed gains from 1 KiB upward remain. LGTM

eaftan (approved):

PR #737, opened 2026-08-19
Add multi-broadcast start accelerator for small literal alternations on UTF-8

cushon:
This adds vectorized start acceleration for 2–4 literal alternations (e.g. `foo|bar|baz|qux`) on UTF-8 `byte[]` inputs using the JDK Vector API.

This is complementary to Teddy (https://github.com/eaftan/safere/pull/736), which targets larger keyword dictionaries, as direct multi-broadcast avoids nibble LUT setup and outperforms it for small alternations.
 
This currently applies to the `byte[]` APIs, it can be generalized to `String` in the future (#656).

For small alternations (2 to 4 keywords), compiling full DFA states or nibble LUTs carries unnecessary constant-factor overhead. This PR adds a multi-broadcast vector filter:
* Uses `RarityOracle.rarestAsciiOffset()` to select the least frequent ASCII character for each literal, keeping false-positive candidate density low (< 5%).
* Broadcasts anchor bytes into `ByteVector` registers (`v0`..`v3`), compares 32-byte chunks in parallel, and combines match masks using `.or()`.
* Extracts matching lane offsets and confirms candidate matches directly via `Ascii.regionMatches()`.
* Restricts the accelerator to 2–4 literals to keep register pressure within AVX2 limits (leaving 10+ YMM registers for loop unrolling) and falls back cleanly to character class / DFA scanning for >= 5 literals.

```
./safere-benchmarks/scripts/compare-branch.sh   --baseline benchmarks-baseline   --vector   'SearchScalingBenchmark\.multiLiteral.*@safere-utf8'
```

| Benchmark                                                | baseline (ns/op) | current (ns/op) | Speedup |
| -------------------------------------------------------- | ---------------- | --------------- | ------- |
| SearchScalingBenchmark.multiLiteral3Fail.1024            |        101 ± 5.7 |        38 ± 3.0 |   2.63x |
| SearchScalingBenchmark.multiLiteral3Success.1024         |        141 ± 1.8 |       292 ± 4.7 |   0.48x |
| SearchScalingBenchmark.multiLiteral6Fail.1024            |        174 ± 9.4 |       168 ± 3.0 |   1.03x |
| SearchScalingBenchmark.multiLiteral6Success.1024         |        164 ± 5.5 |       161 ± 1.8 |   1.02x |
| SearchScalingBenchmark.multiLiteral3Fail.10240           |         319 ± 24 |       212 ± 3.8 |   1.50x |
| SearchScalingBenchmark.multiLiteral3Success.10240        |         319 ± 17 |       458 ± 6.5 |   0.70x |
| SearchScalingBenchmark.multiLiteral6Fail.10240           |         537 ± 27 |       521 ± 4.8 |   1.03x |
| SearchScalingBenchmark.multiLiteral6Success.10240        |         514 ± 15 |        519 ± 51 |   0.99x |
| SearchScalingBenchmark.multiLiteral3Fail.102400          |        2440 ± 29 |      1996 ± 238 |   1.22x |
| SearchScalingBenchmark.multiLiteral3Success.102400       |       2152 ± 166 |       2224 ± 61 |   0.97x |
| SearchScalingBenchmark.multiLiteral6Fail.102400          |       4350 ± 256 |       4206 ± 23 |   1.03x |
| SearchScalingBenchmark.multiLiteral6Success.102400       |        3798 ± 48 |       3796 ± 82 |   1.00x |
| SearchScalingBenchmark.multiLiteral3Fail.1048576         |      24866 ± 416 |    29809 ± 1594 |   0.83x |
| SearchScalingBenchmark.multiLiteral3Success.1048576      |     21794 ± 1134 |    30564 ± 1010 |   0.71x |
| SearchScalingBenchmark.multiLiteral6Fail.1048576         |     41781 ± 1786 |    41194 ± 1853 |   1.01x |
| SearchScalingBenchmark.multiLiteral6Success.1048576      |     43284 ± 1171 |     43150 ± 640 |   1.00x |
| SearchScalingBenchmark.multiLiteralProse3Fail.1024       |        2198 ± 88 |        641 ± 13 |   3.43x |
| SearchScalingBenchmark.multiLiteralProse3Success.1024    |        425 ± 5.3 |       222 ± 5.1 |   1.91x |
| SearchScalingBenchmark.multiLiteralProse6Fail.1024       |        2156 ± 32 |       2156 ± 34 |   1.00x |
| SearchScalingBenchmark.multiLiteralProse6Success.1024    |         432 ± 17 |        433 ± 13 |   1.00x |
| SearchScalingBenchmark.multiLiteralProse3Fail.10240      |      21017 ± 642 |       6097 ± 47 |   3.45x |
| SearchScalingBenchmark.multiLiteralProse3Success.10240   |         431 ± 15 |       221 ± 5.7 |   1.95x |
| SearchScalingBenchmark.multiLiteralProse6Fail.10240      |      20382 ± 104 |      20166 ± 70 |   1.01x |
| SearchScalingBenchmark.multiLiteralProse6Success.10240   |        426 ± 5.9 |        452 ± 25 |   0.94x |
| SearchScalingBenchmark.multiLiteralProse3Fail.102400     |    199067 ± 5435 |    63659 ± 5538 |   3.13x |
| SearchScalingBenchmark.multiLiteralProse3Success.102400  |         433 ± 14 |       221 ± 5.6 |   1.96x |
| SearchScalingBenchmark.multiLiteralProse6Fail.102400     |    200171 ± 9515 |   198867 ± 2857 |   1.01x |
| SearchScalingBenchmark.multiLiteralProse6Success.102400  |         433 ± 14 |       427 ± 9.0 |   1.01x |
| SearchScalingBenchmark.multiLiteralProse3Fail.1048576    |  2074146 ± 69102 |  699746 ± 29495 |   2.96x |
| SearchScalingBenchmark.multiLiteralProse3Success.1048576 |        423 ± 3.7 |       220 ± 7.1 |   1.92x |
| SearchScalingBenchmark.multiLiteralProse6Fail.1048576    |  2023769 ± 14776 | 2018426 ± 11453 |   1.00x |
| SearchScalingBenchmark.multiLiteralProse6Success.1048576 |        427 ± 3.5 |       426 ± 9.5 |   1.00x |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -452,7 +452,7 @@
       "id": "vectorScan.multi.nonascii.32768",
       "recipe": {
         "kind": "repeatToLength",
-        "unit": "é",
+        "unit": "\u00e9",
         "length": 32768
       },
       "shared": true
@@ -1725,7 +1725,7 @@
       },
       "recipe": {
         "kind": "delimitedRepeatToLength",
-        "value": "いくつもの　中国語と日本語の文字　がここにあります　そして　いくつかのスペース　があります",
+        "value": "\u3044\u304f\u3064\u3082\u306e\u3000\u4e2d\u56fd\u8a9e\u3068\u65e5\u672c\u8a9e\u306e\u6587\u5b57\u3000\u304c\u3053\u3053\u306b\u3042\u308a\u307e\u3059\u3000\u305d\u3057\u3066\u3000\u3044\u304f\u3064\u304b\u306e\u30b9\u30da\u30fc\u30b9\u3000\u304c\u3042\u308a\u307e\u3059",
         "delimiterAlphabet": " ",
         "seed": 42,
         "length": "{size}"
@@ -1743,7 +1743,7 @@
       },
       "recipe": {
         "kind": "delimitedRepeatToLength",
-        "value": "いくつもの中国語と日本語の文字がここにありますそしていくつかのスペースがあります",
+        "value": "\u3044\u304f\u3064\u3082\u306e\u4e2d\u56fd\u8a9e\u3068\u65e5\u672c\u8a9e\u306e\u6587\u5b57\u304c\u3053\u3053\u306b\u3042\u308a\u307e\u3059\u305d\u3057\u3066\u3044\u304f\u3064\u304b\u306e\u30b9\u30da\u30fc\u30b9\u304c\u3042\u308a\u307e\u3059",
         "delimiterAlphabet": " ",
         "seed": 42,
         "length": "{size}"
@@ -1761,7 +1761,7 @@
       },
       "recipe": {
         "kind": "delimitedRepeatToLength",
-        "value": "这是一个中文测试句子",
+        "value": "\u8fd9\u662f\u4e00\u4e2a\u4e2d\u6587\u6d4b\u8bd5\u53e5\u5b50",
         "delimiterAlphabet": " ",
         "seed": 42,
         "length": "{size}"
@@ -1797,7 +1797,7 @@
       },
       "recipe": {
         "kind": "delimitedRepeatToLength",
-        "value": "Today is a beautiful day 😇 with warm sunshine.",
+        "value": "Today is a beautiful day \ud83d\ude07 with warm sunshine.",
         "delimiterAlphabet": " ",
         "seed": 42,
         "length": "{size}"
@@ -2109,21 +2109,21 @@
       "id": "utf8.captureFree.multibyteEarly",
       "recipe": {
         "kind": "literal",
-        "text": "错误：请求失败"
+        "text": "\u9519\u8bef\uff1a\u8bf7\u6c42\u5931\u8d25"
       }
     },
     {
       "id": "utf8.captureFree.multibyteLate",
       "recipe": {
         "kind": "literal",
-        "text": "请求处理结束：错误"
+        "text": "\u8bf7\u6c42\u5904\u7406\u7ed3\u675f\uff1a\u9519\u8bef"
       }
     },
     {
       "id": "utf8.captureFree.multibyteNoMatch",
       "recipe": {
         "kind": "literal",
-        "text": "请求处理成功"
+        "text": "\u8bf7\u6c42\u5904\u7406\u6210\u529f"
       }
     },
     {
@@ -2137,7 +2137,7 @@
       "id": "utf8.repeatedFind.multibyte",
       "recipe": {
         "kind": "literal",
-        "text": "one café 東京 naïve"
+        "text": "one caf\u00e9 \u6771\u4eac na\u00efve"
       }
     },
     {
@@ -2151,7 +2151,7 @@
       "id": "utf8.captureBounds.named",
       "recipe": {
         "kind": "literal",
-        "text": "city=東京&name=Renée"
+        "text": "city=\u6771\u4eac&name=Ren\u00e9e"
       }
     },
     {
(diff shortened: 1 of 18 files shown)


eaftan (commented):
This approach seems reasonable and shows good improvements on most workloads. I like the idea of using the lower-overhead multi-broadcast scanner for two-to-four literals and Teddy for larger dictionaries.

I found one issue with the submitted cutoff. Lowering the provider-wide threshold to 64 bytes also forced the existing character-class scanners onto short inputs where Vector is 1.59–4.68x slower than their SWAR path. I've pushed a commit that restores their 1 KiB cutoff, gives multi-broadcast an independent 64-byte cutoff, and avoids building multi-literal metadata when Vector mode is disabled. I added 64- and 256-byte workloads as well; this accelerator does win there, by 1.66x and 2.02x respectively.

However, I measured material regressions on several of the random-input workloads. Time is normalized as `(PR + fixes)/main`, so lower is better:

| Workload | Input size | PR + fixes/main | Regression |
|---|---:|---:|---:|
| `multiLiteral3Success` | 1 KiB | 1.722 | 72% slower |
| `multiLiteral3Success` | 10 KiB | 1.389 | 39% slower |
| `multiLiteral3Success` | 1 MiB | 1.114 | 11% slower |
| `multiLiteral3Fail` | 100 KiB | 1.151 | 15% slower |
| `multiLiteral3Fail` | 1 MiB | 1.151 | 15% slower |

I haven't dug into the causes yet, but I'd like to understand these regressions before deciding whether to merge. If you have time to investigate during your day, that would be great. On my end, I'll dispatch an agent overnight and see what it finds.

cushon:
Thanks, good catch on those regressions.

I think the issue was a missing `WorkLimit` budget, and that the accelerator was being used for 1-byte anchors even if the literals shared prefix characters which was slower than `Teddy`. I've added a `WorkLimit` budget, and a check for distinct prefixes with fallthrough to Teddy.

```
./safere-benchmarks/scripts/compare-branch.sh   --baseline benchmarks-baseline   --vector   'SearchScalingBenchmark\.multiLiteral3(Success|Fail).*@safere-utf8'
```

| Benchmark                                           | baseline (ns/op) | current (ns/op) | Speedup |
| --------------------------------------------------- | ---------------- | --------------- | ------- |
| SearchScalingBenchmark.multiLiteral3Fail.64         |        159 ± 1.4 |       19 ± 0.10 |   8.52x |
| SearchScalingBenchmark.multiLiteral3Fail.256        |        539 ± 2.6 |       21 ± 0.13 |   25.7x |
| SearchScalingBenchmark.multiLiteral3Fail.1024       |        248 ± 1.5 |       37 ± 0.23 |   6.78x |
| SearchScalingBenchmark.multiLiteral3Success.1024    |        533 ± 4.9 |       282 ± 6.4 |   1.89x |
| SearchScalingBenchmark.multiLiteral3Fail.10240      |        2318 ± 13 |       209 ± 3.2 |   11.1x |
| SearchScalingBenchmark.multiLiteral3Success.10240   |        2730 ± 36 |       445 ± 5.4 |   6.13x |
| SearchScalingBenchmark.multiLiteral3Fail.102400     |      23147 ± 135 |      1968 ± 5.6 |   11.8x |
| SearchScalingBenchmark.multiLiteral3Success.102400  |      25132 ± 213 |       2201 ± 70 |   11.4x |
| SearchScalingBenchmark.multiLiteral3Fail.1048576    | 1034590 ± 256739 |     28534 ± 684 |   36.3x |
| SearchScalingBenchmark.multiLiteral3Success.1048576 | 1007673 ± 300097 |     31650 ± 956 |   31.8x |

eaftan (dismissed):
Thanks for following up on the regressions from my previous comment. The WorkLimit and the fallback to Teddy when alternatives share their first byte address the high false-candidate case you identified, and I reproduce the large gains on the uppercase random-input workloads.

While validating the revised implementation against current main, I dug further into the previously slow random-input case. For `APPLE|BANANA|CHERRY`, current main compacts the possible first bytes into the range `A-C`, requiring two Vector comparisons and one mask intersection. The multi-broadcast loop instead performed three equality comparisons and combined their masks. I've pushed a commit that gives the multi-literal descriptor the same compact anchor ranges and reuses the existing range-mask loop.

Your distinct-prefix check correctly sends alternatives that share their first byte to Teddy. I found a separate case involving distinct but common anchors: the existing `multiLiteralProse3` workloads use common `b`, `s`, `t`, and `g` anchors, so selecting multi-broadcast unconditionally still produced many candidates. I've pushed a commit that compiles both scanners for eligible two-to-four-literal patterns, begins with multi-broadcast, and observes candidate density over at most 256 bytes. When estimated literal-verification work reaches 50% of the bytes scanned, it switches to Teddy at the current unprocessed block. The rule lives in a small `MultiLiteralSelectionPolicy` class with named constants, and the transition neither rescans the processed prefix nor changes leftmost-match behavior.

While checking that fallback, I found one linear-time issue in my adaptive implementation: if the observed prefix was sparse but a later suffix became dense, exhausting the WorkLimit restarted an unbounded scalar literal scan from the beginning. I've pushed a commit that returns control to the normal matcher instead, adds scaling coverage for the changing-density case, and corrects the randomized accelerator coverage.

I compared the resulting hybrid with current main using the long configuration. Time is normalized as `(PR + fixes)/main`, so lower is better:

| Workload | 1 KiB | 100 KiB | Result |
|---|---:|---:|---|
| `multiLiteral3Fail` | 0.313 | 0.159 | 3.19–6.27x faster |
| `multiLiteral3Success` | 0.496 | 0.199 | 2.01–5.04x faster |
| `multiLiteralProse3Fail` | 0.990 | 0.929 | neutral to 8% faster |
| `multiLiteralProse3Success` | 1.114 | 1.089 | 9–11% slower |

The prose-success workload finds its match near the beginning, so its 9–11% loss is a bounded setup cost rather than a scaling regression. I think that is a reasonable tradeoff for the 2.0–6.3x sparse-input gains and the neutral-to-faster dense full scan.

LGTM

eaftan (approved):

PR #738, opened 2026-08-19
Vectorize ASCII pair and triple scanning on UTF-8 inputs

cushon:
This adds 256/512-bit SIMD acceleration for 2-character and 3-character searches (`indexOfAsciiPair` and `indexOfAsciiTriple`) on UTF-8 `byte[]` inputs using the JDK Vector API.

`Utf8InputScanner` previously used 64-bit SWAR (`ByteSwarScan`) to scan 8 bytes per iteration for byte pairs (`[ab]`) and triples (`[abc]`). On modern AVX2 and AVX-512 hardware, vector registers can inspect 32 or 64 bytes per cycle.

* Add `indexOfAsciiPair` and `indexOfAsciiTriple` in `ByteVectorScan` using broadcast `ByteVector` comparisons and `.or()` mask reduction.
* Delegate to `scanProvider` in `Utf8InputScanner` when the scan length meets the vector threshold (`>= 64` bytes), with clean fallback to `ByteSwarScan`.
* Provide full unit test coverage for memory alignments (0–63 byte offsets), edge cases, and crosscheck equivalence with SWAR and scalar matching.

```
  --baseline benchmarks-baseline \
  --vector \
  'SearchScalingBenchmark\.ascii(Pair|Triple).*@safere-utf8'
```

| Benchmark                                         | baseline (ns/op) | current (ns/op) | Speedup |
| ------------------------------------------------- | ---------------- | --------------- | ------- |
| SearchScalingBenchmark.asciiPairFail.1024         |        86 ± 0.60 |       28 ± 0.15 |   3.11x |
| SearchScalingBenchmark.asciiPairSuccess.1024      |         96 ± 3.8 |        59 ± 2.4 |   1.62x |
| SearchScalingBenchmark.asciiTripleFail.1024       |        96 ± 0.85 |        34 ± 2.0 |   2.83x |
| SearchScalingBenchmark.asciiTripleSuccess.1024    |        101 ± 2.3 |        68 ± 2.8 |   1.48x |
| SearchScalingBenchmark.asciiPairFail.10240        |        217 ± 2.3 |        166 ± 12 |   1.31x |
| SearchScalingBenchmark.asciiPairSuccess.10240     |         229 ± 10 |        194 ± 21 |   1.18x |
| SearchScalingBenchmark.asciiTripleFail.10240      |        303 ± 2.7 |        213 ± 14 |   1.42x |
| SearchScalingBenchmark.asciiTripleSuccess.10240   |        314 ± 5.8 |       234 ± 2.7 |   1.34x |
| SearchScalingBenchmark.asciiPairFail.102400       |        1872 ± 12 |      1669 ± 210 |   1.12x |
| SearchScalingBenchmark.asciiPairSuccess.102400    |        1932 ± 52 |      1887 ± 104 |   1.02x |
| SearchScalingBenchmark.asciiTripleFail.102400     |       2518 ± 140 |       1989 ± 30 |   1.27x |
| SearchScalingBenchmark.asciiTripleSuccess.102400  |       2520 ± 127 |      1965 ± 108 |   1.28x |
| SearchScalingBenchmark.asciiPairFail.1048576      |      19710 ± 591 |     20750 ± 787 |   0.95x |
| SearchScalingBenchmark.asciiPairSuccess.1048576   |      19572 ± 755 |     19960 ± 707 |   0.98x |
| SearchScalingBenchmark.asciiTripleFail.1048576    |     25603 ± 1032 |     21783 ± 927 |   1.18x |
| SearchScalingBenchmark.asciiTripleSuccess.1048576 |     26377 ± 1861 |     21802 ± 779 |   1.21x |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -452,7 +452,7 @@
       "id": "vectorScan.multi.nonascii.32768",
       "recipe": {
         "kind": "repeatToLength",
-        "unit": "é",
+        "unit": "\u00e9",
         "length": 32768
       },
       "shared": true
@@ -1725,7 +1725,7 @@
       },
       "recipe": {
         "kind": "delimitedRepeatToLength",
-        "value": "いくつもの　中国語と日本語の文字　がここにあります　そして　いくつかのスペース　があります",
+        "value": "\u3044\u304f\u3064\u3082\u306e\u3000\u4e2d\u56fd\u8a9e\u3068\u65e5\u672c\u8a9e\u306e\u6587\u5b57\u3000\u304c\u3053\u3053\u306b\u3042\u308a\u307e\u3059\u3000\u305d\u3057\u3066\u3000\u3044\u304f\u3064\u304b\u306e\u30b9\u30da\u30fc\u30b9\u3000\u304c\u3042\u308a\u307e\u3059",
         "delimiterAlphabet": " ",
         "seed": 42,
         "length": "{size}"
@@ -1743,7 +1743,7 @@
       },
       "recipe": {
         "kind": "delimitedRepeatToLength",
-        "value": "いくつもの中国語と日本語の文字がここにありますそしていくつかのスペースがあります",
+        "value": "\u3044\u304f\u3064\u3082\u306e\u4e2d\u56fd\u8a9e\u3068\u65e5\u672c\u8a9e\u306e\u6587\u5b57\u304c\u3053\u3053\u306b\u3042\u308a\u307e\u3059\u305d\u3057\u3066\u3044\u304f\u3064\u304b\u306e\u30b9\u30da\u30fc\u30b9\u304c\u3042\u308a\u307e\u3059",
         "delimiterAlphabet": " ",
         "seed": 42,
         "length": "{size}"
@@ -1761,7 +1761,7 @@
       },
       "recipe": {
         "kind": "delimitedRepeatToLength",
-        "value": "这是一个中文测试句子",
+        "value": "\u8fd9\u662f\u4e00\u4e2a\u4e2d\u6587\u6d4b\u8bd5\u53e5\u5b50",
         "delimiterAlphabet": " ",
         "seed": 42,
         "length": "{size}"
@@ -1797,7 +1797,7 @@
       },
       "recipe": {
         "kind": "delimitedRepeatToLength",
-        "value": "Today is a beautiful day 😇 with warm sunshine.",
+        "value": "Today is a beautiful day \ud83d\ude07 with warm sunshine.",
         "delimiterAlphabet": " ",
         "seed": 42,
         "length": "{size}"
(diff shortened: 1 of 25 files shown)


eaftan (dismissed):
The pair/triple Vector kernels look correct.  I like that the tests compare their results with the existing SWAR scanners across randomized inputs, starting positions, SIMD block edges, and explicit search limits.

I found that the submitted 64-byte provider-wide cutoff also moved unrelated character-class scans onto Vector where they can be 1.59–4.68x slower than SWAR. I also found a separate crossover in this PR's own kernels: long confirmation showed the triple kernel becoming 12–21% slower at 100 KiB and 1 MiB.

I've pushed a commit that restores the existing scanner cutoff and gives these kernels independent policies based on the measured results. Pair scanning uses Vector from 64 bytes upward. Triple scanning uses Vector from 64 through 10,240 bytes, then retains the existing character-class accelerator. Provider selection now uses the actual remaining bounded scan length rather than the total containing input.

Time is normalized as `PR/main`, so lower is better:

| Workload | 1 KiB | 10 KiB | 100 KiB | 1 MiB |
|---|---:|---:|---:|---:|
| `asciiPairFail` | 0.192 | 0.513 | 0.911 | 0.942 |
| `asciiPairSuccess` | 0.641 | 0.752 | 0.955 | 0.984 |
| `asciiTripleFail` | 0.215 | 0.722 | 1.008 | 1.013 |
| `asciiTripleSuccess` | 0.698 | 0.978 | 0.981 | 1.015 |

This retains the useful 1 KiB and 10 KiB gains without the large-input regression.

LGTM.

cushon:
Thanks for the review, and for the tuning and improvements to the cutoffs!

I resolved merge conflicts, and also did one more refactor to extract `CharClassScanInfo` into a top level class to consolidate char class handling a bit more and set up to generalize this to `String` later.

eaftan (approved):
Thanks for the `CharClassScanInfo` consolidation. I found a few issues while checking the revised implementation. `AsciiSmallSet` and `AsciiBitmapClass` rebuilt their range arrays on every access, which turned precomputed pattern metadata into per-match allocation. The submitted triple benchmark also used `[XYZ]`; because those characters form the contiguous range `X-Z`, it exercised the existing range scanner rather than the new three-character equality kernel. Finally, three-character start acceleration called the dedicated API directly, so scans above its measured cutoff fell back to the slower SWAR triple loop instead of returning to the general Vector character-class scanner. The unknown-provider diagnostic also omitted the requested value.

I've pushed commits that retain the precomputed arrays, correct the diagnostic, change the triple workload to the disjoint set `[XZ_]`, and route three-character start acceleration through the shared character-class dispatcher. That preserves the dedicated triple kernel through its measured cutoff and the existing general scanner above it.

I compared the corrected PR with current main using the declared pair/triple workloads. Time is normalized as `(PR + fixes)/main`, so lower is better:

| Workload | 1 KiB | 10 KiB | 100 KiB | 1 MiB |
|---|---:|---:|---:|---:|
| `asciiPairFail` | 0.185 | 0.505 | 0.916 | 1.000 |
| `asciiPairSuccess` | 0.616 | 0.746 | 0.885 | 1.000 |
| `asciiTripleFail` | 0.167 | 0.492 | 1.015 | 0.997 |
| `asciiTripleSuccess` | 0.547 | 1.011 | 0.988 | 0.990 |

The pair path improves 1.09–5.41x through 100 KiB and is neutral at 1 MiB. The disjoint-triple path improves 1.82–5.97x at 1 KiB and 2.03x on 10 KiB rejection; the other measured points are neutral. LGTM.

PR #739, opened 2026-08-19
Add SafeRE quality gates for vector species, repeated find scaling, and accelerator policy introspection

cushon:
This is another round of additional test coverage inspired by bugs caught in recent PR reviews, similar to https://github.com/eaftan/safere/commit/6f165b3514a404a4c17b32f758a647fa333e8acd

* displaced start-anchored constant work assertions for String and UTF-8
* repeated `find()` scaling linearity assertions across alternating case and multi-branch alternations
* explicit start accelerator and policy assertions on compiled `Pattern`s
* exercise `ByteVectorScan` and `ShortVectorScan` across all vector shapes

### DIFF
--- a/safere-crosscheck/pom.xml
+++ b/safere-crosscheck/pom.xml
@@ -147,6 +147,7 @@
                         <exclude name="Utf8InputTest.java"/>
                         <exclude name="Utf8LinearTimeTest.java"/>
                         <exclude name="UtilsTest.java"/>
+                        <exclude name="VectorSpeciesSimulationTest.java"/>
                         <exclude name="WalkerTest.java"/>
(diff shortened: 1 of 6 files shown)


eaftan (approved):
Thanks for adding these quality gates!  These look like useful things to check.

I did find that the Vector species test could silently pass without exercising any species: reflective lookup errors were converted to empty arrays or null methods, and each test then returned. The species were also stored in untyped `Object[]` fields. I've pushed a commit that treats only an unavailable Vector module as an explicit skipped-test assumption, makes field/method lookup failures fail the test, and stores the reflected species in a typed handle list.

With that fix, LGTM.

PR #740, opened 2026-08-20
Unify StringStartAccelerator structure with Utf8StartAccelerator and use sealed records

cushon:
This makes StringStartAccelerator and Utf8StartAccelerator more consistent.

Matching loops now route through `StringStartAccelerator.findNextCandidate` instead of doing virtual dispatch.

### DIFF
--- a/safere/src/main/java/org/safere/Matcher.java
+++ b/safere/src/main/java/org/safere/Matcher.java
@@ -1683,7 +1683,9 @@ public final class Matcher implements MatchResult {
           if (strategy != null) {
             diagnosticParticipation(strategy, StrategyRole.START_ACCELERATION);
           }
-          int idx = accelerator.findCandidate(text, searchFrom, prog.unixLines());
+          int idx =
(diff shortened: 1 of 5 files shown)


eaftan (dismissed):
This refactoring looks good. Splitting case-insensitive String literals into their own record makes the String and UTF-8 structures consistent, and the explicit sealed-type switches keep the complete accelerator dispatch visible in one place.

I compared a representative set of hot candidate-search paths with current main. Time is normalized as `PR/main`, so lower is better:

| Workload | Representation | PR/main |
|---|---|---:|
| `caseInsensitiveKeywordFind.match.100000` | String | 1.001 |
| `caseInsensitiveKeywordFind.noMatch.100000` | String | 1.002 |
| `literalPrefixNonmatchBytes` | UTF-8 | 1.017 |
| `findPair.longAbsent` | UTF-8 | 1.008 |

Those results are neutral. LGTM.

eaftan (dismissed):

eaftan (approved):

PR #744, opened 2026-08-21
Leading nullable and bounded prefix expansion for start acceleration

cushon:
This adds support for start acceleration of patterns with leading quantified whitespace or character classes followed by a token that can be accelerated. The engine jumps forward to candidate tokens using fast paths and scans backward over the local leading repetition, avoiding scalar character-by-character DFA evaluation across non-matching text.

This was inspired by a real-world pattern like `\s*[\[［]\s*(?:(?:[0-9]+\.?){3,4}...)\s*[\]］]` where SafeRE performance lags RE2 for `replaceAll` operations. This was based on similar optimizations in rust-automata and hyperscan.

I think the slight `noMatch` regression for longer inputs is similar to other marginal slowdowns existing accelerators in cases where they don't pay off.

This should also dovetail with work to vectorize the fast paths, I just measured the scalar performance.

```
./safere-benchmarks/scripts/compare-branch.sh --baseline benchmark-markdown-citation 'llmCitationScrub.*@safere-string'
```

| Benchmark                       | baseline (ns/op) | current (ns/op) | Speedup |
| ------------------------------- | ---------------- | --------------- | ------- |
| llmCitationScrub.match.1000     |       4639 ± 251 |       3604 ± 88 |   1.29x |
| llmCitationScrub.match.10000    |      73976 ± 967 |     46196 ± 733 |   1.60x |
| llmCitationScrub.match.100000   |   460196 ± 24492 |  299847 ± 23447 |   1.53x |
| llmCitationScrub.noMatch.1000   |        3803 ± 72 |       3040 ± 37 |   1.25x |
| llmCitationScrub.noMatch.10000  |      35511 ± 447 |    36097 ± 1177 |   0.98x |
| llmCitationScrub.noMatch.100000 |    354613 ± 3890 |  370001 ± 10446 |   0.96x |

### DIFF
--- a/.agents/skills/safere-pr-review-scout/SKILL.md
+++ b/.agents/skills/safere-pr-review-scout/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: safere-pr-review-scout
-description: "Run a serialized background sweep of open SafeRE GitHub PRs: skip drafts, track reviewed PR head SHAs and discussion changes, assess PR intent against implementation, run the review-fix-loop skill for P2+ findings, reproduce optimization benchmark claims against current main, and write durable reports and artifacts without pushing or commenting."
+description: "Run a serialized background sweep of open SafeRE GitHub PRs: skip drafts, track reviewed PR head SHAs and discussion changes, assess PR intent and merge ordering, run the review-fix-loop skill for P2+ findings, reproduce optimization benchmark claims against current main, and write self-contained durable reports and artifacts without pushing or commenting."
 ---
 
 # SafeRE PR Review Scout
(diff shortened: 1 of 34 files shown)


eaftan (dismissed):
This is a good optimization. Finding the later token first and then scanning backward across the local leading repetition avoids running the general matcher over long stretches that cannot match.

I found two correctness issues in the submitted implementation. Tail analysis recursively processed progressively shorter concatenations before rejecting nested leading expansions, so a valid pattern with thousands of consecutive leading repetitions could overflow the stack during `Pattern.compile()`. Also, when `Matcher.find(int)` started between a supplementary character's surrogate halves, the String backward scan could decode across that boundary and return a match before the requested index. The latter conflicts with [`Matcher.find(int)`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/regex/Matcher.html#find(int)), which searches starting at the specified index.

I've pushed commits that make tail analysis statically one level deep, keep every backward-decoded code point within the requested search range, add public coverage for optional, required, and bounded repetitions, and extend the stateful find fuzzer with split-surrogate starts.

I compared the corrected branch with current main using the long configuration. Time is normalized as `(PR + fixes)/main`, so lower is better:

| `llmCitationScrub` input | 1K | 10K | 100K | Result |
|---|---:|---:|---:|---|
| Match | 0.820 | 0.757 | 0.701 | 1.22–1.43x faster |
| Reject | 0.821 | 0.987 | 0.988 | neutral to 1.22x faster |

The slight longer-input rejection regression in the PR description did not reproduce in the long run; the 10K and 100K rejected cases are neutral.

LGTM

eaftan (dismissed):

eaftan (dismissed):

eaftan (approved):

PR #745, opened 2026-08-21
Improve DFA start state acceleration

cushon:
The DFA inner loop acceleration in #655 optimized cases where the DFA returned to the start state by re-using start accelerators to fast forward, and states where almost all transitions looped back to the same state by searching for 1-3 escape characters.

This change fixes a gap where if the DFA returned to the start state, and a start accelerator wasn't available, the DFA wouldn't check if it could use the escape character fast path. This PR ensures that if the DFA returns to the start state it will use a start accelerator if one is available, and then check if the escape character fast path is available.

Additionally this fixes a missing implementation of `ByteVectorScan.indexOfByte`, previously there was only a SWAR implementation.

```
./safere-benchmarks/scripts/compare-branch.sh --baseline benchmark-wildcard-baseline --vector 'optionalIndentBracketScaled.*@safere-utf8'
```

| Benchmark                                                  | baseline (ns/op) | current (ns/op) | Speedup |
| ---------------------------------------------------------- | ---------------- | --------------- | ------- |
| SearchScalingBenchmark.optionalIndentBracketScaled.64      |        101 ± 1.3 |         87 ± 11 |   1.16x |
| SearchScalingBenchmark.optionalIndentBracketScaled.256     |        102 ± 3.2 |        81 ± 1.1 |   1.26x |
| SearchScalingBenchmark.optionalIndentBracketScaled.1024    |        105 ± 3.7 |        71 ± 1.2 |   1.48x |
| SearchScalingBenchmark.optionalIndentBracketScaled.10240   |        101 ± 1.3 |       71 ± 0.59 |   1.43x |
| SearchScalingBenchmark.optionalIndentBracketScaled.102400  |        104 ± 9.6 |        71 ± 1.9 |   1.46x |
| SearchScalingBenchmark.optionalIndentBracketScaled.1048576 |        105 ± 5.8 |        71 ± 1.1 |   1.48x |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -452,7 +452,7 @@
       "id": "vectorScan.multi.nonascii.32768",
       "recipe": {
         "kind": "repeatToLength",
-        "unit": "é",
+        "unit": "\u00e9",
         "length": 32768
       },
       "shared": true
@@ -730,6 +730,27 @@
         "kind": "appendInput",
         "input": "searchScaling.random.{size}",
         "suffix": "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+      },
+      "shared": true
+    },
+    {
+      "id": "searchScaling.wildcardBody.{size}",
+      "axes": {
+        "size": [
+          64,
+          256,
+          1024,
+          10240,
+          102400,
+          1048576
+        ]
+      },
+      "recipe": {
+        "kind": "surroundToLength",
+        "prefix": "HEADER ",
+        "unit": "abcdefghijklmnopqrstuvwxyz ",
+        "suffix": " FOOTER",
+        "length": "{size}"
       },
       "shared": true
     },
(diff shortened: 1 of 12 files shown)


eaftan (dismissed):
This closes a useful gap in the DFA acceleration added in #655. When the DFA returns to its start state, it now uses a pattern start accelerator when one exists and still gets the automaton-derived escape scan otherwise. The single-byte UTF-8 Vector implementation also preserves the requested scan range.

I found one benchmark-data issue: the submitted `optionalIndentBracketScaled` family actually executes `(?:a*x|b*y|c*z)`. I've pushed a commit that renames it to `starredAlternationScaled` so generated benchmark manifests and reports describe the behavior being measured.

I compared the corrected branch with current main using the long Vector configuration. Time is normalized as `(PR + fix)/main`, so lower is better:

| Input size | PR/main | Result |
|---:|---:|---|
| 64 B | 0.612 | 1.64x faster |
| 256 B | 0.533 | 1.87x faster |
| 1 KiB | 0.531 | 1.90x faster |
| 10 KiB | 0.437 | 2.30x faster |
| 100 KiB | 0.486 | 2.06x faster |
| 1 MiB | 0.529 | 1.91x faster |

LGTM

eaftan (approved):

eaftan (dismissed):

eaftan (approved):

PR #746, opened 2026-08-21
Retain required infix literals in RejectPrefilter and wire into replace and split operations

cushon:
Previously, `extractRequiredLiteral` was suppressed whenever a start prefix was present. When extracting required literals, the AST search only tracked a single longest literal, which could end up being the same as the start prefix.

Additionally, `Matcher.replaceImpl()` and `Matcher.findSplitPositions()` were not evaluating RejectPrefilter at their entry points.

This change allows both prefix and required literal extraction to occur, and `extractRequiredLiteral` excludes candidates identical to the start prefix or end suffix to find distinct required literals. It also wires RejectPrefilter into replacement and split operations.

```
./safere-benchmarks/scripts/compare-branch.sh --baseline main 'RealWorldRegexBenchmark\.runBenchmark\.metadataBlock\.noMatch.*@safere-string'
```

| Benchmark                    | baseline (ns/op) | current (ns/op) | Speedup |
| ---------------------------- | ---------------- | --------------- | ------- |
| metadataBlock.noMatch.1000   |         952 ± 16 |       128 ± 5.7 |   7.43x |
| metadataBlock.noMatch.10000  |        8157 ± 31 |      1039 ± 9.3 |   7.85x |
| metadataBlock.noMatch.100000 |    124751 ± 2608 |     10137 ± 157 |   12.3x |

### DIFF
--- a/safere/src/main/java/org/safere/Matcher.java
+++ b/safere/src/main/java/org/safere/Matcher.java
@@ -2820,12 +2820,14 @@ public final class Matcher implements MatchResult {
   private String replaceImpl(String replacement, int limit) {
     Objects.requireNonNull(replacement, "replacement");
     reset();
-    RejectPrefilter rejectPrefilter = parentPattern.rejectPrefilter();
-    if (rejectPrefilter != null && text != null) {
-      if (rejectPrefilter.canReject(activeScanner(), text, searchFrom, enginePathOptions())) {
-        diagnosticParticipation(rejectPrefilter.strategy(), StrategyRole.REJECT_PREFILTER);
-        diagnosticBoundary(rejectPrefilter.strategy());
-        return text;
+    if (!parentPattern.prog().anchorStart()) {
+      RejectPrefilter rejectPrefilter = parentPattern.rejectPrefilter();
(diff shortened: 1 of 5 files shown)


eaftan (approved):
This looks like a sound extension of the rejection prefilter. Required-literal extraction remains conservative, and applying the same prefilter at replacement and split entry points avoids entering the matching engines when the input cannot match.

I found one diagnostics issue in the replacement shortcut. A composite prefilter could reject on a later child while reporting its first child's strategy. For example, `foo.*bar.*[0-9]$` against `foo---7` rejects because `bar` is absent, but diagnostics reported the end character-class filter. I've pushed a commit that reports the child that actually rejected, as the normal find path already does, and covers both `replaceFirst` and `replaceAll`.

I compared the corrected branch with current main using the long configuration. Time is normalized as `(PR + fix)/main`, so lower is better:

| `metadataBlock.noMatch` size | PR/main | Result |
|---:|---:|---|
| 1K | 0.156 | 6.40x faster |
| 10K | 0.112 | 8.90x faster |
| 100K | 0.117 | 8.53x faster |

LGTM

PR #748, opened 2026-08-22
Support Unicode case folding and zero-allocation submatch extraction in OnePass

cushon:
Add support for `Pattern.UNICODE_CASE` in `OnePass`.

RE2's `OnePass` does not support `UNICODE_CASE` because its engine processes a byte at a time. Because SafeRE operates on 32-bit Unicode code points, it can support Unicode case folding without multi-byte state branching.

Also:
- Port `Dfa.java`'s 256-entry direct-mapped cache (`cacheCps[cp & 255]`) into `OnePass.classOf()`, avoiding `Arrays.binarySearch()` for non-ASCII Unicode characters.
- Re-use capture arrays in `Matcher` to avoid allocations in `OnePass.search()`.

```
./safere-benchmarks/scripts/compare-branch.sh --baseline unicode-onepass-benchmarks-baseline 'UnicodeCaseFoldingBenchmark.*safere-string'
```

| Benchmark                                      | baseline (ns/op) | current (ns/op) | Speedup |
| ---------------------------------------------- | ---------------- | --------------- | ------- |
| UnicodeCaseFoldingBenchmark.linearChainCapture |     38800 ± 3108 |     18533 ± 303 |   2.09x |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -2447,6 +2447,13 @@
         "length": "{size}"
       },
       "shared": true
+    },
+    {
+      "id": "unicodeCaseFold.capture.input",
+      "recipe": {
+        "kind": "literal",
+        "text": "transaction_id=\u212aEY:98765-2024-03-15; status=OK"
+      }
     }
   ],
   "workloads": [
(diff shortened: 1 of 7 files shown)


eaftan (dismissed):
Nice find and improvement to OnePass!  

I noticed that the new direct-mapped Unicode lookup cache was mutable state on `OnePass`, which is shared by every matcher created from a `Pattern`. Its code-point and class arrays were updated separately, so concurrent matchers could observe a code point from one update with the class index from another. I've pushed a commit that keeps `OnePass` immutable by retaining binary search for non-ASCII classes and adds concurrent shared-pattern coverage.

I measured both the submitted and corrected revisions against current main using the long configuration. Time is normalized as `revision/main`, so lower is better:

| Revision | `linearChainCapture` | Result |
|---|---:|---|
| Submitted PR | 0.245 | 4.08x faster |
| PR + fix | 0.244 | 4.09x faster |

The fix has no measurable cost on the declared workload, and the improvement against current main is stronger than the 2.09x reported in the description.

LGTM

eaftan (dismissed):

cushon:
Thanks for the review and catching the thread-safety issue! I also filed #757 to consider adding `@Immutable` and `@ThreadSafe` annotations as a follow-up.

eaftan (dismissed):

eaftan (approved):

PR #749, opened 2026-08-22
Hoist WorkLimit checks out of Boyer-Moore Horspool inner loops

cushon:
This fixes a performance bug in Boyer-Moore Horspool by hoisting `WorkLimit` checks out of the inner loops.

#751 replaces BMP with HashChain, so that makes this fix obsolete. If #751 runs into any issues then this is definitely worth fixing as long as BMP is still used.

```
./safere-benchmarks/scripts/compare-branch.sh --baseline origin/main 'captureFreeBytes'
```

| Benchmark                                               | baseline (ns/op) | current (ns/op) | Speedup |
| ------------------------------------------------------- | ---------------- | --------------- | ------- |
| Utf8MatchingBenchmark.captureFreeBytes.asciiEarly       |         20 ± 2.9 |       16 ± 0.28 |   1.21x |
| Utf8MatchingBenchmark.captureFreeBytes.asciiLate        |        48 ± 0.65 |        46 ± 3.0 |   1.04x |
| Utf8MatchingBenchmark.captureFreeBytes.asciiNoMatch     |        19 ± 0.33 |       19 ± 0.49 |   0.96x |
| Utf8MatchingBenchmark.captureFreeBytes.multibyteEarly   |         20 ± 2.9 |       17 ± 0.06 |   1.15x |
| Utf8MatchingBenchmark.captureFreeBytes.multibyteLate    |        37 ± 0.50 |       34 ± 0.30 |   1.09x |
| Utf8MatchingBenchmark.captureFreeBytes.multibyteNoMatch |        15 ± 0.03 |       15 ± 0.34 |   0.98x |

### DIFF
--- a/safere/src/main/java/org/safere/Utf8InputScanner.java
+++ b/safere/src/main/java/org/safere/Utf8InputScanner.java
@@ -532,7 +532,7 @@ final class Utf8InputScanner extends ByteSwarScan implements InputScanner {
    * @return the index of the first match, {@code -1} if the literal is absent, or {@code -2} if the
    *     work budget was exhausted before either could be established
    */
-  private int boundedBoyerMooreHorspool(byte[] literal, int[] shifts, int start) {
+  int boundedBoyerMooreHorspool(byte[] literal, int[] shifts, int start) {
     int last = literal.length - 1;
     int position = start + last;
     long work = 0;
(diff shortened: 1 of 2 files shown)


cushon:
The outcome of #751 is that there are cases where Boyer-Moore Horspool is worth keeping, so this performance fix is still relevant.

eaftan (approved):
Thanks for the fix, I agree it's worth merging even with the hash-chain work.

I noticed that the winning-candidate regression test did not get close enough to the work budget to distinguish this change from the old inner-loop check. I've pushed a commit that uses a candidate where the old implementation exhausts the budget partway through the eventual match, so the test now covers the behavior this PR changes.

I also reproduced the UTF-8 benchmark direction against current main: the late-match cases take about 10–12% less time, the multibyte early/no-match cases take about 5–6% less time, and the remaining cases are effectively neutral within uncertainty. LGTM.

PR #751, opened 2026-08-23
Add 2-gram `HashChain` and `ClassHashChain` literal search acceleration

cushon:
This PR adds 2-gram `ClassHashChain` sublinear acceleration for case-insensitive `String` APIs. UTF-8 retains its existing BMH/scalar/SWAR paths.

Background: https://arxiv.org/abs/2310.15711v1

For case-insensitive literals on `String` inputs, matching performance improves by ~4x across all buffer sizes. A hybrid strategy pairs the HotSpot `String.indexOf` intrinsic with `shiftAt` candidate skipping on anchor hits.

---

* 2-Gram Sublinear Search Kernel
   - `ClassHashChain`: 1024-entry shift table using 16-bit 2-gram hashing and Unicode simple folding for sublinear candidate shift queries on `String`.

- Precompiles `ClassHashChain` instances once during pattern compilation inside `StartDescriptor` and `MatchDescriptor`. Shares precompiled instances across `LiteralPreparedRunner` and `StringStartAccelerator`, eliminating repetitive table allocations during matching.
- In `Matcher.indexOfIgnoreCase`, uses `String.indexOf(low)` / `String.indexOf(high)` intrinsics for fast sweeps over dead space. When an anchor character is hit, queries `classHashChain.shiftAt(text, pos)` to skip forward up to $M - 1$ characters without paying $O(M)$ linear verification overhead. Falls back to `classHashChain.search(...)` only if pathological false anchors exhaust the `WorkLimit` budget.
- Makes Knuth-Morris-Pratt (KMP) failure table allocation lazy, computing it only upon `WorkLimit` linear fallback rather than eagerly at pattern compilation.

---

```
./safere-benchmarks/scripts/compare-branch.sh --baseline origin/main 'caseInsensitiveKeyword.*@safere-string'
```

| Benchmark                                    | baseline (ns/op) | current (ns/op) | Speedup |
| -------------------------------------------- | ---------------- | --------------- | ------- |
| ApplicationBenchmark.caseInsensitiveKeywords |         461 ± 14 |        460 ± 16 |   1.00x |
| caseInsensitiveKeyword.match.1000            |      10405 ± 126 |       2591 ± 40 |   4.02x |
| caseInsensitiveKeyword.match.10000           |    107615 ± 1086 |     26190 ± 255 |   4.11x |
| caseInsensitiveKeyword.match.100000          |  1041152 ± 27376 |   258957 ± 2293 |   4.02x |
| caseInsensitiveKeyword.noMatch.1000          |        121 ± 7.7 |        84 ± 1.1 |   1.45x |
| caseInsensitiveKeyword.noMatch.10000         |         433 ± 14 |       336 ± 9.2 |   1.29x |
| caseInsensitiveKeyword.noMatch.100000        |        3588 ± 63 |      2871 ± 265 |   1.25x |
| caseInsensitiveKeywordFind.match.1000        |        158 ± 9.8 |       38 ± 0.76 |   4.15x |
| caseInsensitiveKeywordFind.match.10000       |         159 ± 14 |       38 ± 0.69 |   4.19x |
| caseInsensitiveKeywordFind.match.100000      |         157 ± 12 |       38 ± 0.76 |   4.13x |
| caseInsensitiveKeywordFind.noMatch.1000      |         86 ± 2.8 |        65 ± 1.1 |   1.32x |
| caseInsensitiveKeywordFind.noMatch.10000     |        397 ± 7.4 |       314 ± 5.0 |   1.26x |
| caseInsensitiveKeywordFind.noMatch.100000    |       3418 ± 145 |       3027 ± 40 |   1.13x |


eaftan (commented):
I wasn't aware of this approach; thanks for sharing the paper.

I found several correctness issues in the submitted scanners. A terminal 2-gram hash collision could be accepted without verifying the complete literal. Exhausting the hash-chain work budget could be treated as no-match instead of continuing with the linear fallback. The Unicode path assumed fixed-width two-member folds, which did not cover all parser-produced Unicode prefixes, supplementary code points, or equivalents whose UTF-8 width changes. Captured case-insensitive literals could also lose the anchor needed by replacement, and malformed trusted UTF-8 could be read past its candidate range. I've pushed a commit that fixes those cases, adds regression coverage, and avoids retaining a duplicate exact-literal hash table.

The performance picture differs substantially across input representation, case sensitivity, and whether the search succeeds. These results compare the corrected PR with current main using the long configuration:

| Representation | Case sensitivity | Successful input | Rejected input | Takeaway |
|---|---|---|---|---|
| String | Case-sensitive | Continues to use HotSpot `String.indexOf`. | Same. | Keep the existing path. |
| String | Case-insensitive | 4.09–4.33x faster. | Neutral to 1.25x faster. | Keep the hash-chain-shift hybrid. |
| UTF-8 | Case-sensitive | Mostly neutral; one late ASCII match is 1.12x faster. | No consistent winning range below SWAR; the ASCII sweep is as much as 2.15x slower. | Restore the previous short-input BMH path. |
| UTF-8 | Case-insensitive | 14–48% slower at 64–256 bytes and about 12% faster at 512–599 bytes. | Takes 1.68–4.13x as long at 64–256 bytes and is neutral at 512–599 bytes. | Retain the previous scalar/SWAR selection. |

I've also pushed focused case-insensitive UTF-8 benchmarks covering successful `replaceAll()` and rejected `find()` at sizes around the hash-chain-to-SWAR cutoff. Those measurements are summarized in the last row of the table.

My takeaway is to use hash-chain for case-insensitive String search, where it is a clear win, but retain the existing BMH/scalar/SWAR paths for UTF-8. The UTF-8 gains are either isolated or confined to a narrow size window, while the regressions are broader and substantially larger. What do you think?

cushon:
Thanks for the analysis and improvements to the approach here!

> My takeaway is to use hash-chain for case-insensitive String search, where it is a clear win, but retain the existing BMH/scalar/SWAR paths for UTF-8.

I took another look at this and agree with you. I have updated this to use that approach, and remove the changes to UTF-8 and case-sensitive String handling.

eaftan (dismissed):
Thanks for narrowing this to the case-insensitive String path. I noticed two Unicode compatibility issues in the updated fast paths. An ASCII-only `CASE_INSENSITIVE` literal could broaden an exact non-ASCII character such as `ß`, and optimized replacement could miss a one-code-unit Unicode-folded prefix even when `find()` matched it. I've pushed commits that fix both cases and add regression and fuzz coverage.

The String benchmark result matches the intended outcome against current main: successful searches are 4.01–4.11× faster, while rejected inputs range from neutral to 1.27× faster. A separate UTF-8 control is effectively unchanged, as expected for the restored UTF-8 paths.

eaftan (dismissed):

eaftan (approved):

PR #758, opened 2026-08-24
Shift based scalar DFA

cushon:
This change introduces `ShiftDfa`, a specialized compact scalar DFA execution engine for small deterministic regular expressions (≤ 10 states) without capturing groups.

Background:

* https://gist.github.com/pervognsen/218ea17743e1442e59bb60d29b1aa725
* which is mentioned in https://burntsushi.net/regex-internals/

Implementation

- Packed Shift-Based Transition Table: Encodes the state machine in a single flat `long[256]` transition table. Transitions are computed via bit-shift and mask: `nextState = (table[c] >>> state) & 0x3F`. Each state occupies 6 bits (allowing up to 10 states * 6 bits = 60 bits within a 64-bit `long`), with a dead state sentinel at offset `60` preventing bit-shift overflow on invalid transitions.
- Zero Allocations & Cache Locality: Eliminates state object dereferences, allocation overhead, and branching per character during DFA traversal.
- Self-Loop Acceleration: Integrates `StateAccelerator` (SIMD / vector and SWAR byte-scanning) to skip ASCII self-loop runs while stopping at either an ASCII escape or non-ASCII input before resuming scalar transitions.
- Engine Selection Precedence: Uses `ShiftDfa` ahead of `OnePass` for eligible full-input `matches()` operations when the pattern has 0 capturing groups and compiles to <= 10 deterministic states. Priority-sensitive `find()` and `lookingAt()` operations continue to use the existing engines.

The following standard results compare the corrected PR with a controlled current-main baseline using identical benchmark declarations:

| Benchmark | String speedup | UTF-8 speedup |
|---|---:|---:|
| ShiftDfaBenchmark.booleanToken | 1.30x | 1.44x |
| ShiftDfaBenchmark.identifier.16 | 1.61x | 1.80x |
| ShiftDfaBenchmark.identifier.256 | 2.94x | 3.36x |
| ShiftDfaBenchmark.identifier.4096 | 2.99x | 4.94x |
| ShiftDfaBenchmark.digitsRange | 1.88x | 1.16x |
| ShiftDfaBenchmark.stringLiteral.16 | 1.92x | 1.82x |
| ShiftDfaBenchmark.stringLiteral.256 | 5.41x | 13.4x |
| ShiftDfaBenchmark.stringLiteral.4096 | 6.46x | 21.3x |

The submitted 4096-character String-literal result reproduced at 79.7x in long mode, but that implementation could skip a non-ASCII dead transition. After making the accelerator stop at non-ASCII input, the corrected long-mode result is 6.52x.


cushon:
I thought some of the performance here might be from DFA inner loop acceleration and not specifically the shift DFA, but #760 suggests it's mostly the shift DFA.

eaftan (dismissed):
This is really interesting.  How did you hear about this approach?

I noticed some bugs in the new prepared runner. Its `lookingAt()` result used the last accepting state, which changes Java's alternation and reluctant-quantifier priority; its ASCII self-loop accelerator could skip an intervening non-ASCII character even though ShiftDfa treats all non-ASCII transitions as dead; and direct UTF-8 matching ignored nonzero backing-array offsets. I've pushed a commit that delegates priority-sensitive prefix matching to the existing core engines, makes ShiftDfa's accelerator stop at non-ASCII input, honors UTF-8 view offsets, and adds regression and fuzz coverage. The commit also adds deterministic work accounting and replaces a benchmark that was actually selecting the keyword-alternation fast path.

Against current main, the corrected workloads produce these speedups:

| Workload | String | UTF-8 |
|---|---:|---:|
| Boolean token | 1.30× | 1.44× |
| Identifier, 16 characters | 1.61× | 1.80× |
| Identifier, 256 characters | 2.94× | 3.36× |
| Identifier, 4,096 characters | 2.99× | 4.94× |
| Digits range | 1.88× | 1.16× |
| String literal, 16 characters | 1.92× | 1.82× |
| String literal, 256 characters | 5.41× | 13.4× |
| String literal, 4,096 characters | 6.46× | 21.3× |

The submitted 4,096-character String-literal result reproduced at 79.7× in long mode, but that speedup depended on the unsafe non-ASCII skipping; the corrected long-mode result is 6.52×. I've updated the description and benchmark table to reflect the corrected behavior and numbers.

LGTM

cushon:
Thanks for the fixes!

I came across the gist through the burntsushi blog post above, I think the gist is the canonical writeup of the approach by Per Vognsen. I see a couple of other citations of the gist:

* https://x.com/_rsc/status/1413843059972923394
* https://blogs.perl.org/users/chansen/2026/04/faster-utf-8-validation.html

eaftan (approved):

PR #760, opened 2026-08-24
Enable DFA self-loop escape acceleration for anchored start states and interior loops

cushon:
In #745, automata-derived escape acceleration via `StateAccelerator` was introduced for DFA states, but inner loop was guarded with `!s.isStartState`.

For anchored searches, the start state can also safely accelerate self-loops over non-escape characters directly within the inner scan loop. Updating the inner loop condition from `!s.isStartState` to `(anchored || !s.isStartState)` enables inner loop escape acceleration for anchored start states while preserving unanchored `find()` search dispatch.

I thought some of the performance in #758 might be from adding more DFA loop acceleration, but this change by itself is a relatively small improvement for the Lazy DFA. I think it's probably worth doing to apply the existing inner loop acceleration where we can.

```
./safere-benchmarks/scripts/compare-branch.sh --baseline dfa-self-loop-baseline '(DfaInteriorLoopBenchmark|DfaLoopMatchesBenchmark).*@safere-string'
```

 | Benchmark                                  | baseline (ns/op) | current (ns/op) | Speedup |
| ------------------------------------------ | ---------------- | --------------- | ------- |
| DfaInteriorLoopBenchmark.quotedString.16   |         91 ± 6.8 |        100 ± 14 |   0.91x |
| DfaInteriorLoopBenchmark.quotedString.256  |        110 ± 4.9 |       108 ± 2.4 |   1.02x |
| DfaInteriorLoopBenchmark.quotedString.4096 |         261 ± 39 |       234 ± 5.2 |   1.11x |
| DfaLoopMatchesBenchmark.ambiguousLoop.16   |        171 ± 7.2 |        174 ± 11 |   0.98x |
| DfaLoopMatchesBenchmark.ambiguousLoop.256  |        638 ± 9.8 |        636 ± 13 |   1.00x |
| DfaLoopMatchesBenchmark.ambiguousLoop.4096 |       8269 ± 410 |      8555 ± 106 |   0.97x |
| DfaInteriorLoopBenchmark.cComment.16       |       144 ± 0.91 |       150 ± 2.4 |   0.96x |
| DfaInteriorLoopBenchmark.cComment.256      |        268 ± 6.3 |        269 ± 11 |   1.00x |
| DfaInteriorLoopBenchmark.cComment.4096     |        2236 ± 20 |       2093 ± 21 |   1.07x |
| DfaLoopMatchesBenchmark.wildcardQuote.16   |        173 ± 6.8 |       172 ± 3.4 |   1.01x |
| DfaLoopMatchesBenchmark.wildcardQuote.256  |        166 ± 5.8 |       149 ± 7.3 |   1.12x |
| DfaLoopMatchesBenchmark.wildcardQuote.4096 |        270 ± 6.8 |       240 ± 5.4 |   1.12x |


eaftan (approved):
This seems like a good improvement that increases the applicability of the DFA self-loop escape accelerator.

I was puzzled by the benchmark results, particularly the regressions in the `cComment` and `quotedString` workloads. Looking closely, this change should not affect those patterns: they use acceleration from an interior DFA state, so they should take the accelerated path both before and after this PR.

I dug into the regressions and found that changing the conditional appears to have perturbed the code generated by HotSpot for the existing interior-state path. Profiling showed additional time inside the same inlined input scanner, rather than additional matching or scanning work. I reordered the conditions so the existing length check is performed before distinguishing between start and interior states. This is logically equivalent, but produces better compiled code and removes the reproducible regressions.

I reran the complete workload family after the fix. Time is normalized as `(PR + fix)/main`, so lower is better:

| Benchmark | String PR/main | UTF-8 PR/main | Assessment |
|---|---:|---:|---|
| quotedString.16 | 0.952 | 1.000 | Neutral |
| quotedString.256 | 0.975 | 1.000 | Neutral |
| quotedString.4096 | 1.024 | 0.964 | Neutral |
| ambiguousLoop.16 | 1.048 | 1.040 | Neutral |
| ambiguousLoop.256 | 0.995 | 0.975 | Neutral |
| ambiguousLoop.4096 | 0.994 | 1.002 | Neutral |
| cComment.16 | 1.176 | 1.084 | Inconclusive; point estimates slower |
| cComment.256 | 1.048 | 1.007 | Neutral; long-mode String was 0.985 |
| cComment.4096 | 1.037 | 1.030 | String neutral in long mode; UTF-8 inconclusive |
| wildcardQuote.16 | 0.970 | 1.017 | Neutral |
| wildcardQuote.256 | 0.919 | 0.962 | Faster for both representations |
| wildcardQuote.4096 | 0.927 | 0.936 | String confirmed faster; UTF-8 favorable but inconclusive |
| Geometric mean | 1.003 | 1.001 | Neutral overall |

I've labeled results where the 99.9% confidence intervals overlapped as either neutral or inconclusive.  

I’ve pushed commit 3c76368e with the condition reordering. Overall, the updated PR is performance-neutral on the previously accelerated workloads while retaining measurable improvements where the accelerator is newly applicable, particularly the larger `wildcardQuote` workloads.

PR #761, opened 2026-08-24
Reverse suffix acceleration

cushon:
Implements reverse suffix and inner search acceleration for unanchored regular expressions containing a complex prefix followed by a selective literal or multi-literal anchor.

See also 'reverse suffix' discussion in https://burntsushi.net/regex-internals/

When searching large haystacks for unanchored patterns without a simple literal prefix, SafeRE previously had to evaluate prefix character classes or fallback to forward DFA byte-by-byte stepping. This PR introduces bidirectional acceleration:

1. Use existing start accelerators to search forward for the selective suffix/inner anchor
2. Run a reverse DFA backwards from the anchor position to locate the leftmost match start position for the preceding prefix
3. Pass the confirmed match start directly to the linear execution pipeline

Implementation:

* `StartDescriptor.ReverseAnchor` encapsulates the inner anchor's `StartDescriptor`, the compiled reverse prefix `Prog`, and the minimum required match length
* `StringStartAccelerator.ReverseAnchor` and `Utf8StartAccelerator.ReverseAnchor` implement bidirectional search, scanning forward to candidate anchors and executing a bounded reverse DFA search backwards.
- Analysis and gating in `Pattern.java`:
  - `extractReverseAnchor` decomposes `metadataAst` concatenations into prefix + anchor pairs
  - `isSelectiveReverseAnchor` requires high-selectivity anchors (literals of length $\ge 2$, high-rarity single characters, Teddy models, or multi-literals; rejects generic character classes)
  - `containsWildcardOrZeroWidth` recursively ensures the prefix contains no wildcards (`.`, `\C`) or zero-width assertions (`^`, `$`, `\b`, `\B`)
  - non-nullable prefix invariant (`minPrefixLen >= 1`) ensures reverse scan determines exact JDK leftmost-match start boundaries
* When the start accelerator anchor already covers the required literal, `RejectPrefilter` is suppressed to avoid redundant double-scanning on matching inputs, while retaining `RejectPrefilter` when infix literals are disjoint and rarer than the start prefix.

```
 ./safere-benchmarks/scripts/compare-branch.sh --baseline reverse-suffix-benchmark-baseline '(emailPathFind|serviceNodeFind).*@safere-string'
```

| Benchmark                      | baseline (ns/op) | current (ns/op) | Speedup |
| ------------------------------ | ---------------- | --------------- | ------- |
| RegexBenchmark.emailPathFind   |        3052 ± 39 |        776 ± 23 |   3.93x |
| RegexBenchmark.serviceNodeFind |        7922 ± 58 |        700 ± 11 |  11.30x |

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -195,6 +195,26 @@
       "shared": true
     },
     {
+      "id": "crossEngine.RegexBenchmark.emailFindDoc.input",
+      "recipe": {
+        "kind": "suffixToLength",
+        "prefixUnit": "The quick brown fox jumps over the lazy dog. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer nec odio. Praesent libero. Sed cursus ante dapibus diam. ",
+        "suffix": "Contact support directly at service_alerts123@support.internal.org for any system assistance.",
+        "length": 4096
+      },
+      "shared": true
+    },
+    {
+      "id": "crossEngine.RegexBenchmark.tokenSuffixFind.input",
+      "recipe": {
+        "kind": "suffixToLength",
+        "prefixUnit": "timestamp=2026-08-24T12:00:00Z level=INFO module=engine message=\"processing request packet chunk 0x4f2a\"\n",
+        "suffix": "target_service_node=node987-staging status=READY",
+        "length": 4096
+      },
+      "shared": true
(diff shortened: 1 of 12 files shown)


cushon:
I still think this is worth trying but the implementation will benefit from #769

cushon:
This has been subsumed into the series for #769

Issue #764, opened 2026-08-25
Single-character required-literal find-all ~8% slower since WorkLimit hoisting (#749)


wendigo:
While benchmarking SafeRE as a regexp backend for Trino, we found that find-all over a single-character required literal got ~8% slower with #749 (Hoist WorkLimit checks out of Boyer-Moore Horspool inner loops, 4636c4d), and the slowdown persists on current main (e58ce08).

This is much smaller than the dense-input regression reported separately, but it is reproducible across two commits and looks like collateral from a change that was intended to be neutral-or-faster, so it seemed worth flagging.

## Workload

`.*x.*` matched with a `Utf8Matcher.find()` loop over 32 KB of `aaaa...` (no match ever occurs, so each op is a single accelerated scan for the required literal `x`).

## Measurements

Pinned-commit builds, JMH harness, 5x2s iterations, `--add-modules=jdk.incubator.vector`, Apple M4 Max, OpenJDK 26, macOS 15:

| commit | ns/op |
|---|---|
| c41f220 (#702) | 991 ± 13 |
| 4636c4d (#749) | 1,082 ± 45 |
| e58ce08 (#758, current main) | 1,070 ± 20 |

Commits between c41f220 and 4636c4d (#762, #744, #748, #745, #737, #756, #746, ...) were covered by a bisect over the wider range and did not move this cell; the step lands at 4636c4d. A standalone loop (no JMH) shows the same direction but with more run-to-run wobble (~950-1,130), so treat the exact percentage as approximate; the JMH error bars above don't overlap.

## Reproducer

```java
import org.safere.Pattern;
import org.safere.Utf8Input;
import org.safere.Utf8Matcher;

public class Repro {
    public static void main(String[] args) {
        byte[] text = new byte[32768];
        java.util.Arrays.fill(text, (byte) 'a');
        Pattern pattern = Pattern.compile(".*x.*");

        for (int round = 0; round < 10; round++) {
            long matches = 0;
            long start = System.nanoTime();
            for (int i = 0; i < 2000; i++) {
                Utf8Matcher matcher = pattern.matcher(Utf8Input.trusted(text, 0, text.length));
                while (matcher.find()) {
                    matches++;
                }
            }
            System.out.printf("%8.1f ns/op  (%d matches/op)%n",
                    (System.nanoTime() - start) / 2000.0, matches / 2000);
        }
    }
}
```

[...]

PR #766, opened 2026-08-26
Gate character class acceleration on selectivity and restore primitive DFA loop

cushon:
This partially reverts #720 (keeping some of the refactoring on other cleanups from that PR). That change added character class acceleration for broad character classes  (e.g. `\d`, `\w`, `[0-9]`), which can have high match density and regress performance for some inputs.

Restricting start acceleration to selective classes ($\le 3$ ASCII characters, such as delimiters `[,;:]` or quotes `['"]`) ensures broad classes execute directly in the scalar DFA without prefilter overhead.

This improves the benchmarks that regressed in #763, using the JMH version in the gist from https://github.com/eaftan/safere/issues/764#issuecomment-5417263093 shows `issue763_matchDense_digitsFindAll32k` runs in ~23% less time after this change:

| version | time | 
| ---     | --- | 
| 906d4c317ccd2f91015344d509dc00081ef70d5a | 102746.775 ± 549.625  ns/op |
| this change  | 79105.647 ± 643.503  ns/op |  |

---

* Restrict `Dfa.analyzeStateAcceleration` to $\le 3$ exact ASCII escape characters (`SingleAsciiEscape`, `AsciiPairEscape`, `AsciiTripleEscape`) and remove `CharClassEscape`.
* Restore the pure primitive `int` inner transition loop in `Dfa.doSearch()`, removing per-byte `offsetToState` reference array lookups and branch checks inside `while (pos < limit)`.

### DIFF
--- a/safere/src/main/java/org/safere/CharClassScanInfo.java
+++ b/safere/src/main/java/org/safere/CharClassScanInfo.java
@@ -24,11 +24,8 @@ sealed interface CharClassScanInfo {
   int[] ranges();
 
   default boolean isSelective() {
-    if (!isAscii()) {
-      return true; // Unicode classes have wide gaps in ASCII text
-    }
     int count = Long.bitCount(bitmap0()) + Long.bitCount(bitmap1());
-    return count > 0 && count <= 3;
+    return count <= 3;
   }
 
   /** Matches 1, 2, or 3 exact ASCII characters via single-instruction SIMD equality. */
--- a/safere/src/main/java/org/safere/CharClassScanInfo.java
+++ b/safere/src/main/java/org/safere/CharClassScanInfo.java
@@ -23,6 +23,14 @@ sealed interface CharClassScanInfo {
 
   int[] ranges();
 
+  default boolean isSelective() {
+    if (!isAscii()) {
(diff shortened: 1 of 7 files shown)


eaftan (approved):
Thanks for this.  I can reproduce positive benchmark results. I compared this branch plus a fix (see below) with current main on the checked-in dense digits-and-slashes control:

| Benchmark | main | PR + fixes | PR/main | Result |
|---|---:|---:|---:|---|
| String, standard | 54,048 ± 2,520 ns/op | 50,563 ± 2,892 ns/op | 0.936 | 6.5% less time |
| UTF-8, standard | 57,926 ± 1,834 ns/op | 50,986 ± 2,062 ns/op | 0.880 | 12.0% less time |
| UTF-8, long | 57,937 ± 1,267 ns/op | 50,129 ± 1,054 ns/op | 0.865 | 13.5% less time |

I noticed that `isSelective()` treated every Unicode class as selective, so a class such as `[0-9é]` continued to pay dense start-acceleration overhead on numeric ASCII input. I've pushed a commit that applies the ASCII bitmap-density threshold to Unicode classes too, with coverage for dense mixed, sparse mixed, and non-ASCII-only classes.

LGTM

PR #768, opened 2026-08-26
Consolidate redundant AST walks into a unified single-pass AstAnalysis walker

cushon:
This is a pre-factoring / cleanup to prepare for some improvements to reject prefilters and start descriptor handling, see #769.

---

- Introduce `AstAnalysis` and `AstAnalysisWalker` to compute structural AST metadata in a single bottom-up traversal using the stack-safe `Walker`.
- Extract named capture groups, user capture presence, lazy quantifiers, alternation structure, nullable alternation, empty match capability, and minimum match length simultaneously.
- Replace separate, redundant AST traversals in `Pattern.compile` and `Compiler` (`extractNamedGroups`, `hasLazyQuantifiers`, `hasAlternation`, `hasNullableAlternation`, `canMatchEmpty`, `CanMatchEmptyWalker`).
- Add unit tests in `AstAnalysisTest`.


cushon:
Closing in favor of #770, which is the same change but from a branch on `eaftan/safere` instead of my fork to try to make it a stacked PR

Issue #769, opened 2026-08-26
Unified gap-and-segment model for accelerators, reject prefilters, and multi-anchor matching


cushon:
Today, SafeRE implements several related performance optimizations across different tiers of the matching pipeline:

*  Rejection Filters: Required substring search, end-anchored suffixes, required character classes
* Start Accelerators: Literal prefixes, fixed-offset literals, leading repetition expansions, multiline anchors

This issue proposes adopting a unifying 'gaps and segments' model across AST analysis, reject prefilters, and matcher acceleration, that can be used to express those existing optimizations and generalize them to support more cases.

The idea is that any regular expression can be decomposed into a sequence of anchors and gaps:

* An anchor is an element that can be found with hardware accelerated scanning (e.g. Teddy, `String.indexOf` intrinsics, etc.), including
  * single ASCII literals (`foo`)
  * disjoint literal sets / alternations (`foo|bar|baz`)
  * high-selectivity character classes
* A gap is an intervening sub-expression
  * gaps can be validated with ASCII bitmap checks, single-byte loops, or bounded sub-engines

Every existing start-position accelerator and whole-input reject filter in SafeRE is a 1-anchor specialization of the general model:

| Existing Optimization | Special Case Representation |
| :--- | :--- |
| **Literal Prefix** (`hello.*`) | `Gap 0: empty`, `Anchor 1: "hello"`, `Gap 1: .*` |
| **Fixed-Offset Literal** (`[0-9]{4}-target`) | `Gap 0: width 4`, `Anchor 1: "-target"`, `Gap 1: empty` |
| **Leading Expansion** (`\s+keyword`) | `Gap 0: \s+ (width >= 1)`, `Anchor 1: "keyword"`, `Gap 1: empty` |
| **End-Anchored Suffix** (`.*\.json$`) | `Gap 0: .*`, `Anchor 1: ".json"`, `Gap 1: end anchor ($)` |
| **Reject Filter: Required Literal** (`.*foo.*`) | `Gap 0: .*`, `Anchor 1: "foo"`, `Gap 1: .*` |
| **Reject Filter: Disjoint Literals** (`.*(cat\|dog).*`) | `Gap 0: .*`, `Anchor 1: "cat" \| "dog"`, `Gap 1: .*` |

Having a single unified representation will replace ad-hoc AST extractors across Pattern with a single consistent model, and make it easier to add any new reject prefilters or start accelerators.

[...]

PR #770, opened 2026-08-26
Consolidate redundant AST walks into a unified single-pass AstAnalysis walker

cushon:
This is a pre-factoring / cleanup to prepare for some improvements to reject prefilters and start descriptor handling, see #769.

---

- Introduce `AstAnalysis` and `AstAnalysisWalker` to compute structural AST metadata in a single bottom-up traversal using the stack-safe `Walker`.
- Extract named capture groups, user capture presence, lazy quantifiers, alternation structure, nullable alternation, empty match capability, and minimum match length simultaneously.
- Replace separate, redundant AST traversals in `Pattern.compile` and `Compiler` (`extractNamedGroups`, `hasLazyQuantifiers`, `hasAlternation`, `hasNullableAlternation`, `canMatchEmpty`, `CanMatchEmptyWalker`).
- Add comprehensive unit tests in `AstAnalysisTest`.

### DIFF
--- a/safere-crosscheck/pom.xml
+++ b/safere-crosscheck/pom.xml
@@ -104,6 +104,7 @@
                         <exclude name="AnchoredDfaCachingTest.java"/>
                         <exclude name="AsciiBitmapTest.java"/>
                         <exclude name="AsciiTest.java"/>
+                        <exclude name="AstAnalysisTest.java"/>
                         <exclude name="BitStateTest.java"/>
                         <exclude name="CharClassTest.java"/>
                         <exclude name="CompilerTest.java"/>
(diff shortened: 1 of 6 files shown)


eaftan (approved):
This is a nice cleanup.  

I noticed that the new literal analysis treated rune value zero as an absent rune, but `re.rune == 0` represents the valid U+0000 literal. That marked `\x00` nullable and unnecessarily disabled OnePass primary for an alternation such as `\x00|a`. I've pushed a commit that fixes the literal case and adds regression coverage for the AST and public capability metadata.

LGTM

eaftan (approved):

PR #772, opened 2026-08-26
Keep the SWAR filter's scalar tail out of the hot word loop

wendigo:
## Summary

- Extract the scalar tail scan out of `ByteSwarScan.indexOfFiltered` into a private `scalarTail` method, keeping the zero-allocation tail from #751 while letting C2 fully optimize the SWAR word loop again
- Add a `Utf8MatchingBenchmark.literalScan.longWordPrefix` workload that reaches the same long literal through the start accelerator (`coolfunctionname.*`), the shape SQL engines produce when rewriting LIKE-style patterns

## Root cause

#751 replaced `indexOfFiltered`'s tail call to linear KMP with an inlined scalar loop (to avoid materializing the now-lazy failure table on long-buffer misses). Inlining that loop into the method body pushed `indexOfFiltered` past C2's inlining budget and deoptimized the hot SWAR word loop: long-literal find over 32 KB of random lowercase text regressed from ~4.1 us to ~6.3 us (+54%). JFR shows both versions spending all samples in `indexOfFiltered` — same algorithm, same path, worse compilation.

This was first observed downstream in Trino, where `regexp_like('...', '.*coolfunctionname.*')` over 32 KB inputs slowed by ~80% after upgrading past #751.

## Measurements

Interleaved A/B, Apple M4 Max, OpenJDK 26, `--add-modules=jdk.incubator.vector`, min of 5 rounds x 3 alternating JVM runs, `Pattern.find(Utf8Input)` with `coolfunctionname.*` over 32 KB seeded random `a-z`:

| build | ns/op |
|---|---|
| before #751 (e58ce08) | 4,100-5,300 |
| after #751 (f6b5578) | 6,500-8,100 |
| this change | 4,000-5,200 |

The existing `literalScan.longWord` workload covers the pure-literal fast path for this shape; the new `longWordPrefix` workload covers the start-accelerator route where the regression was observed downstream.

Correctness is covered by the existing tail tests (`multiByteLiteralSearchCoversEveryWindowAlignmentAndPosition` exercises every window alignment and tail position); all `safere` scanner tests and `safere-benchmarks` plan tests pass.

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -9691,6 +9691,29 @@
       }
     },
     {
+      "id": "Utf8MatchingBenchmark.literalScan.longWordPrefix",
+      "operation": "find",
+      "patterns": [
+        "coolfunctionname.*"
+      ],
+      "inputs": [
+        "utf8Matching.literalScan.text"
+      ],
+      "inputRepresentations": [
+        "preexistingUtf8"
+      ],
+      "inputRepresentationReason": "Exercises the long-literal prefix through the start accelerator rather than the pure-literal fast path, matching how SQL engines rewrite LIKE-style patterns.",
+      "requirements": [],
+      "resultConsumption": "boolean",
+      "measurement": {
+        "mode": "averageTime",
+        "timingUnit": "nanoseconds",
+        "constraints": [
+          "prematerializedInput"
+        ]
+      }
+    },
+    {
       "id": "Utf8MatchingBenchmark.hardFailure.1024",
       "operation": "find",
       "patterns": [
--- a/safere/src/main/java/org/safere/ByteSwarScan.java
+++ b/safere/src/main/java/org/safere/ByteSwarScan.java
@@ -408,6 +408,16 @@ abstract class ByteSwarScan {
         return -2;
       }
     }
+    return scalarTail(bytes, offset, length, literal, position);
+  }
+
+  /**
+   * Scans the trailing bytes that the word loop could not cover (fewer than {@link Long#BYTES} plus
+   * the literal length). Kept out of {@link #indexOfFiltered} so the hot word loop stays within
+   * C2's inlining budget; merging this loop into it measurably deoptimizes the SWAR loop.
+   */
+  private static int scalarTail(
+      byte[] bytes, int offset, int length, byte[] literal, int position) {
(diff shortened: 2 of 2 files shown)


wendigo:
cc @cushon

cushon:
Thanks! This makes sense to me. I was able to reproduce a speedup.

x86_64 GNU/Linux

| Utf8MatchingBenchmark.literalScan.fourByte       |       9117 ± 113 |      9187 ± 154 |   0.99x |
| Utf8MatchingBenchmark.literalScan.shortWord      |       8882 ± 157 |      9071 ± 785 |   0.98x |
| Utf8MatchingBenchmark.literalScan.longWord       |       9255 ± 222 |       8740 ± 46 |   1.06x |
| Utf8MatchingBenchmark.literalScan.longWordPrefix |        9117 ± 50 |       8710 ± 40 |   1.05x |

macOS

| Utf8MatchingBenchmark.literalScan.fourByte       |       4120 ± 109 |       4342 ± 67 |   0.95x |
| Utf8MatchingBenchmark.literalScan.shortWord      |        4207 ± 75 |       4281 ± 72 |   0.98x |
| Utf8MatchingBenchmark.literalScan.longWord       |        6116 ± 45 |      4115 ± 108 |   1.49x |
| Utf8MatchingBenchmark.literalScan.longWordPrefix |        4094 ± 86 |      4252 ± 169 |   0.96x |

eaftan (approved):
This looks good to me. Moving the scalar tail into a private helper is a small, behavior-preserving way to keep the SWAR word loop within C2's preferred compilation shape.

I compared current main with this PR using the four UTF-8 literal-scan workloads. Time is normalized as `PR/main`, so lower is better:

|---|---:|---|
| `Utf8MatchingBenchmark.literalScan.fourByte` | 0.943 | 5.7% faster |
| `Utf8MatchingBenchmark.literalScan.shortWord` | 0.967 | Favorable point estimate; noisy |
| `Utf8MatchingBenchmark.literalScan.longWord` | 0.964 | 3.6% faster |
| `Utf8MatchingBenchmark.literalScan.longWordPrefix` | 0.952 | 4.8% faster |

The long-literal results are close to the x86 measurements @cushon posted. The larger Apple-silicon result appears to be architecture-specific, but the change is consistently favorable in my run and does not regress the neighboring workloads.

LGTM

PR #776, opened 2026-08-26
Implement Crochemore-Perrin Two-Way string matching algorithm

cushon:
This replaces KMP fallback with [Crochemore-Perrin two-way string-matching](https://en.wikipedia.org/wiki/Two-way_string-matching_algorithm), which has the same linear time guarantee but avoids the need for allocating failure arrays, and i

The benchmarks I ran were flat, KMP is only used as a fallback, the benefit may be more avoiding having to allocate the failure array which doesn't show up in the throughput numbers.

It's more code than KMP and it's unfortunate to have five copies of it for variations of case-sensitivity and `String`/`byte[]`/`char[]`. On the other hand it's possible to have good isolated unit test coverage and fuzzing for it.


eaftan (commented):
I understand the idea of replacing the runtime KMP fallbacks with an O(1)-space algorithm, but I'm concerned about adding complexity without evidence that the removed allocations are material in practice.

I also found that this PR changed compile-time required-literal pruning from KMP to a nested substring scan. For disjoint literals, increasing the pattern size by 4x produced approximately 15x as much comparison work, violating SafeRE's linear-time guarantee during compilation. I've pushed a commit that restores KMP for that compile-time operation and adds scaling coverage.

This does not undo the main runtime benefit: the runtime search plans no longer retain KMP failure arrays, and the affected runtime fallbacks no longer allocate them. It does mean that the PR no longer eliminates failure-table allocation entirely, because required-literal pruning temporarily creates those tables during compilation.

Can you help quantify the benefit of this PR?  I'm open to the approach if the memory savings are meaningful, but without that evidence I'm not yet sure the added implementation and maintenance complexity is a good tradeoff. What do you think?

cushon:
Thanks for fixing that issue, and I think that's the right question to be asking about the benefit.

I added a benchmark comparing KMP and Two-Way, and I'm seeing Two-Way be 1.5x-3.6x faster across a range of inputs and lengths. I'm not sure how significant zero-allocation is in practice, but it should help cold patterns and with retained heap.

My understanding is that regex-automata uses two-way, and it's the state of the art choice here, and not just theoretically interesting.

There's also the maintenance and complexity to consider. If we use this approach, I think it's worth considering a build-time templating step to generate the five variations instead of duplicating them, then if there's a bug or performance fix to the loops we don't have to worry about them getting out of sync.

I have pushed the benchmark and the script to generate the variations of the two-way methods.

What do you think?

eaftan (commented):
I fixed an out-of-range work-accounting bug, added a regression test, and pushed the fix. 

The results are pretty mixed. I can reproduce the wins you reported on some inputs, but not as an overall result:

| Workload | Result versus precomputed KMP |
|---|---|
| Medium and large periodic inputs | Two-Way is 1.37-1.53× faster |
| Medium and large dense-prefix inputs | Two-Way is 2.40-4.52× faster |
| Random text inputs | Two-Way is 2.08-7.05× slower |
| All 27 input and length combinations | Two-Way is 31% slower by geometric mean |
| End-to-end String and UTF-8 benchmarks | Flat |

I'm concerned about the added complexity, especially generating five implementations and having to keep the generated code in sync, when we can't demonstrate a clear and substantial improvement. Avoiding the retained KMP failure arrays may still be worthwhile, but we haven't quantified that benefit. Can we either demonstrate a meaningful application-level or retained-heap improvement, or find a simpler way to get the memory benefit?

cushon:
I pushed a commit that improves performance on random inputs, by adding a fast loop that scans for the first matching character before entering the two-way factorization logic which is branchier and more expensive.

With that change I think it's mostly coming out ahead of KMP, but I don't have good evidence this actually matters: it's a rarely executed fallback path and I don't think the allocations KMP is doing are an issue.

I'll give this some more thought but it seems fine to shelve as not being worth the complexity.

| Workload | Haystack | Two-Way | KMP (Precomputed) | KMP (Dynamic) | Two-Way vs KMP (Precomputed) | Two-Way vs KMP (Dynamic) |
|---|---|---|---|---|---|---|
| **RANDOM_TEXT** | 64 B | 55.1 ns | 54.4 ns | 74.8 ns | 1.0x (Parity) | 1.4x faster |
| | 1 KB | 301.2 ns | 282.7 ns | 502.7 ns | 1.1x slower | 1.7x faster |
| | 64 KB | 24.5 µs | 18.7 µs | 18.8 µs | 1.3x slower | 1.3x slower |
| **DENSE_PREFIX_NOISE** | 64 B | 51.5 ns | 61.8 ns | 76.1 ns | 1.2x faster | 1.5x faster |
| | 1 KB | 372.6 ns | 893.6 ns | 1007.0 ns | 2.4x faster | 2.7x faster |
| | 64 KB | 32.2 µs | 55.0 µs | 63.3 µs | 1.7x faster | 2.0x faster |
| **PERIODIC_ADVERSARIAL** | 64 B | 147.6 ns | 108.7 ns | 184.6 ns | 1.4x slower | 1.3x faster |
| | 1 KB | 1.87 µs | 2.36 µs | 2.43 µs | 1.3x faster | 1.3x faster |
| | 64 KB | 115.8 µs | 158.7 µs | 145.2 µs | 1.4x faster | 1.3x faster |

PR #783, opened 2026-08-29
Implement Gap Engine extensions for alternation prefix factoring, weak-anchor coalescing, and exact-width assertion guards

cushon:
This PR builds on the multi-anchor gap engine (PR #782) by introducing AST alternation prefix factoring, dynamic weak-anchor coalescing, and exact-width assertion guards to accelerate branching and structured pattern matching.

---

* Alternation Common Prefix Factoring & Literal Coalescing:
  * In `MultiAnchorCompiler.factorAlternations`, extracts common multi-character literal prefixes ($\ge 2$ characters) across alternation branches (e.g. `(?:api/v2/users|api/v2/orders|...)`).
  * Seamlessly coalesces factored prefixes with adjacent literals in `CONCAT` nodes (e.g. merging outer `"/"` with inner `"api/v2/"` into `"/api/v2/"`).
  * Promotes the shared prefix into an accelerated scan, preventing DFA fallback.
* Dynamic Weak-Anchor Coalescing:
  * Analyzes intermediate literal anchors in multi-anchor chains and coalesces sub-threshold delimiters ($\le 4$ bytes) with adjacent bounded gaps into contiguous sub-expressions executed directly on the Lazy DFA.
  * Eliminates scalar loop dispatch overhead between short, high-density delimiters (such as `:` or `-`).
* Alternation Anchor Safety & Width Guards:
  * Hardens alternation anchor extraction with `hasZeroWidthAssertions` and `computeAsciiWidthRange(branch).isExact() && width.minWidth == 1`.
  * Guarantees that zero-width boundary assertions (`^`, `\A`, `\b`, `\B`) and multi-character branches are safely handled without being prematurely collapsed into single-character class anchors, preserving fidelity with JDK regex.

---

```
./safere-benchmarks/scripts/compare-branch.sh --baseline gap-extensions-benchmark-baseline 'RealWorldRegexBenchmark.*(wideFactoredApiRoute|isoTimestampLog|uuidExtraction|multiInfixLog|structuredJsonPath).*safere-string'
```

| Benchmark                            | baseline (ns/op) | current (ns/op) | Speedup |
| ------------------------------------ | ---------------- | --------------- | ------- |
| multiInfixLog.match.1000             |        2466 ± 62 |       2447 ± 28 |   1.01x |
| multiInfixLog.match.10000            |     22059 ± 1483 |     23030 ± 940 |   0.96x |
| multiInfixLog.match.100000           |  286525 ± 127971 | 286495 ± 127011 |   1.00x |
[...]

PR #797, opened 2026-09-03
Vector Aho-Corasick

cushon:
This implements vectorized Aho-Corasick to accelerate larger alternations.

A scalar version of Aho-Corasick was investigated in https://github.com/eaftan/safere/pull/540 and didn't have a benefit. This implements a vectorized version of it which is used for `byte[]` / UTF-8 inputs, which seems to have a significant speedup in that case, but AC is bypassed if the Vector API isn't available. The optimization is also gated behind a check for dense roots to fall back to the DFA is AC wouldn't be beneficial.

This also adds a missing `WorkLimit` budget to Teddy.

I experimented with some variations around Multi-Group Teddy to accelerate larger alternations and couldn't get a benefit from that, empirically handing off from the current Teddy implementation to Vector AC around 32 alternations seems to produce the best results.

```
java --add-modules jdk.incubator.vector -jar safere-benchmarks/target/benchmarks.jar   "LargeDictionaryAlternationBenchmark|MediumAlternationBenchmark"   -f 1 -wi 1 -i 2 -tu us -p haystackLength=65536
```

Large-Scale Dictionaries (`LargeDictionaryAlternationBenchmark`)

| Workload | Pattern Count ($K$) | SafeRE UTF-8 (Vector AC) | Baseline SafeRE (Lazy DFA) | Speedup vs Baseline |
| **`URL_ROUTES_256`** | 256 | **4.07 µs** (16.09 GB/s) | 32.54 µs (2.01 GB/s) | **8.0x faster** |
| **`SECURITY_SIGNATURES_1024`** | 1,024 | **122.74 µs** (534 MB/s) | 127.71 µs (513 MB/s) | **1.04x faster** |
| **`DICTIONARY_TOKENS_4096`** | 4,096 | **334.87 µs** (196 MB/s) | 357.25 µs (183 MB/s) | **1.07x faster** |

Alternation Acceleration Tiers (`MediumAlternationBenchmark`)

Non-Matching Haystack (`ABSENT` — Full Throughput Scan)

| Tier | Workload | Accelerator Engine | SafeRE UTF-8 | Baseline SafeRE (Lazy DFA) | Speedup vs Baseline |
| :--- | :--- | :--- | :---: | :---: | :---: |
| **Tier 1 ($K \le 4$)** | `TIER1_DIRECT_4` | Direct SIMD Equality | **21.40 µs** (3.06 GB/s) | 32.21 µs (2.03 GB/s) | **1.5x faster** |
| **Tier 2 ($5 \le K \le 32$)** | `TIER2_TEDDY_16` | Teddy SIMD Shuffle | **7.54 µs** (8.69 GB/s) | 32.08 µs (2.04 GB/s) | **4.3x faster** |
| **Dense Fallback** | `DENSE_FALLBACK_36` | Lazy DFA (AC bypassed) | **41.30 µs** (1.59 GB/s) | 32.05 µs (2.05 GB/s) | *Safe handoff* |
| **Tier 3 ($K > 32$, sparse)** | `TIER3_AC_ROUTES_256` | Vector Aho-Corasick | **4.07 µs** (16.08 GB/s) | 32.36 µs (2.03 GB/s) | **8.0x faster** |

Late Match (`MATCH_LATE` — Scanning + Late Match Verification)

| Tier | Workload | Accelerator Engine | SafeRE UTF-8 | Baseline SafeRE (Lazy DFA) | Speedup vs Baseline |
| :--- | :--- | :--- | :---: | :---: | :---: |
| **Tier 1 ($K \le 4$)** | `TIER1_DIRECT_4` | Direct SIMD Equality | **42.40 µs** | 82.24 µs | **1.9x faster** |
| **Tier 2 ($5 \le K \le 32$)** | `TIER2_TEDDY_16` | Teddy SIMD Shuffle | **13.94 µs** | 64.53 µs | **4.6x faster** |
| **Dense Fallback** | `DENSE_FALLBACK_36` | Lazy DFA (AC bypassed) | **207.39 µs** | 152.59 µs | *Safe handoff* |
| **Tier 3 ($K > 32$, sparse)** | `TIER3_AC_ROUTES_256` | Vector Aho-Corasick | **8.11 µs** | 74.08 µs | **9.1x faster** |


wendigo:
Impressive work! :)

eaftan (commented):
Thanks Liam 🙂 

I found three correctness issues and one resource-safety issue in the new Aho-Corasick path. Tries above 65,535 states can overflow the `char` state IDs and miss matches; a vector root filter returning `UNSUPPORTED` is treated as “not found” instead of falling back; and the ASCII root filter can skip matches for mixed ASCII/Unicode dictionaries. The dense transition table also has no size ceiling. I've pushed fixes that use 32-bit states, continue with scalar Aho-Corasick when the vector filter is unsupported, avoid the ASCII root filter for mixed dictionaries, and fall back before the dense table exceeds 32 MiB.

The `UNSUPPORTED` fix eliminates a false-negative fast exit in the security-signature workload and uncovers a large regression. Some of the other reported results either do not reproduce or reproduce worse. I compared the corrected implementation at `92309d19ef3f7bd6d4d1d525365795b051a71ddc` with controlled baseline `313ff659bb62cd02f30534d67e9d6cab6213b07b`, which contains production code from then-current main `eeee01693c24b5c4ad75f663941970f353996d91` plus only the identical workload declarations. The vector provider was enabled for both. `PR/base` is elapsed-time ratio, so lower is better.

| Workload | Main ns/op | Corrected PR ns/op | PR/base | Result |
|---|---:|---:|---:|---|
| URL-256 absent | 2,807 | 2,800 | 0.998 | neutral, not approximately 8x faster |
| URL-256 late match | 5,160 | 5,312 | 1.029 | 3% slower |
| security-1,024 absent | 81,514 | 197,660 | 2.425 | 2.43x as long |
| security-1,024 late match | 90,954 | 225,131 | 2.475 | 2.48x as long |
| dictionary-4,096 absent | 219,949 | 406,975 | 1.850 | 85% slower |
| dictionary-4,096 late match | 218,016 | 407,695 | 1.870 | 87% slower |

The URL result was neutral on the submitted code too, so the correctness fixes do not explain the missing headline gain. The submitted dictionary path was already slower than main before the fixes. Main has since advanced to `f7d415c301d2e28bffcad730b03e9d648a95f58b`, so these results need to be rerun after rebasing onto current main. Could you also profile the regressing workloads and compare forced Aho-Corasick execution with the existing-engine path? That should tell us whether there is implementation overhead to fix or whether the selection gate should be tightened so Aho-Corasick does not displace a faster fallback for these dictionary shapes. If the new tier cannot demonstrate a benefit after that, limiting or dropping it may be simpler than accepting the regressions. The corrected implementation is functionally sound, but I don't think the available performance results justify merging the additional accelerator machinery yet.

cushon:
Thanks for the review and catching those issues, and bearing with me while I try to make AC happen :)

I agree this is a dead end. I think alternations are well optimized by Teddy for <32 alternations, and prefix factoring (#783) for alternations with a common prefix. Given that the Lazy DFA is outperforming AC for the remaining cases it might have been helpful, I'll close this out.

PR #798, opened 2026-09-03
Treat comments-mode character class trivia uniformly

### LINKED ISSUE #796: Divergence on Pattern.compile - PatternSyntaxException: Unmatched closing ')'
I found this fuzzing, not in real world code. I think it may be a `java.util.regex` issue, if it is I can repro it there instead.

```java
class Repro {
    String pattern = "(?x)[a&&& [b])]";
    org.safere.Pattern.compile(pattern);
    try {
      java.util.regex.Pattern.compile(pattern);
    } catch (Exception e) {
      e.printStackTrace();
    }
  }
}
```

cushon:
Remove the `skippedCommentsTrivia` workarounds from `Parser` so SafeRE treats comments-mode whitespace and comments uniformly as ignored trivia everywhere within character classes. Document the intentional divergence in INTENTIONAL_DIVERGENCES.md.

Fixes #796


eaftan:
#801 provides a cleaner solution to the same underlying problem, and I’ve approved it. I think we can close this PR as superseded by #801.

PR #799, opened 2026-09-03
Implement SIMD Gap Guard-Byte Rejection ("Anti-Anchors") and Bounded Gap Scanning

cushon:
Many common regular expressions contain negated character classes acting as delimiters or boundary barriers between literal anchors (e.g., `header:[^\r\n;]*val`, `"[^"\\]*"`, or `foo[^\n]*bar`). Previously, the multi-anchor engine either failed to recognize these gaps or relied on scalar character-by-character checks during candidate verification, missing opportunities for early candidate rejection.

This PR introduces Guard-Byte Rejection ("Anti-Anchors"):

1. Extracts forbidden ASCII barrier bytes (guard bytes) from negated character class AST nodes and dot-quantifiers (`SINGLE_LINE_ANY_STAR` / `UNIX_LINES`).
2. Vectorizes delimiter and newline detection using AVX-512 / AVX2 / NEON byte vectors on UTF-8 inputs via `VectorEngine` and single-pass bounded scalar scans on Java `String`.
3. Employs guard bytes to accelerate forward gap termination (`scanClassEnd`), candidate slice validation (`matchesSlice`), trailing expansion (`expandTrailing`), and reverse driver searching (`findLastGuardByte`).
4. Replaces unbounded JDK search routines with strictly window-bounded loops, eliminating catastrophic $O(N \cdot K)$ scans when guard bytes are absent (codified as Invariant 4.31).

Measurements taken with `MultiAnchorGapEngineBenchmark`:

| Benchmark | `origin/main` | PR #799 | Speedup |
| :--- | :---: | :---: | :---: |
| `guardByteMatch_safere` (`header:[^\r\n;]*val`) | 17.82 µs | **15.30 µs** | **1.17x (+17%)** |
| `guardByteRejection_safere` (Absent guard bytes) | 3.52 µs | **3.55 µs** | **1.00x (No overhead)** |
| `fixedGapBaseline_safere` (`error:\[[A-Z]\] code:500`) | 11.00 µs | **11.00 µs** | **1.00x (No regression)** |


eaftan (commented):
I found several correctness and scaling problems in the guarded-gap executor. Greedy gaps could choose an anchor after the first forbidden byte and miss a valid earlier match; bounded gaps counted UTF-16/UTF-8 storage units instead of Unicode code points; reverse pruning could discard an upstream anchor containing the guard, including multibyte UTF-8 anchors; and guarded chains with a fallible continuation could require anchor retry that this executor does not provide. Overlapping driver candidates could also rescan the same delimiter or validated slice quadratically.

I've pushed c7d5f3c5, which keeps guarded execution on deterministic shapes, uses code-point-safe bounds, makes reverse overlap conservative, and reuses monotonic scan and validation state. It adds systematic tests for these cases, including work-counter and fuzz coverage.

Benchmark question: What benchmark did you see the claimed 17% improvement on? The PR describes `MultiAnchorGapEngineBenchmark` with `header:[^\r\n;]*val`, but I don't see that benchmark or workload on main or in this PR.  Could you add the workload that produced the 17% result to `benchmark-data.json` so that I can reproduce the result?

cushon:
Thanks for the review and those fixes!

I had added some benchmarks in #806 covering the series, but I think the 17% was on an ad-hoc benchmark that didn't make it into that set.

I took another look and found some more improvements:

* Bounded `ExactLiteralAnchor.lastIndexOf`: replaced unbounded `String.lastIndexOf(String, maxStart)` with `String.indexOf(literal, fromIndex, endBound)`
* Cascaded Bounds in findFirstGuardByte: When multiple guard bytes are present calling `text.indexOf` unconditionally for each byte meant that absent bytes scanned to the end of the input on every candidate. We cascaded the search bounds so a match narrows the search window for subsequent guard bytes.
* Guard Byte Ordering via `RarityOracle`: In `extractGuardBytes`, guard bytes are now sorted in ascending order of rarity (most frequent byte first) using `RarityOracle`.
* Bounded `findLastGuardByte`: Replaced unbounded `text.lastIndexOf` calls in `findLastGuardByte` with an unrolled backward scalar scan bounded by minLimit (matching the contract in `Utf8InputScanner`), preventing backward scans to 0 on upstream gaps.

The new results are below. The `match` overhead should be addressed by #804 #825.

```
./safere-benchmarks/scripts/compare-branch.sh --baseline gap-engine-benchmark-baseline '(guardByteHeader|guardByteDelimitedRecord).*safere-string'
```

| Benchmark                               | baseline (ns/op) | current (ns/op) | Speedup |
| --------------------------------------- | ---------------- | --------------- | ------- |
| guardByteHeader.match.1000              |        447 ± 9.0 |        486 ± 50 |   0.92x |
| guardByteHeader.match.10000             |        2296 ± 49 |      3235 ± 174 |   0.71x |
| guardByteHeader.match.100000            |      20855 ± 820 |    30746 ± 2463 |   0.68x |
| guardByteHeader.noMatch.1000            |        131 ± 5.3 |        131 ± 15 |   1.00x |
| guardByteHeader.noMatch.10000           |       1087 ± 117 |       1034 ± 25 |   1.05x |
| guardByteHeader.noMatch.100000          |        9831 ± 40 |      9900 ± 151 |   0.99x |
| guardByteDelimitedRecord.match.1000     |        3744 ± 83 |       1785 ± 67 |   2.10x |
| guardByteDelimitedRecord.match.10000    |     48078 ± 3014 |     16922 ± 427 |   2.84x |
| guardByteDelimitedRecord.match.100000   |   476702 ± 16279 |   166783 ± 4344 |   2.86x |
| guardByteDelimitedRecord.noMatch.1000   |         133 ± 11 |        139 ± 45 |   0.96x |
| guardByteDelimitedRecord.noMatch.10000  |        1034 ± 38 |       1050 ± 48 |   0.98x |
| guardByteDelimitedRecord.noMatch.100000 |        9818 ± 40 |      9908 ± 118 |   0.99x |

```
./safere-benchmarks/scripts/compare-branch.sh --baseline gap-engine-benchmark-baseline '(guardByteHeader|guardByteDelimitedRecord).*safere-utf8'
```

| Benchmark                               | baseline (ns/op) | current (ns/op) | Speedup |
| --------------------------------------- | ---------------- | --------------- | ------- |
| guardByteHeader.match.1000              |         496 ± 16 |        504 ± 11 |   0.98x |
| guardByteHeader.match.10000             |        3533 ± 75 |       3525 ± 83 |   1.00x |
| guardByteHeader.match.100000            |     32134 ± 2246 |     33828 ± 483 |   0.95x |
| guardByteHeader.noMatch.1000            |       187 ± 0.77 |       188 ± 3.1 |   0.99x |
| guardByteHeader.noMatch.10000           |        1564 ± 29 |      1565 ± 113 |   1.00x |
| guardByteHeader.noMatch.100000          |      15886 ± 118 |     16080 ± 328 |   0.99x |
| guardByteDelimitedRecord.match.1000     |        3737 ± 57 |       2175 ± 11 |   1.72x |
| guardByteDelimitedRecord.match.10000    |      49828 ± 628 |    32478 ± 1008 |   1.53x |
| guardByteDelimitedRecord.match.100000   |   539440 ± 24401 |   342002 ± 3938 |   1.58x |
| guardByteDelimitedRecord.noMatch.1000   |        179 ± 1.5 |       179 ± 1.9 |   1.00x |
| guardByteDelimitedRecord.noMatch.10000  |        1535 ± 33 |       1543 ± 77 |   1.00x |
| guardByteDelimitedRecord.noMatch.100000 |      16160 ± 488 |     15970 ± 361 |   1.01x |

eaftan (approved):
I found and fixed several correctness and scaling bugs in the guarded-gap executor. Greedy gaps could skip valid matches by choosing an anchor beyond the first forbidden byte; bounded gaps counted storage units instead of Unicode code points; reverse pruning could discard an upstream anchor containing the guard; and fallible continuations required retries this executor does not provide. Reluctant searches and overlapping candidates could rescan the same suffix or validated slice, and multi-guard String searches repeated full-window scans. Invalid UTF-16 boundaries could also skip candidates or split surrogate pairs, while unpaired surrogates were unsafe in the UTF-8 path.

I've pushed fixes that keep guarded execution on deterministic, encoding-safe shapes, enforce code-point and surrogate-safe bounds, make reverse pruning conservative, and reuse monotonic scan and validation state. They add regression tests for these cases, including scaling and fuzz coverage.

I also looked more closely at the header-match performance regression. Profiling and code tracing showed that `Matcher` checked for a required suffix, then `MultiAnchorExecutor` searched for that suffix again before locating its driver anchor. I've pushed `be8548c7`, which removes the executor's preliminary rejection scan in both input representations. The executor continues to verify every anchor and gap. This recovers the String header-match performance without weakening the correctness fixes or adding pattern-specific logic.

Here is the full String comparison against main 92822128, using the PR's unchanged benchmark definitions. “Header” is `guardByteHeader`; “Delimited record” is `guardByteDelimitedRecord`. The before and after columns both include the correctness fixes; they isolate the performance change. These are long-mode measurements, and lower ns/op is better.

| String workload | Main ns/op | PR before perf fix ns/op | PR after perf fix ns/op | Interpretation |
|---|---:|---:|---:|---|
| Header, match, 1,000 bytes | 300 | 351 | 253 | Header regression recovered |
| Header, match, 10,000 bytes | 1,782 | 2,522 | 1,728 | Header regression recovered |
| Header, match, 100,000 bytes | 16,428 | 24,276 | 16,351 | Header regression recovered |
| Header, no match, 1,000 bytes | 96 | 97 | 96 | Neutral |
| Header, no match, 10,000 bytes | 811 | 813 | 810 | Neutral |
| Header, no match, 100,000 bytes | 7,943 | 7,948 | 7,992 | Neutral |
| Delimited record, match, 1,000 bytes | 2,218 | 1,097 | 1,081 | Record-matching gain preserved |
| Delimited record, match, 10,000 bytes | 28,230 | 10,228 | 9,248 | Record-matching gain preserved |
| Delimited record, match, 100,000 bytes | 315,613 ± 59,289 | 101,884 | 92,362 | Record-matching gain preserved |
| Delimited record, no match, 1,000 bytes | 96 | 96 | 96 | Neutral |
| Delimited record, no match, 10,000 bytes | 811 | 809 | 810 | Neutral |
| Delimited record, no match, 100,000 bytes | 7,981 | 8,018 | 7,938 | Neutral |

The largest main delimited-record measurement was noisy; its cell includes the confidence-interval half-width. The header regression is recovered, the delimited-record benefit is preserved, and no-match results are effectively unchanged. These results support the optimization for the measured workloads rather than a general speedup for every pattern.

For completeness, here are the standard-mode UTF-8 before/after results. These workloads call `Pattern.find(Utf8Input)`, which uses a separate boolean search path, so they do not measure the modified executor. Lower ns/op is better.

| UTF-8 workload | PR before perf fix ns/op | PR after perf fix ns/op | Interpretation |
|---|---:|---:|---|
| Header, match, 1,000 bytes | 390 | 387 | Broadly neutral |
| Header, match, 10,000 bytes | 2,590 | 2,589 | Broadly neutral |
| Header, match, 100,000 bytes | 24,199 | 24,330 | Broadly neutral |
| Header, no match, 1,000 bytes | 136 | 140 | Broadly neutral |
| Header, no match, 10,000 bytes | 1,189 | 1,183 | Broadly neutral |
| Header, no match, 100,000 bytes | 11,620 | 11,591 | Broadly neutral |
| Delimited record, match, 1,000 bytes | 2,282 | 2,268 | Broadly neutral |
| Delimited record, match, 10,000 bytes | 32,392 | 32,431 | Broadly neutral |
| Delimited record, match, 100,000 bytes | 324,885 | 326,487 | Broadly neutral |
| Delimited record, no match, 1,000 bytes | 135 | 135 | Broadly neutral |
| Delimited record, no match, 10,000 bytes | 1,163 | 1,170 | Broadly neutral |
| Delimited record, no match, 100,000 bytes | 12,274 | 11,638 | Broadly neutral |

LGTM

eaftan:
PR Shepherd is now monitoring this PR at `be8548c75e965192e63415526328f39b7bf1d2a2`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `be8548c75e965192e63415526328f39b7bf1d2a2`.

PR #805, opened 2026-09-07
Support variable gap chains and wildcards in multi-anchor engine

cushon:
Extends SafeRE's multi-anchor acceleration engine to support chains containing multiple variable-width, wildcard, and quantified character-class gaps (`ANY_STAR`, `SINGLE_LINE_ANY_STAR`, and `BOUNDED_CLASS_REPEAT`).

Previously, multi-anchor execution was restricted to fixed-width internal gaps or single bounded repetitions under narrow constraints; patterns with multiple variable or wildcard gaps (e.g. `foo.*bar`, `foo[^\n]*bar[^\n]*baz`, `header:.*?\r\n`) fell back to the standard forward DFA. This change enables SIMD multi-anchor acceleration across chains of arbitrary variable gaps and character-class repetitions for both String and UTF-8 inputs, delivering up to **8.0x** throughput improvements on real-world log and structured text extraction workloads.

---

1. **Downstream Recursive Matching (`MultiAnchorExecutor`)**:
   - Replaced flat single-gap execution with call-stack recursion over chain segments (`matchDownstream`) to support multi-gap chains without heap allocations, candidate collections, or queue wrappers.
   - **Zero heap allocations**: All per-search allocations have been eliminated. State progression is tracked entirely through primitives and registers.
   - **Stack-safety**: Recursion depth is strictly bounded by the number of literal segments in the pattern ($M = \text{segments.length}$, typically 2–5), while candidate positions in the text are traversed iteratively via `while` loops within each frame.
   - **Greedy gaps**: Evaluates candidate anchors backwards from `maxHop` down to `minHop` using `anchor.lastIndexOf`.
   - **Reluctant gaps**: Evaluates candidate anchors forwards from `minHop` up to `maxHop` using `anchor.findNextWithin`.
   - **Fixed-width gaps**: Evaluated in $O(1)$ without candidate iteration.

2. **Linear-Time Guarantees & Zero-Allocation Watermarking**:
   - **Bitwise-Encoded Watermark**: Propagates watermark progress back up the recursion stack on match failure using bitwise complement (`~newWatermark`) without allocating heap wrappers or state objects. When downstream verification fails, subsequent candidate searches do not re-scan text already explored during prior attempts.
   - **Direct SIMD Vector Scanning**: Reluctant and unbounded gaps invoke SIMD vectorized scanners directly (`guardedSearchEnd` / `scanClassEnd`) without per-invocation array allocations.
   - **Upstream Bounds**: Maintains reverse driver search bounds in `scanClassStart` and `findUpstreamMatch` to guarantee linear execution when multiple candidate positions are evaluated.

3. **$O(1)$ Verification for Unbounded Gaps**:
   - For unbounded gaps (`maxLength == Integer.MAX_VALUE`, e.g. `.*` or `[^\n]*`), `scanClassEnd` and `scanClassStart` already guarantee that all characters in the scanned window belong to the gap class.
   - Added `isUnboundedGapSatisfied` (UTF-16) and `isUnboundedGapSatisfiedUtf8` (UTF-8) predicates to bypass redundant $O(N)$ slice rescanning and code point counting on candidate slices, eliminating multi-pass overhead on large (100KB+) inputs.
   - Preserved strict $O(1)$ surrogate-pair boundary checks (`gap.endsAtCodePointBoundary`) in UTF-16 matching so candidate anchors never split surrogate pairs.

4. **Unicode & JDK Equivalence (`GapScanner`)**:
   - Accurately counts full Unicode code points on bounded gaps (`maxLength < Integer.MAX_VALUE`) across UTF-16 surrogate pairs and UTF-8 multi-byte sequences.
   - Distinguishes standard dot (rejecting `\r`, `\n`, `\u0085`, `\u2028`, and `\u2029`) from `UNIX_LINES` dot (rejecting only `\n`).
   - Short-circuits `SINGLE_LINE_ANY_STAR` in `GapScanner.matchesSlice` when `maxLength == Integer.MAX_VALUE && minLength == 0` once guard bytes pass.
   - Skips code point counting in `expandTrailing` when `minLength == 0`.

5. **Compiler Updates (`MultiAnchorDescriptor`, `MultiAnchorCompiler`)**:
   - Broadened multi-anchor chain eligibility to accept multiple variable and wildcard gaps while enforcing guard-byte and driver selectivity constraints.

---

```
./safere-benchmarks/scripts/compare-branch.sh   --baseline origin/main   --current variable-gap-chains   'RealWorldRegexBenchmark\.runBenchmark\.(multiInfixLog|multiClauseSequence|structuredJsonPath).*@safere-string'
```

| Benchmark                            | baseline (ns/op) | current (ns/op) | Speedup |
| ------------------------------------ | ---------------- | --------------- | ------- |
| multiInfixLog.match.1000             |        4242 ± 46 |       1642 ± 44 |   2.58x |
| multiInfixLog.match.10000            |      63341 ± 814 |     13426 ± 208 |   4.72x |
| multiInfixLog.match.100000           |   499614 ± 25280 |   134985 ± 1156 |   3.70x |
| multiInfixLog.noMatch.1000           |       146 ± 0.79 |       148 ± 5.5 |   0.99x |
| multiInfixLog.noMatch.10000          |       1113 ± 9.1 |      1115 ± 8.5 |   1.00x |
| multiInfixLog.noMatch.100000         |       10734 ± 55 |      10758 ± 67 |   1.00x |
| multiInfixLogCaptures.match.1000     |        6284 ± 39 |       6299 ± 71 |   1.00x |
| multiInfixLogCaptures.match.10000    |     64782 ± 1181 |    62364 ± 4676 |   1.04x |
| multiInfixLogCaptures.match.100000   |   515726 ± 48516 |  524680 ± 50739 |   0.98x |
| multiInfixLogCaptures.noMatch.1000   |        4220 ± 31 |       4303 ± 87 |   0.98x |
| multiInfixLogCaptures.noMatch.10000  |      48790 ± 912 |    48929 ± 1005 |   1.00x |
| multiInfixLogCaptures.noMatch.100000 |    475365 ± 2099 |  528453 ± 10519 |   0.90x |
| multiClauseSequence.match.1000       |        2089 ± 24 |       2206 ± 29 |   0.95x |
| multiClauseSequence.match.10000      |      18537 ± 500 |     19700 ± 241 |   0.94x |
| multiClauseSequence.match.100000     |    174329 ± 1277 |   217215 ± 2752 |   0.80x |
| multiClauseSequence.noMatch.1000     |       130 ± 0.88 |       130 ± 1.4 |   1.00x |
| multiClauseSequence.noMatch.10000    |        1040 ± 13 |       1055 ± 44 |   0.99x |
| multiClauseSequence.noMatch.100000   |      10052 ± 286 |     10029 ± 217 |   1.00x |
| structuredJsonPath.match.1000        |       7535 ± 158 |       1290 ± 26 |   5.84x |
| structuredJsonPath.match.10000       |     78501 ± 3825 |      11224 ± 83 |   6.99x |
| structuredJsonPath.match.100000      |   892826 ± 78028 |   111797 ± 2139 |   7.99x |
| structuredJsonPath.noMatch.1000      |        184 ± 1.5 |       185 ± 7.0 |   0.99x |
| structuredJsonPath.noMatch.10000     |        1437 ± 16 |       1440 ± 15 |   1.00x |
| structuredJsonPath.noMatch.100000    |      13949 ± 238 |     14030 ± 177 |   0.99x |


safere/src/main/java/org/safere/MultiAnchorExecutor.java
```diff
@@ -583,137 +579,296 @@ static Result find(MultiAnchorDescriptor descriptor, String text, int searchFrom
...
-              ? trailingGap.matchExecutorFixedForward(text, currentPos, textLen)
-              : trailingGap.expandTrailing(text, currentPos, textLen);
-      if (matchEnd < 0) {
-        candidatePos = advanceCandidatePos(candidatePos, pDriver, minUpstreamLen);
-        continue;
+      int anchorLen = anchor.lengthAt(text, p);
+      if (anchorLen <= 0) {
+        return -1;
       }
+      return matchDownstream(text, segments, i + 1, p + anchorLen, textLen, trailingGap, 0, -1);
```
eaftan: This recursion makes stack depth depend on the number of segments in the pattern. I reproduced StackOverflowError with -Xss256k through both public String and UTF-8 matchers:

```java
String regex = "AAA" + ".*?BBB".repeat(2000) + ".*?CCC";
String input = "AAA" + "BBB".repeat(2000) + "CCC";
Pattern.compile(regex).matcher(input).find();
```

The descriptor accepts all 2,002 segments. The fixed-gap and variable-gap branches both recurse, and the UTF-8 implementation has the same problem. Could we use an explicit stack or worklist here so that longer chains don’t exhaust the Java stack?

safere/src/main/java/org/safere/MultiAnchorExecutor.java
```diff
@@ -583,137 +579,296 @@ static Result find(MultiAnchorDescriptor descriptor, String text, int searchFrom
...
+          }
+        }
+        if (gap.endsAtCodePointBoundary(text, p)
+            && ((gap.maxLength() == Integer.MAX_VALUE
+                    && isUnboundedGapSatisfied(gap, p - currentPos))
+                || gap.matchesSlice(text, currentPos, p))) {
+          int anchorLen = anchor.lengthAt(text, p);
+          if (anchorLen > 0) {
+            int matchEnd =
+                matchDownstream(text, segments, i + 1, p + anchorLen, textLen, trailingGap, 0, -1);
```
eaftan: Each candidate searches the suffix again from scratch, resetting its watermark and cached guard end. The caller discards failed suffix progress, and this recursive work isn't charged to the outer `WorkLimit`.

I tested `RAREST_TOKEN.*?BBB.*?CCC` against `"RAREST_TOKEN" + "xBBB".repeat(n) + "\nCCC"`. At 1K, 2K, 4K, and 8K candidates, the wildcard loops perform 2,006,000, 8,012,000, 32,024,000, and 128,048,000 scan iterations -- roughly four times the work whenever input doubles. This happens through both public String and UTF-8 matchers, with greedy gaps too. Adding another variable gap makes the growth approximately eightfold per doubling.

Could we retain suffix progress across retries, or fall back to the general engine when the work budget is exhausted? I think this needs a separate change from replacing the recursion, since an iterative version could still repeat the same suffix searches.

eaftan (changes requested):
Thanks for working on this. The broader gap support looks useful. I found two cases where the new traversal can overflow the stack or do nonlinear work, and left examples inline. Could you PTAL?

safere/src/main/java/org/safere/MultiAnchorExecutor.java
cushon: Thanks for the catch, I pushed a commit to replaced the recursion.

safere/src/main/java/org/safere/MultiAnchorExecutor.java
cushon: Done, I updated the `WorkLimit` accounting and also added watermarks to avoid re-scanning.

safere/src/main/java/org/safere/MultiAnchorExecutor.java
```diff
@@ -583,137 +670,637 @@ static Result find(MultiAnchorDescriptor descriptor, String text, int searchFrom
...
+                : trailingGap.expandTrailing(scanner, p + anchorLen, textLen);
+        if (matchEnd >= 0) {
+          return matchEnd;
+        }
+        backtrack = true;
+        continue;
+      }
+
+      s++;
+      state[s * STATE_STRIDE + OFFSET_CUR_POS] = p + anchorLen;
```
eaftan, line 1026: When upstream backtracking changes this start position, the descendant can still retain a guard endpoint and reluctant-search watermark from its previous start. I found cases where that either matches across a forbidden delimiter or skips a valid match. Could we invalidate those cached values when the start changes? The String path has the same issue.

safere/src/main/java/org/safere/GapScanner.java
```diff
@@ -204,40 +204,46 @@ static int scanClassEnd(Gap gap, String text, int fromPos, int maxPos) {
...
           break;
         }
+        count++;
         cur += Character.charCount(cp);
       }
       return cur;
     }
     if (gap.maxLength() != Integer.MAX_VALUE) {
-      return Math.min(maxPos, fromPos + gap.maxLength());
+      return (int) Math.min(maxPos, fromPos + (long) gap.maxLength() * 2);
```
eaftan: Widening the search window accounts for surrogate pairs, but the later ANY_STAR slice check still compares the quantifier bound with to - from—UTF-16 units for String, and bytes in the UTF-8 path. A supplementary character can therefore give thewrong result for a bounded dot gap. Could the final slice check count code points in both paths?

safere/src/main/java/org/safere/MultiAnchorDescriptor.java
```diff
@@ -460,35 +455,31 @@ private static boolean isExecutableAnchor(Anchor anchor) {
...
-      case BOUNDED_CLASS_REPEAT -> gap.isExecutorFixedGap();
-      case ANY_STAR,
-          SINGLE_LINE_ANY_STAR,
-          TEXT_END,
-          LINE_START,
-          LINE_END,
-          WORD_BOUNDARY,
-          NO_WORD_BOUNDARY ->
-          false;
+      case EMPTY, TEXT_START, ANY_STAR, SINGLE_LINE_ANY_STAR -> true;
```
eaftan, line 458: I found that admitting a variable leading gap here can change which match wins when its greedy or reluctant priority interacts with later anchors. For example, `.*AAA.*?BBB` on `AAABBB AAABBB` returns two matches here, whereas JDK returns the whole input as one match. The draft fix routes these shapes through the general engine for now; perhaps there’s a narrower accelerated subset we can identify.

safere/src/main/java/org/safere/MultiAnchorExecutor.java
```diff
@@ -729,4 +1316,26 @@ private static int findNextCountingWork(
...
+  private static boolean isUnboundedGapSatisfiedUtf8(MultiAnchorDescriptor.Gap gap, int len) {
+    if (gap.minLength() <= 1) {
+      return len >= gap.minLength();
+    }
+    return len >= gap.minLength() * 4;
+  }
+
+  private static boolean gapScansText(MultiAnchorDescriptor.Gap gap, boolean reluctantGuarded) {
+    if (reluctantGuarded) {
+      return gap.maxLength() != Integer.MAX_VALUE;
```
eaftan: An unbounded reluctant guard scan can be repeated after upstream backtracking, but this excludes its scan cost from the work budget. I also found repeated fixed, reverse, leading, and trailing checks that need accounting. Could we charge those retries and fall back when the budget is exhausted? That would let us retain the repaired state semantics without risking nonlinear attempted work.

safere/src/main/java/org/safere/MultiAnchorExecutor.java
```diff
@@ -501,14 +558,14 @@ static Result find(MultiAnchorDescriptor descriptor, String text, int searchFrom
...
           int searchUpperBound = curAnchorStart - minHop;
           int searchLowerBound = Math.max(minReverseWatermark, curAnchorStart - maxHop);
-          if (gap.guardBytes() != null) {
-            int lastGuard = gap.findLastGuardByte(text, searchLowerBound, curAnchorStart - 1);
-            if (lastGuard >= 0) {
-              int firstOverlappingAnchorStart =
-                  (int) Math.max(0L, (long) lastGuard - upstreamAnchor.maxLength() + 1L);
-              searchLowerBound = Math.max(searchLowerBound, firstOverlappingAnchorStart);
-            }
+          int earliestGapStart = gap.scanClassStart(text, searchLowerBound, curAnchorStart);
```
eaftan, line 561: his reverse class scan can be repeated for successive driver candidates, but its cost isn’t included in workHolder. For `AAA[AB]{64}BBB` on `"AAAX" + "B".repeat(128)`, the executor completes without exhausting the budget; charging the repeated scans causes it to fall back. Could we include this work and the repeated slice checks in the budget? #852 adds that accounting in both input paths.

eaftan (changes requested):
Thanks for continuing to work through this. I found the matching and work-budget concerns inline and put a possible repair in draft #852, targeting this branch. The demonstrated matching failures return the expected results with that patch for both String and UTF-8. The draft is there for you to inspect or adapt.

There is a significant performance tradeoff, though. I compared the repaired source with main using the same workload definitions on both revisions and the standard JMH schedule. Ratios below are repaired time / main time, geometric means over equally weighted 1K, 10K, and 100K rows. Values near 1 are effectively parity. The UTF-8 plan does not include the capture variant.

| API | Workload | Rows | Repaired/main |
|---|---|---:|---:|
| String | multiInfixLog match | 3 | 0.991 |
| String | multiInfixLogCaptures match | 3 | 1.000 |
| String | multiClauseSequence match | 3 | 2.134 |
| String | structuredJsonPath match | 3 | 0.987 |
| String | Four no-match families | 12 | 1.008 |
| UTF-8 | multiInfixLog match | 3 | 1.020 |
| UTF-8 | multiClauseSequence match | 3 | 1.003 |
| UTF-8 | structuredJsonPath match | 3 | 0.985 |
| UTF-8 | Three no-match families | 9 | 1.007 |

A longer run confirmed the important 100K String result: `multiClauseSequence` took 275,978 ± 6,173 ns/op with the repair versus 121,912 ± 1,364 on main -- 2.264× as long. The other tested String and UTF-8 100K matches were near parity with main.

To separate the repair’s cost from the submitted PR’s, I also compared the two directly in standard mode on matching 100K String inputs:

| Workload | Submitted head ns/op | Repaired ns/op | Repaired/submitted |
|---|---:|---:|---:|
| multiInfixLog | 109,766 ± 3,052 | 337,349 ± 8,999 | 3.073 |
| multiClauseSequence | 273,798 ± 2,334 | 274,408 ± 4,079 | 1.002 |
| structuredJsonPath | 85,639 ± 910 | 561,255 ± 44,266 | 6.554 |

The safer leading-gap handling removes two of the headline String speedups. The `multiClauseSequence` slowdown relative to main was already present before my repair; I haven’t profiled that path, so I don’t yet know its specific cause.

I’d be hesitant to merge #805 as a performance optimization with these results. Could you look at the draft and decide whether there’s a way to recover the benefit while preserving match semantics, or whether the accelerated scope should be narrower? The higher PRs will also need revalidation once #805’s direction is settled.

cushon:
Thanks for the review and finding those issues.

I am rethinking the approach here a bit.

I think the multi-anchor stuff makes sense as a general approach to  prefilters and start/end acceleration.

I think the part that was trying to characterize kinds of gaps, and the `MultiAnchorExecutor`, was flawed.

I want to try paring back the gap analysis and `MultiAnchorExecutor` stuff, and keeping multi-anchor.

Something close to regex-automata's reverse-inner would get most of the same benefits and cover some patterns the gap executor doesn't, but without the complexity of trying to handle arbitrary interior gaps and the surface area of another executor.

Also most of the benchmarks where the gap executor is winning were added  in PRs that implemented the gap engine, partly with the goal of confirming it was actually working. But benchmarking with the gap executor disabled shows it doesn't apply to very many cases, and it's a wash on the cases that predate it. One of the ones where it wins (fixedAnchorLog) looks like an anchor selection issue more than a case where the gap executor is really needed.

I'm going to move this and #806/#831/#835 back to draft.

#845 and #847 don't depend on the executor, so I'll restack those.

PR #771, opened 2026-08-26
Unify AST analysis with MultiAnchorCompiler and sealed MultiAnchorDescriptor

cushon:
This change consolidates compile-time pattern analysis and acceleration metadata in SafeRE into a unified architecture.

* Unified bounded-linear AST analysis in `MultiAnchorCompiler` replaces the separate AST extractors in `Pattern` (`extractStartDescriptor`, `extractRejectDescriptor`, `extractDisjointLiterals`, `extractLeadingExpansion`)
* Introduces `MultiAnchorDescriptor` with `StartPlan` (`None`, `Literal`, `CharClass`, `FixedOffset`, `MultiLiteral`, `LeadingExpansion`, `LineAnchor`) and `RejectPlan` (`None`, `RequiredLiteral`, `RequiredCharClass`, `DisjointLiterals`, `EndAnchoredSuffix`, `EndAnchoredCharClass`, `Composite`)
* Migrates `StringStartAccelerator`, `Utf8StartAccelerator`, and `RejectPrefilter` to evaluate plans.
* Removes `StartDescriptor.java` and `RejectDescriptor.java`

Part of #769.

PR #807, opened 2026-09-07
Subsume StartPlan.FixedOffset literal in RejectPlan extraction

cushon:
When extracting prefilter plans in `MultiAnchorCompiler`, `StartPlan.Literal` has always had its prefix excluded from `RejectPlan.RequiredLiteral` (`extractRequiredLiteral(metadataAst, prefix, suffixStr)`). This prevents SafeRE from running an uncoordinated existence probe in `RejectPrefilter` for the exact same literal that `StartPlan.Literal` is already scanning for.

However, when `StartPlan.FixedOffset` was introduced in PR #771, `extractRequiredLiteral` was not updated to exclude `fol.literal()`. As a result, for all fixed-offset patterns without a leading prefix literal (e.g. `"[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z error"` or `"[0-9]{2}404_NOT_FOUND"`), `MultiAnchorCompiler` produced both:
- `StartPlan.FixedOffset` (which scans for the literal to locate candidate starts)
- `RejectPlan.RequiredLiteral` (which scans for the same literal across the entire document in `RejectPrefilter`)

On matching inputs, this forced SafeRE to scan the entire input text twice for the exact same literal string: first in `RejectPrefilter`, which threw the found offset away upon confirming existence, and then again in `StringStartAccelerator.FixedOffset` / `Utf8StartAccelerator.FixedOffset`.

```
./safere-benchmarks/scripts/compare-branch.sh     --baseline origin/main     --current fixed-offset-reject-subsumption     'isoTimestampLog.*@safere-string'
```

| Benchmark                      | baseline (ns/op) | current (ns/op) | Speedup |
| ------------------------------ | ---------------- | --------------- | ------- |
| isoTimestampLog.match.1000     |         599 ± 18 |        458 ± 13 |   1.31x |
| isoTimestampLog.match.10000    |       3331 ± 114 |       2224 ± 14 |   1.50x |
| isoTimestampLog.match.100000   |      29846 ± 494 |     19923 ± 351 |   1.50x |
| isoTimestampLog.noMatch.1000   |        134 ± 5.3 |       175 ± 7.3 |   0.77x |
| isoTimestampLog.noMatch.10000  |        1041 ± 48 |       1065 ± 41 |   0.98x |
| isoTimestampLog.noMatch.100000 |       9948 ± 299 |       9866 ± 27 |   1.01x |

### DIFF
--- a/safere/src/main/java/org/safere/MultiAnchorCompiler.java
+++ b/safere/src/main/java/org/safere/MultiAnchorCompiler.java
@@ -311,8 +311,6 @@ final class MultiAnchorCompiler {
             : startPlan instanceof MultiAnchorDescriptor.StartPlan.FixedOffset fo
                 ? fo.fol().literal()
                 : null;
-    String prefix =
-        startPlan instanceof MultiAnchorDescriptor.StartPlan.Literal lit ? lit.prefix() : null;
     CharClassScanInfo ccPrefix =
         startPlan instanceof MultiAnchorDescriptor.StartPlan.CharClass cc ? cc.scanInfo() : null;
     boolean hasLeadingExpansion =
@@ -328,7 +326,7 @@ final class MultiAnchorCompiler {
     }
 
     CharClassScanInfo requiredMatchClass = null;
-    if (!anchorStart && prefix == null && endAnchoredCharClass == null) {
+    if (!anchorStart && excludeStartLiteral == null && endAnchoredCharClass == null) {
       CharClass reqClass = reject.bestRequiredClass();
       if (reqClass != null) {
         if (ccPrefix == null) {
@@ -356,7 +354,7 @@ final class MultiAnchorCompiler {
     }
 
     String[] disjointLiterals =
-        (!anchorStart && prefix == null && requiredLiteral == null)
+        (!anchorStart && excludeStartLiteral == null && requiredLiteral == null)
             ? reject.disjointRequiredLiterals()
             : null;
     if (disjointLiterals != null && disjointLiterals.length > 1) {
--- a/safere/src/main/java/org/safere/MultiAnchorCompiler.java
+++ b/safere/src/main/java/org/safere/MultiAnchorCompiler.java
@@ -305,6 +305,12 @@ final class MultiAnchorCompiler {
       plans.add(new MultiAnchorDescriptor.RejectPlan.EndAnchoredCharClass(endAnchoredCharClass));
     }
 
+    String excludeStartLiteral =
+        startPlan instanceof MultiAnchorDescriptor.StartPlan.Literal lit
+            ? lit.prefix()
+            : startPlan instanceof MultiAnchorDescriptor.StartPlan.FixedOffset fo
+                ? fo.fol().literal()
+                : null;
     String prefix =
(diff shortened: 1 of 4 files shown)


eaftan (approved):
I found a few bugs in the fixed-prefix collector. It recursively walked the AST and called exact-literal extraction at every nested node, so sufficiently deep patterns could throw `StackOverflowError` and nested literal structures could take quadratic compile work. It also collected literals from the first variable-width element after the fixed prefix, which could discard a useful independent reject token. Finally, the containment check treated a longer downstream token such as `zzRARE` as subsumed by a separate fixed-offset `zz`, even though finding the first `zz` says nothing about the later token.

I've pushed a commit that uses a single-pass explicit worklist, stops collection before the first variable-width element, and makes subsumption one-way. It adds tests covering 10,000 nested levels, the fixed/variable-width boundary, and repeated prefix text in a longer downstream token.

I was able to reproduce the reported benchmark results. `PR/main` is elapsed-time ratio, so lower is better.

| Workload | Main ns/op | PR + fixes ns/op | PR/main | Result |
|---|---:|---:|---:|---|
| match, 1,000 bytes | 431 | 330 | 0.766 | 1.31x faster |
| match, 10,000 bytes | 2,612 | 1,780 | 0.681 | 1.47x faster |
| match, 100,000 bytes | 24,033 | 16,279 | 0.677 | 1.48x faster |
| no match, 1,000 bytes | 97 | 135 | 1.392 | 39% slower, with high measurement error |
| no match, 10,000 bytes | 816 | 859 | 1.053 | 5% slower |
| no match, 100,000 bytes | 7,970 | 8,563 | 1.074 | 7% slower |

I dug into the no-match regression. Main's compiled plan uses both a fixed-offset `Z error` accelerator and an early required-literal prefilter for the same text. On these no-match inputs, that prefilter performs one `String.indexOf` and returns immediately. The PR also performs one absent-literal scan, but removing the prefilter lets execution enter the multi-anchor engine first, where it checks chain eligibility twice, inspects gaps and anchors, and selects a driver. Profiling showed that setup work, and a controlled diagnostic ablation restoring only the early rejection returned the 1,000-byte result to baseline:

| Diagnostic revision | ns/op |
|---|---:|
| Main | 95.585 ± 1.705 |
| PR + fixes | 125.179 ± 3.788 |
| PR + fixes with early rejection restored | 92.855 ± 3.370 |

The diagnostic runs used no-fork mode for profiler attachment; the standard forked results above remain the performance evidence for the review. I filed #825 to investigate reducing the repeated per-search multi-anchor setup without restoring the duplicate scan when the literal is present.

The matching gains justify the bounded short no-match tradeoff, and the remaining setup-cost opportunity is separable from this change. LGTM.

eaftan:
PR Shepherd is now monitoring this PR at `c5f6d69728f00f4b919425be7eec92f96ccf0936`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `c5f6d69728f00f4b919425be7eec92f96ccf0936`.

Issue #52, opened 2026-03-30
Suspected JDK bug: capture groups leak across starting positions in (?:(a))*$ patterns


eaftan:
## Summary

JDK's `java.util.regex` engine appears to leak capture group values from **failed** match attempts at earlier starting positions when a capture group is inside a non-capturing group under repetition (e.g., `(?:(a))*$`).

This was discovered while investigating [#42 (bug 2)](https://github.com/eaftan/safere/issues/42).

## Reproduction

```java
import java.util.regex.*;

public class JdkCaptureLeak {
    public static void main(String[] args) {
        // Direct quantification: g1 correctly null (zero iterations)
        Matcher m1 = Pattern.compile("(a)*$").matcher("aaab");
        m1.find();
        System.out.println("(a)*$ on \"aaab\": g1=" + m1.group(1));
        // Output: g1=null  ✓

        // Wrapped in non-capturing group: g1 incorrectly leaks [2,3)
        Matcher m2 = Pattern.compile("(?:(a))*$").matcher("aaab");
        m2.find();
        System.out.println("(?:(a))*$ on \"aaab\": g1=" + m2.group(1));
        // Output: g1=a  (start=2, end=3)  ✗
        // This capture can only come from a FAILED starting position
    }
}
```

Tested on OpenJDK 25.

## Evidence

| Pattern | Text | JDK g1 | SafeRE g1 | RE2/J g1 | Correct |
|---------|------|--------|-----------|----------|---------|
| `(a)*$` | `"aaab"` | null | null | null | null |
| `(?:(a))*$` | `"aaab"` | [2,3) | null | null | null |
| `(?:(a))*$` | `"xyzb"` | null | null | null | null |
| `(?:(a))*$` | `"ab"` | [0,1) | null | null | null |
| `(?:(a))*$` | `"aaa"` | [2,3) | [2,3) | [2,3) | [2,3) |
| `(?:(?:(a)))*$` | `"aaab"` | [2,3) | null | — | null |

## Analysis

The match for `(?:(a))*$` on `"aaab"` succeeds at position 4 (end of text) with zero iterations of `*`. Since the repetition matched zero times, no `(a)` was ever successfully matched in the winning match — g1 should be null/-1.

However, JDK reports g1=[2,3), which is the character `a` at index 2. This value can only come from a **failed** match attempt at an earlier starting position (e.g., starting at pos 0 where `(?:(a))*` matched `aaa` but `$` failed because `b` follows).

Key observations:
[...]

PR #809, opened 2026-09-07
Support static Jazzer entry points in fuzz targets and filter failed-path capture leakage

cushon:
The goal of this change is to allow running the SafeRE fuzzers with Bazel (or Blaze).

This addes `public static fuzzerTestOneInput` entry points to the fuzzer classes and makes them final, allowing direct invocation from standalone Jazzer and Bazel, while preserving the current JUnit 5 `@FuzzTest` integration with Maven Surefire.

Also, this filters documented JDK failed-path capture leakage (#52) in `FuzzSupport` to prevent false-positive differential assertion failures during continuous random fuzzing.

### DIFF
--- a/safere-fuzz/pom.xml
+++ b/safere-fuzz/pom.xml
@@ -35,6 +35,12 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <version>3.27.7</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.safere</groupId>
       <artifactId>safere</artifactId>
       <version>${project.version}</version>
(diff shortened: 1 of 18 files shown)


eaftan (dismissed):
The static Jazzer entry points look good. I noticed that the capture filter used a broad regex-source shape without checking which group was quantified, and the replacement filter accepted any output difference for that shape even when there was no capture discrepancy. The filter also missed the known same-start leakage case `(?:(?:())*{0}{1}?{1}|a).` on `ab`.

I've pushed d8554909, which identifies captures inside quantifiers from the parsed pattern. When both engines agree on the full match's text and boundaries and SafeRE reports one of those groups completely unmatched while JDK reports it participating, we waive that group's discrepancy, including same-start leakage. We also retain the waiver for captures entirely before the accepted match. Numbered and named accessors and `MatchResult` use the same policy.

For replacements, an independent check of the relevant matches must encounter an actual waived capture difference before output comparison is skipped; `appendReplacement` checks the current match. The replacement API still runs, exceptions are still compared, and full matches and other capture differences are checked. The fix adds tests for the known earlier-start, same-start, and replacement leakage examples, as well as differences we continue to report.

This makes the waiver more precise than the previous regex-text heuristic while covering the known leakage examples. It can still miss real bugs that lose quantified captures, or replacement-output bugs when a waived capture difference occurs, but I think that’s a reasonable tradeoff to reduce false positives during fuzzing.

LGTM

eaftan (approved):

eaftan:
PR Shepherd is now monitoring this PR at `8492c5a5a280a341958c0673afa550895144217f`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `8492c5a5a280a341958c0673afa550895144217f`.

PR #811, opened 2026-09-07
Preserve find continuation after failed anchored matches

### LINKED ISSUE #808: Failed matches() resets Matcher search position to regionStart, breaking subsequent find()
When Matcher.matches() fails, SafeRE unconditionally resets searchFrom back to regionStart (typically 0). If the matcher previously had a successful match from find() or lookingAt() and was not explicitly reset via reset(), a subsequent call to find() restarts searching from 0 rather than resuming after the previous successful match.

This diverges from JDK java.util.regex.Matcher semantics and can cause find() to re-match previously matched text (e.g. duplicate matches or infinite loops).

From https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/regex/Matcher.html#find()

> This method starts at the beginning of this matcher's region, or, if a previous invocation of the method was successful and the matcher has not since been reset, at the first character not matched by the previous match.

```java

public class Repro {
    try {
      new Repro().one();
    } catch (CrosscheckException e) {
      System.err.println(e.getMessage());
    }
    try {
      new Repro().two();
    } catch (CrosscheckException e) {
      System.err.println(e.getMessage());
    }
  }

  void one() {
    Matcher m = Pattern.compile("a").matcher("a a");
    m.find();
    m.matches();
    // Throws CrosscheckException: SafeRE re-matches at 0, while JDK advances to 2
    m.find();
  }

  void two() {
    Matcher m = Pattern.compile("a*").matcher("a!");
    m.lookingAt();
    m.matches();
    // Throws CrosscheckException: SafeRE restarts at 0, while JDK starts at 1
    m.find();
  }
}
```

```
Crosscheck divergence in find → start()():
  SafeRE: 0
  JDK: 2

API call trace (5 calls):
  [1] find() → true
  [2] find() → start=0, end=1
  [3] matches() → false
  [4] find() → true
  [5] find.start() → DIVERGENCE
        SafeRE: 0
        JDK:    2

Crosscheck divergence in find → start()():
  SafeRE: 0
  JDK: 1

API call trace (4 calls):
  [1] lookingAt() → true
  [2] matches() → false
  [3] find() → true
  [4] find.start() → DIVERGENCE
        SafeRE: 0
        JDK:    1

```

eaftan:
Failed `matches()` and `lookingAt()` attempts reset the search cursor and cause a subsequent `find()` to repeat an earlier match. For example, `find(); matches(); find()` with `a` on `a a` returns the match at 0 twice instead of advancing to 2.

Retain the previous successful match end independently of current result validity and empty-match advancement. Keep search-cursor checks in `find()` rather than anchored matching helpers. Add systematic lifecycle tests, direct public API regressions, and a fuzz generator axis covering interleaved attempts, empty matches, captures, Unicode, anchors, regions, exhaustion, and resets. The change adds constant-size bookkeeping and no search retry loop.

Fixes #808.

Verification passed on JDK 26.0.2:

- `mvn -pl safere -Dtest=MatcherStateMachineTraceTest test -q` (new regression failed before the fix and passed afterward).
- `mvn -pl safere test -q` (full SafeRE suite).
- `mvn -pl safere-fuzz -am -Dtest=FindSequenceFuzzer -Dsurefire.failIfNoSpecifiedTests=false test -q` (regression mode).
- `mvn verify -pl safere,safere-crosscheck -am -Pcrosscheck-public-api-tests '-Dtest=*Test,!CrossEngineExhaustiveTest*' -q` (including the new direct regressions; the already-passed exhaustive suite was excluded from this repeat).
- `git diff --check`.
- Independent read-only review of the complete branch/worktree diff: no findings.

No live fuzzing campaign, benchmarks, or tests on other JDK versions were run.

### DIFF
--- a/safere-fuzz/src/test/java/org/safere/fuzz/FindSequenceFuzzer.java
+++ b/safere-fuzz/src/test/java/org/safere/fuzz/FindSequenceFuzzer.java
@@ -13,6 +13,31 @@ import org.junit.jupiter.api.Test;
 final class FindSequenceFuzzer {
 
   @Test
+  void anchoredFailureAfterSuccessfulMatchRegression() {
+    assertAnchoredContinuation("a", "a a", false, false);
+    assertAnchoredContinuation("a*", "a!", true, false);
+    assertAnchoredContinuation("a", "ba a", false, true);
+    assertAnchoredContinuation("a*", "!a", false, false);
+  }
+
+  private static void assertAnchoredContinuation(
+      String regex, String input, boolean initialLookingAt, boolean attemptLookingAt) {
+    FuzzSupport.MatcherPair matcher = FuzzSupport.compileOrSkip(regex, 0).matcher(input);
+    if (initialLookingAt) {
+      matcher.lookingAt();
+    } else {
+      matcher.find();
+    }
+    if (attemptLookingAt) {
+      matcher.lookingAt();
+    } else {
+      matcher.matches();
+    }
+    matcher.find();
+    matcher.find();
+  }
+
+  @Test
   void delimitedPrefixBeforeRequiredSuffixRegression() {
     FuzzSupport.CompiledPattern pattern =
         FuzzSupport.compileOrSkip("[^{']*(?:'[^']*'[^{']*)*\\{([^}]*)\\}", 0);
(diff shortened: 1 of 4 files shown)


eaftan:
PR Shepherd is now monitoring this PR at `968aaebcfffc894c31c8165cbeb888a2d4fe43e7`.

cushon:
This LGTM, for what it's worth. Thanks!

eaftan:
All required checks have passed. PR Shepherd is about to merge `968aaebcfffc894c31c8165cbeb888a2d4fe43e7`.

PR #816, opened 2026-09-08
Honor scoped UNIX_LINES flags in line assertions

### LINKED ISSUE #708: Honor scoped UNIX_LINES flags for end anchors
## Summary

SafeRE does not honor a scoped `(?-d:...)` override around an end anchor when the pattern was
compiled with `Pattern.UNIX_LINES`.

## Reproducer

```java
var safeRe = org.safere.Pattern.compile(
    "(?:foo|bar)(?-d:$)", org.safere.Pattern.UNIX_LINES);
var jdk = java.util.regex.Pattern.compile(
    "(?:foo|bar)(?-d:$)", java.util.regex.Pattern.UNIX_LINES);

System.out.println(safeRe.matcher("foo\r").find()); // false
System.out.println(jdk.matcher("foo\r").find());    // true
```

The alternation is intentional: it prevents the anchored-suffix reject prefilter from applying,
which confirms this is an engine-wide behavior rather than a prefilter-only bug.

The same behavior should be checked for String and UTF-8 matchers, `$` and `\Z`, locally enabling
and disabling `UNIX_LINES`, and all JDK line terminators.

## Specification

The JDK 26 `Pattern` documentation says embedded flags take effect at the point where they appear
and are restored after a scoped group. It also says that only newline is recognized while
`UNIX_LINES` is active; disabling it therefore restores the full line-terminator set.

- https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/regex/Pattern.html

## Likely design boundary

The parser retains scoped `ParseFlags.UNIX_LINES` on the anchor, but execution consults the
program-wide `Prog.unixLines()` value for empty-width and trailing-terminator checks. A principled
fix likely needs the effective line mode encoded in the relevant instruction or engine state rather
than inferred from the pattern-wide compile flag.

eaftan:
Scoped `UNIX_LINES` overrides were parsed correctly but matching engines and suffix prefilters used the pattern-wide flag. For example, `(?:foo|bar)(?-d:$)` compiled with `UNIX_LINES` failed to find `foo\r`; it now matches as the JDK does.

Encode each line assertion's scoped mode in the compiled program, retain the mode on stripped trailing anchors, and evaluate both modes across the engines and String/UTF-8 paths. Expand the OnePass assertion encoding with a size guard that falls back for oversized tables. Preserve CRLF context in DFA state and start-cache keys, and bound the start cache to 256 slots with full-key checks so the additional assertion bits do not multiply retained cache allocations.

Regression coverage includes locally enabling/disabling line modes, mixed assertions, all JDK line terminators, regions and bounds, String/UTF-8 matching, and mixed CRLF/standalone LF searches within and across matchers sharing a pattern. Add corresponding find-sequence fuzz coverage and an allocation regression test.

Includes the find-continuation fix and regression coverage from #820 to resolve the fuzz-regression CI failure tracked in #817. Restore the pre-advancement search position after a failed find without performing additional engine searches. The merge retains the CRLF cache regression tests from main; explicit DFA state tracks the CRLF context for scoped line modes.

Fixes #708. Refs #817.

Validation on merged head `0ad63bb`:

- Passed: `mvn verify -pl safere,safere-crosscheck -am -Pcrosscheck-public-api-tests -q` — full SafeRE suite, generated JDK public API crosscheck, and verification checks.
- Passed: `mvn -pl safere-fuzz -am -Dtest='*Fuzzer,*Fuzz*Test' -Dsurefire.failIfNoSpecifiedTests=false test -q` — all fuzz regression targets and fuzz support tests.
- The previously failing `fuzz-regression` CI job passes in run 34184996921. The remaining CI jobs were still running when these notes were updated.
- Not run locally: throughput benchmarks, sustained fuzzing, or the multi-JDK matrix.

### DIFF
--- a/.agents/skills/repo-assist/SKILL.md
+++ b/.agents/skills/repo-assist/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: repo-assist
-description: "Prepare one self-contained SafeRE maintainer report over trusted open PRs and issues: review stacked PRs in their whole-stack context, preserve the PR scout's fix-loop, benchmark, and ordering behavior, triage issue state and linked-PR coverage, and enforce a fail-closed content trust boundary before text reaches the model."
+description: "Prepare one self-contained SafeRE maintainer report over trusted open PRs: review stacked PRs in whole-stack context, preserve the PR scout's bounded fix-loop, benchmark, and ordering behavior, and enforce a fail-closed content trust boundary before text reaches the model."
 ---
 
 # Repo Assist
@@ -12,14 +12,14 @@ Prepare the data needed for a human SafeRE repository review while the reviewer 
 - which open non-draft PRs need attention;
 - whether each PR's idea makes sense and matches its implementation;
 - how each stacked PR contributes to the stack's shared objective and affects adjacent layers;
-- P2+ code-review findings fixed locally with `$review-fix-loop`;
+- P2+ code-review findings fixed locally with `$review-fix-loop` when the repair is bounded, or
+  reported to the author when correction requires a redesign;
 - benchmark reproduction for optimization PRs;
 - durable reports and artifacts that can be inspected later.
-- the current state, disposition, and linked-PR coverage of every trusted open issue.
 - one paste-ready, self-contained PR review containing everything the PR author needs to understand
   the findings, evidence, fixes, requests, and recommendation.
 
-Do not push branches, post PR comments, close issues, or publish review text unless the user
+Do not push branches, post PR comments, or publish review text unless the user
 explicitly asks.
 
 ## Required Inputs And Defaults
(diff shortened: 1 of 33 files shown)


eaftan:
PR Shepherd is now monitoring this PR at `a69356d6acf2bc9f37cd58867b41cc2c4cd56a4e`.

eaftan:
This PR is no longer tracked by PR Shepherd because its head moved to `0ad63bb66bf181fb6e64698451a07cf8bfab0739` without a verified PR Shepherd push. PR Shepherd cannot merge this head. To resume tracking, post a new comment containing only `/shepherd merge 0ad63bb66bf181fb6e64698451a07cf8bfab0739`.

eaftan:
PR Shepherd is now monitoring this PR at `2bc46327f103d4ce3e63ebe721d886f9e3214cf5`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `2bc46327f103d4ce3e63ebe721d886f9e3214cf5`.

PR #819, opened 2026-09-08
Document initial find behavior after failed full matches

### LINKED ISSUE #818: Find continuation diverges after failed full match with optional anchored capture
Found while adding lifecycle coverage for #817. This occurs before any successful or failed `find()` and is separate from the empty-match advancement defect in that issue.

```java
var m = org.safere.Pattern.compile("\\A(a?)").matcher("a!");
System.out.println(m.matches()); // false in both engines
System.out.println(m.find());    // SafeRE true; JDK 26 false
```

Replacing `org.safere.Pattern` with `java.util.regex.Pattern` produces the JDK result. The failing trace was observed on JDK 26.0.1. SafeRE retains the initial continuation position after the failed full match; JDK appears to retain a position reached during the failed attempt. Javadoc classification and whether this can be implemented with engine-native linear-time state remain to be assessed.

eaftan:
After a failed `matches()` on `\A(a?)` against `a!`, SafeRE's first `find()` matches `a`, while JDK 26.0.1 returns false. The documented initial search rule requires starting at the region beginning. Preserve SafeRE's behavior and document this as an intentional divergence, as approved in the issue assessment.

Add specification-based regression coverage for repeated failed full matches, greedy and reluctant quantifiers, empty matches, supplementary characters, and nonzero region starts. Exclude these intentional differences only from generated JDK crosscheck tests. No matcher implementation changes.

Fixes #818.

## Verification

Passed:
- `mvn -pl safere -Dtest=FailedFullMatchFindTest,MatcherStateMachineTraceTest test`
- `mvn -pl safere -Dtest=FailedFullMatchFindTest test -q` (review pass)
- `git diff --check`
- Independent review: no findings.

Not run: full SafeRE test suite, generated public API crosscheck, fuzz regression targets, or benchmarks. This change preserves existing runtime behavior.

### DIFF
--- a/INTENTIONAL_DIVERGENCES.md
+++ b/INTENTIONAL_DIVERGENCES.md
@@ -38,6 +38,38 @@ SafeRE preserves a coherent overall match instead of reproducing that
 implementation behavior. The JDK inconsistency is tracked upstream as
 [JDK-8390449](https://bugs.openjdk.org/browse/JDK-8390449).
 
+## Initial `find()` after a Failed Full Match
+
+Issue reference: #818.
+
+If a newly created or region-reset matcher fails `matches()` before any
+successful match, SafeRE preserves the region beginning as the initial
+`find()` search position. Failed full-match attempts do not consume input for
+that search. This follows the
+[JDK 26 `Matcher.find()` specification](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/regex/Matcher.html#find()),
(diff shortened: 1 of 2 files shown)


eaftan:
PR Shepherd is now monitoring this PR at `74b20f2e6fee6328415881552043622e54f66e22`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `74b20f2e6fee6328415881552043622e54f66e22`.

PR #804, opened 2026-09-07
Refactor Gap Engine: Extract GapScanner, Precompute Drivers, and Eliminate Heap Allocations

cushon:
Refactor `GapScanner` out of `MultiAnchorDescriptor`, leaving `MultiAnchorDescriptor` as immutable data.

Also

* Precompute Driver Indices: Computes `stringDriverIndex`, `utf8VectorDriverIndex`, and `utf8ScalarDriverIndex` during `Chain` initialization, turning `selectDriver` into an $O(1)$ field access.
* Eliminate Heap Allocations for Watermarks: Replaces allocated watermark collections in `MultiAnchorExecutor` with primitive local state.
* Prune Dead Code & Obsolete Fields: Removes unread record components (`Chain.driverIndex`, `Chain.isUpstreamBounded`), dead factory overloads, and obsolete prototype methods.

PR #827, opened 2026-09-09
Cache compiled multi-anchor eligibility

### LINKED ISSUE #825: Investigate per-search multi-anchor setup cost after reject-prefilter subsumption
## Problem

PR #807 removes a required-literal reject prefilter when a fixed-offset start plan already searches
for the same literal. This avoids a second full-input scan when the literal is present, but the
`isoTimestampLog.noMatch.1000` benchmark shows that the remaining execution path has materially
higher fixed setup cost.

The workload uses this pattern:

```regex
[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z error
```

Its no-match input repeats log lines in the form:

```text
2026/08/30 12:00:00 [worker-42] info: normal periodic heartbeat
```

The controlled standard comparison from the #807 review was:

| Workload | Main ns/op | PR + fixes ns/op | PR/main |
|---|---:|---:|---:|
| `isoTimestampLog.noMatch.1000` | 97 | 135 | 1.392 |
| `isoTimestampLog.noMatch.10000` | 816 | 859 | 1.053 |
| `isoTimestampLog.noMatch.100000` | 7,970 | 8,563 | 1.074 |

The same PR improves the matching variants by 1.31x to 1.48x, so restoring the duplicate scan
unconditionally would give back the intended benefit.

## Initial investigation

On baseline `e8ab7cd06fe0686ab97b145e4f5f9b8344ef4e29`, the compiled plans for this pattern are:

- `StartPlan.FixedOffset`: literal `Z error`, exact offset 19
- `RejectPlan.RequiredLiteral`: `Z error`
- an additional required-character-class reject plan

On corrected #807 commit `2af2bb6df2e0c93254940bba7c311017b87a1e38`, the fixed-offset start plan is unchanged and the
reject plan is `None`.

For this no-match input, main calls `String.indexOf("Z error")` from the reject prefilter and
returns immediately. The PR still performs one absent-literal scan, but it reaches that scan through
the multi-anchor path. `Matcher.doFindCore` calls `isExecutableChain()`, and
`MultiAnchorExecutor.find` calls it again before inspecting the chain, selecting its driver, and
calling `findNextCountingWork`.

An async-profiler CPU sample of the exact 1,000-byte workload showed the additional PR-side work in
`MultiAnchorDescriptor.isExecutableChain`, gap and segment accessors, record equality, and
`MultiAnchorExecutor`. The baseline returned from `RejectPrefilter` before that work.

A diagnostic no-fork comparison reproduced the difference:

| Revision | ns/op |
|---|---:|
| baseline `e8ab7cd0` | 95.585 ± 1.705 |
| corrected PR `2af2bb6d` | 125.179 ± 3.788 |
| controlled PR ablation restoring the early required-literal rejection | 92.855 ± 3.370 |

These diagnostic runs used JDK 26.0.2, 3 × 1-second warmups, and 12 × 1-second measurements with
`-f 0` for profiler attachment. The standard forked comparison above remains the authoritative
performance measurement; the diagnostic runs and profile support the path attribution.

## Investigation goal

Reduce or avoid the repeated per-search multi-anchor eligibility and setup work while preserving
#807's removal of the second literal scan on inputs where the fixed-offset literal is present.
Possible areas to measure include precomputing executable-chain metadata at pattern compilation and
avoiding repeated eligibility checks between `Matcher` and `MultiAnchorExecutor`.

Any change should retain the current correctness and linear-time invariants and be validated across
both matching and no-match fixed-offset workloads.

Related to #807.

eaftan:
Fixes #825.

After #807 removes the redundant required-literal reject prefilter, short absent-literal searches reach multi-anchor eligibility checks in both matcher dispatch and execution. Each check walks the compiled segments. Cache String and UTF-8 eligibility at descriptor construction so both callers can check it in constant time, preserving #807's scan removal.

Convert the internal descriptor from a record to a final class to hold the derived fields. Static computation helpers take the chain and prerequisite eligibility explicitly. Check empty gaps by kind while retaining the required `scanInfo != null` guard for bounded class gaps. Driver selection, candidate scans, verification, work limits, and fallbacks are unchanged. The predicates depend only on compiled metadata, which production consumers do not mutate. Move the existing array-component suppression to the specific nested record that needs it.

Add String/UTF-8 coverage for fixed chains across absent literals, failed candidates, repeated finds, and resets. A deterministic work-counter regression reports 600/1,200 segment visits before caching and zero afterward.

## Performance

Caching comparison baseline: merged #807, `54284571860bb1dbdd7cbcaf1bc22f43f54e2728`. Candidate: `b1a182c19c1710b57e260bc587a62a32445f32ca`, before the static-helper and empty-kind cleanup. OpenJDK 26.0.2+10-55, default SafeRE String engine, unchanged declared workloads, sequential runs through `./run-java-benchmarks.sh --fastbuild`. Work counters were disabled for benchmarks; the long baseline included only the disabled counter hooks and tests, without caching.

Standard mode (2 forks, 2 × 500 ms warmup, 5 × 500 ms measurement) reduced `isoTimestampLog.noMatch.1000` from **126.703 ± 7.181 to 98.964 ± 5.202 ns/op**, about 22% less time.

Long confirmation (2 forks, 3 × 1 s warmup, 5 × 1 s measurement):

| isoTimestampLog outcome | Size | Before ns/op | After ns/op | After / before |
|---|---:|---:|---:|---:|
| noMatch | 1000 | 123.170 ± 1.231 | 96.776 ± 1.682 | 0.786 |
| noMatch | 10000 | 825.924 ± 7.279 | 783.583 ± 23.522 | 0.949 |
| noMatch | 100000 | 7962.745 ± 331.673 | 7831.486 ± 147.420 | 0.984 |
| match | 1000 | 321.712 ± 3.390 | 296.106 ± 9.882 | 0.920 |
| match | 10000 | 1724.512 ± 12.398 | 1707.925 ± 30.829 | 0.990 |
| match | 100000 | 15881.538 ± 337.455 | 15515.188 ± 306.146 | 0.977 |

Errors are JMH 99.9% confidence interval half-widths. Long mode confirms about **21% less time on the target** and 8% less time on the matching 1000-byte case. Small differences with overlapping intervals are not established improvements; no regression was measured across these six cases. The baseline CPU profile confirms eligibility and gap-equality work on the multi-anchor path.

To reproduce, select all six `RealWorldRegexBenchmark.runBenchmark.isoTimestampLog.{noMatch,match}.{1000,10000,100000}@safere-string` trials as the comma-separated `crossEngineTrial` parameter of `org.safere.benchmark.CrossEngineBenchmark.run`, using the wrapper in standard and `--long` modes. Raw JSON, logs, the profile, exact runner script, and source patch are retained locally in `/tmp/safere-825-evidence/`; they are not checked in.

### Final implementation measurements

The final implementation was also measured in standard and long mode. Standard target time was **98.796 ± 2.443 ns/op**. A fresh long-mode comparison against `b1a182c19c1710b57e260bc587a62a32445f32ca` gives:

| isoTimestampLog outcome | Size | Existing PR head ns/op | Final implementation ns/op |
|---|---:|---:|---:|
| noMatch | 1000 | 99.907 ± 2.216 | 102.446 ± 4.196 |
| noMatch | 10000 | 824.417 ± 16.469 | 850.320 ± 50.869 |
| noMatch | 100000 | 8113.088 ± 192.089 | 8578.748 ± 330.403 |
| match | 1000 | 323.779 ± 26.278 | 306.164 ± 9.115 |
| match | 10000 | 1906.596 ± 89.998 | 1753.894 ± 27.337 |
| match | 100000 | 16670.963 ± 895.304 | 16165.072 ± 213.487 |

The target intervals overlap, as do those for the other two no-match cases. These runs do not establish an additional speedup from the cleanup; the target remains below the original uncached baseline. Raw outputs are `review-after*` and `review-before-fresh-long*` under the same local evidence directory. The fresh existing-head baseline was run after the final implementation; no benchmark runs overlapped.

## Verification

Passed on the original cached commit `b1a182c19c1710b57e260bc587a62a32445f32ca`:

- `mvn -pl safere test -q` (full SafeRE suite).
- Independent read-only review of the complete change: no findings.

Passed again on the final implementation:

- `mvn -pl safere test -q -Pwork-counters -Dtest=MultiAnchorGapEngineTest`.
- `mvn -pl safere test -q -Dtest=MultiAnchorGapEngineTest,MultiAnchorCompilerTest,PatternInternalTest,WorkCounterTest`.
- `git diff --check`.
- Standard and long benchmark comparisons for all six timestamp workloads.

The full suite and independent review were not repeated for the final cleanup. Not run: generated public API crosscheck, additional JDK versions, dedicated fuzzing, compilation benchmarks, broader benchmark workloads, or the external OpenJDK-derived suite.

### DIFF
--- a/safere/src/main/java/org/safere/MultiAnchorDescriptor.java
+++ b/safere/src/main/java/org/safere/MultiAnchorDescriptor.java
@@ -13,13 +13,15 @@ import java.util.Objects;
  * regular expression AST. Enables divide-and-conquer execution by pinning match positions around
  * fast SIMD anchors and verifying intermediate gaps.
  */
-@SuppressWarnings("ArrayRecordComponent")
-record MultiAnchorDescriptor(
-    Chain chain,
-    StartPlan startPlan,
-    RejectPlan rejectPlan,
-    String anchoredPrefix,
-    CharClassScanInfo anchoredCharClassPrefix) {
+final class MultiAnchorDescriptor {
+
+  private final Chain chain;
+  private final StartPlan startPlan;
+  private final RejectPlan rejectPlan;
+  private final String anchoredPrefix;
+  private final CharClassScanInfo anchoredCharClassPrefix;
+  private final boolean executableChain;
+  private final boolean executableUtf8Chain;
 
   public static final MultiAnchorDescriptor NONE =
       new MultiAnchorDescriptor(Chain.EMPTY, StartPlan.None.INSTANCE, RejectPlan.None.INSTANCE);
@@ -28,10 +30,42 @@ record MultiAnchorDescriptor(
     this(chain, startPlan, rejectPlan, null, null);
   }
 
-  public MultiAnchorDescriptor {
-    Objects.requireNonNull(chain, "chain");
-    Objects.requireNonNull(startPlan, "startPlan");
-    Objects.requireNonNull(rejectPlan, "rejectPlan");
+  MultiAnchorDescriptor(
+      Chain chain,
+      StartPlan startPlan,
+      RejectPlan rejectPlan,
+      String anchoredPrefix,
+      CharClassScanInfo anchoredCharClassPrefix) {
+    this.chain = Objects.requireNonNull(chain, "chain");
+    this.startPlan = Objects.requireNonNull(startPlan, "startPlan");
+    this.rejectPlan = Objects.requireNonNull(rejectPlan, "rejectPlan");
+    this.anchoredPrefix = anchoredPrefix;
+    this.anchoredCharClassPrefix = anchoredCharClassPrefix;
+    // Eligibility depends only on the compiled chain, whose segments are never mutated after
+    // construction. Cache both input domains so dispatch and execution can check it in constant
+    // time.
+    this.executableChain = computeExecutableChain();
+    this.executableUtf8Chain = computeExecutableUtf8Chain();
+  }
+
+  Chain chain() {
+    return chain;
+  }
+
+  StartPlan startPlan() {
+    return startPlan;
+  }
+
+  RejectPlan rejectPlan() {
+    return rejectPlan;
+  }
+
+  String anchoredPrefix() {
+    return anchoredPrefix;
+  }
+
+  CharClassScanInfo anchoredCharClassPrefix() {
+    return anchoredCharClassPrefix;
   }
 
   enum InputDomain {
@@ -427,6 +461,10 @@ record MultiAnchorDescriptor(
   }
 
   boolean isExecutableChain() {
+    return executableChain;
+  }
+
+  private boolean computeExecutableChain() {
     int n = chain.segments().length;
     if (n < 1 || chain.isEndAnchored() || !isExecutableLeadingGap(chain.segments()[0].gap())) {
       return false;
@@ -437,6 +475,9 @@ record MultiAnchorDescriptor(
       return false;
     }
     for (int i = 0; i < n; i++) {
+      if (WorkCounterConfig.ENABLED) {
+        WorkCounter.record();
+      }
       Segment segment = chain.segments()[i];
       if (!isExecutableAnchor(segment.anchor())) {
         return false;
@@ -487,10 +528,17 @@ record MultiAnchorDescriptor(
   }
 
   boolean isExecutableUtf8Chain() {
+    return executableUtf8Chain;
+  }
+
+  private boolean computeExecutableUtf8Chain() {
     if (!isExecutableChain()) {
       return false;
     }
     for (Segment segment : chain.segments()) {
+      if (WorkCounterConfig.ENABLED) {
+        WorkCounter.record();
+      }
       switch (segment.anchor()) {
         case Anchor.Single single -> {
           if (single.foldCase() && !isAscii(single.literal())) {
@@ -1719,6 +1767,8 @@ record MultiAnchorDescriptor(
       }
     }
 
+    // Ranges share compiled class metadata; array value equality is not used.
+    @SuppressWarnings("ArrayRecordComponent")
     record CharClass(AsciiBitmap bitmap, int[] ranges, CharClassScanInfo scanInfo)
         implements Anchor {
       static CharClass create(AsciiBitmap bitmap) {
--- a/safere/src/main/java/org/safere/MultiAnchorDescriptor.java
+++ b/safere/src/main/java/org/safere/MultiAnchorDescriptor.java
@@ -44,8 +44,8 @@ final class MultiAnchorDescriptor {
     // Eligibility depends only on the compiled chain, whose segments are never mutated after
     // construction. Cache both input domains so dispatch and execution can check it in constant
     // time.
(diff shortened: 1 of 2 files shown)


safere/src/main/java/org/safere/MultiAnchorDescriptor.java
cushon, line 645: While you're here consider:

```
    boolean isExecutorFixedGap() {
      return kind == GapKind.EMPTY
          || (kind == GapKind.BOUNDED_CLASS_REPEAT && minLength == maxLength);
    }
```

The full record `.equals` is comparing more than it needs to for `EMPTY`, this micro-optimizes `computeExecutableChain()` and `selectDriver()`

safere/src/main/java/org/safere/MultiAnchorDescriptor.java
```diff
@@ -28,10 +30,42 @@ record MultiAnchorDescriptor(
...
+    this.chain = Objects.requireNonNull(chain, "chain");
+    this.startPlan = Objects.requireNonNull(startPlan, "startPlan");
+    this.rejectPlan = Objects.requireNonNull(rejectPlan, "rejectPlan");
+    this.anchoredPrefix = anchoredPrefix;
+    this.anchoredCharClassPrefix = anchoredCharClassPrefix;
+    // Eligibility depends only on the compiled chain, whose segments are never mutated after
+    // construction. Cache both input domains so dispatch and execution can check it in constant
+    // time.
+    this.executableChain = computeExecutableChain();
+    this.executableUtf8Chain = computeExecutableUtf8Chain();
```
cushon: Maybe

```
    this.executableUtf8Chain = computeExecutableUtf8Chain(this.executableChain);
```

and making `computeExecutableUtf8Chain` static, to avoid calling instance methods in the constructor.

cushon (dismissed):
LGTM, thanks!

This looks like a good solution to #825.

There are a couple of small optimizations in #804 including driver index precomputation that should be complementary wtih this.

I left two small suggestions:
1. Replacing `Gap.equals(EMPTY)` with `kind == GapKind.EMPTY` to avoid `Record.equals` overhead.
2. Passing the precomputed `executableChain` boolean into `computeExecutableUtf8Chain(...)` to avoid invoking instance methods on `this` during constructor initialization.

safere/src/main/java/org/safere/MultiAnchorDescriptor.java
eaftan: Done

safere/src/main/java/org/safere/MultiAnchorDescriptor.java
eaftan: Also done

eaftan:
PR Shepherd is now monitoring this PR at `d60bbf829f1f1d6d95ac8e6354303bd71a2be21a`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `d60bbf829f1f1d6d95ac8e6354303bd71a2be21a`.

PR #830, opened 2026-09-10
Prune unanchored BitState start positions with earlier failed starts

wendigo:
Makes unanchored BitState searches linear by pruning start positions with the results of earlier, failed start positions.

## Problem

BitState only visit-tracks ALT instructions that sit on epsilon cycles. Non-cycle ALTs must stay revisitable *within* one start position: an outer repetition has to be able to re-enter a shared ALT entry point, otherwise nested quantifiers match prematurely. The side effect is that nothing is remembered *across* start positions either. For an unanchored search over a pattern with a leading greedy loop and a failing tail (`(a*)b`, `(.*),(.*)` on a line without a comma, `[a-z]+@` on a run of letters), every start position re-walks the entire remaining run. That is quadratic until the `MAX_WORK_PER_SLOT` budget trips, at which point the engine has spent the whole budget (≈1.1M steps for a 20k-char text) and then falls back to the Pike NFA anyway.

This is a production path, not just a micro-benchmark one: `Matcher.searchWithBitStateOrNfa` is called unanchored over the full text from the `preferCaptureEngine` branch once a caller consumes inner captures.

## Change

`BitState` keeps a second bitmap, `dead`, alongside the per-start `visited` bitmap:

- Every ALT reached during a start position's search is recorded in `visited` (non-cycle ALTs are recorded but not blocked, so within-start semantics are unchanged; cycle ALTs are check-and-set exactly as before).
- When a start position's search fails, its `visited` marks are folded into `dead` and cleared.
- Every ALT checks `dead` first. Whether a match is reachable from an `(ALT, position)` pair does not depend on capture values, so a pair that led nowhere from one start leads nowhere from any later start. This is the same property RE2's `bitstate.cc` relies on when it never clears its visited bitmap between starts.
- `PROGRESS_CHECK` loop registers do not break this: they only guard nullable loop bodies, so a first visit's body-only branch still reaches the exit through a zero-width iteration, and a path arriving with the register equal to the current position (exit-only) exists only because the same start already took the body from that position. The pre-change code already relied on this implicitly, because cycle-ALT `visited` bits persisted across starts with `PROGRESS_CHECK` downstream.
- Pruning is off for anchored searches (single start, no benefit, no bitmap cost).

Supporting details:

- The bitmap layout changes from instruction-major to position-major (`(pos - basePos) * progSize + instId`) so one start's marks are contiguous and the fold costs no more than the search that produced them.
- `reset()` now clears only the dirty word ranges of both bitmaps instead of the whole used prefix.
- `cycleAlts` is replaced by a per-instruction `altKind` byte so `shouldVisit` can tell "record" from "record and block" with one load.
- `DESIGN.md` describes the mechanism.

The dispatch loop itself is untouched. Two other RE2 techniques were tried first and rejected because they measured slower on the JVM despite doing less work: inlining single-successor transitions instead of push/pop (44% fewer stack operations, ~20% slower — the nested loop hurt branch prediction more than the L1-resident stack ops cost) and run-length-encoded job-stack entries (~35% slower on `(a+)+b`).

### Standard Java benchmark configuration (`./run-java-benchmarks.sh --fastbuild`, 2 forks, 2×500ms warmup, 5×500ms measurement)

Trials selected because they exercise BitState capture extraction; `unanchoredLogCaptures` is the flow this change targets.

| Trial (safere-string) | Before | After | Change |
|---|---|---|---|
| `RealWorld.unanchoredLogCaptures.match.1000` | 1,371.2 ± 19.9 ns/op | 1,387.3 ± 18.3 ns/op | +1.2% |
| `RealWorld.unanchoredLogCaptures.noMatch.1000` | 257.6 ± 3.0 ns/op | 268.0 ± 3.9 ns/op | +4.0% |
| `RealWorld.unanchoredLogCaptures.match.10000` | 8,695.9 ± 251.7 ns/op | 8,710.3 ± 391.5 ns/op | +0.2% |
| `RealWorld.unanchoredLogCaptures.noMatch.10000` | 2,320.8 ± 18.4 ns/op | 2,408.5 ± 98.3 ns/op | +3.8% |
| `RealWorld.multiInfixLogCaptures.match.1000` | 4,058.3 ± 170.5 ns/op | 4,090.8 ± 141.8 ns/op | +0.8% |
| `RealWorld.multiInfixLogCaptures.noMatch.1000` | 2,164.7 ± 33.9 ns/op | 2,205.8 ± 25.9 ns/op | +1.9% |
| `CaptureScalingBenchmark.capture1` | 55.0 ± 2.7 ns/op | 54.6 ± 3.2 ns/op | -0.8% |
| `CaptureScalingBenchmark.capture3` | 75.1 ± 1.7 ns/op | 74.2 ± 0.8 ns/op | -1.2% |
| `CaptureScalingBenchmark.capture10` | 198.3 ± 5.6 ns/op | 193.8 ± 1.9 ns/op | -2.3% |
| `RegexBenchmark.captureGroups` | 74.5 ± 1.2 ns/op | 75.7 ± 1.1 ns/op | +1.6% |
| `DiagnosticsDisabledBenchmark.bitStateShortMatches` | 12.5 ± 0.2 ns/op | 12.5 ± 0.1 ns/op | -0.5% |

These are the standard plan's capture-extraction trials. Their patterns start with a required literal (`error:\[`, digit runs) or are bounded by the DFA first, so a start position that fails does so within a few instructions and there is no long greedy run for pruning to remove; they serve as the no-regression check for the extra per-ALT bitmap write. All differences are within a few percent. The two `noMatch` trials sit just outside their error bars, so they were re-run with `--long`:

| Trial (safere-string), `--long` | Before | After | Change |
|---|---|---|---|
| `RealWorld.unanchoredLogCaptures.noMatch.1000` | 267.2 ± 7.9 ns/op | 269.8 ± 7.7 ns/op | +1.0% |
| `RealWorld.unanchoredLogCaptures.noMatch.10000` | 2,396.5 ± 88.1 ns/op | 2,425.0 ± 87.3 ns/op | +1.2% |

Both are inside their error bars: no measurable regression.

The measured improvement is on unanchored BitState searches whose leading greedy loop fails, which the standard plan does not contain; those are shown with the direct harness below.

### Direct `BitState.search` micro-harness (ad hoc, not part of the repo; best-of-7 × reps, same JVM flags for both sides)

| Case | Before | After | Change |
|---|---|---|---|
| `a*b`, unanchored, 20 000 chars, no match | 2.97 ms/op | 0.29 ms/op | ~10× faster; no longer exhausts the work budget |
| `[a-z]+@`, unanchored, 20 000 chars, no match | 3.11 ms/op | 0.34 ms/op | ~9× faster |
| `Matcher.find()` loop, `(.*?)(\d+),`, 1 000 matches, BitState forced via `EnginePathOptions` | 0.21 ms/op | 0.16 ms/op | −22% |
| `(a+)+b`, unanchored, 3 000 chars | 0.66 ms/op | 0.68 ms/op | unchanged (blowup is within one start and still budget-capped) |
| 150 sequential groups, anchored | 0.0044 ms/op | 0.0044 ms/op | unchanged (pruning off when anchored) |

## Rejected follow-up: lazy-allocating `dead`

`dead` is the same size as `visited` but only used by unanchored searches, so it doubles the memory footprint of a `BitState` instance that only ever runs anchored (e.g. one backing only `matches()`/`lookingAt()`). Allocating it lazily, on the first unanchored `doSearch`, seemed like a free win — total allocation count for an unanchored-heavy workload is unchanged; only the anchored-only case saves anything.

Measured instead: 40-60% *slower* on unanchored searches, i.e. exactly the case that does prune and whose allocation count didn't change (`(a+)+b` micro-harness: 0.68ms → 1.09ms; isolated re-run: 0.67ms → 0.88ms). Moving the always-happens-anyway allocation from the constructor into `doSearch` was enough to cause it, which points at allocation-site/escape-analysis effects rather than allocation count. Reverted; `BitState.java` has a code comment recording this so it isn't retried blind. Left as a real but unclaimed opportunity if someone wants to chase the actual mechanism.

## Verification

Ran:

- `mvn -pl safere test -Dtest=BitStateTest,MatcherTest,RE2RegressionTest,CrossEngineExhaustiveTest,MatcherDeferredCaptureStateTest,ScopedLineModeAnchorTest,EnginePathEquivalenceTest,DiagnosticsTest,Utf8EnginePathEquivalenceTest,Utf8DiagnosticsTest` — pass (includes the exhaustive BitState-vs-NFA cross-engine suite).
- `mvn -pl safere test -Pwork-counters -Dtest=MatcherLinearTimeTest,SearchScalingRegressionTest,Utf8LinearTimeTest,GraphemeLinearTimeTest,WorkCounterTest` — pass.
- `mvn -pl safere spotless:check` — clean.
- Tests in `BitStateTest.Budget`: a 20k-char greedy run no longer exceeds the work budget; a leftmost match that begins after several pruned starts is still found with correct captures; BitState/NFA agreement on seven nullable-body (`PROGRESS_CHECK`) loop patterns × eight inputs, plain and ×40 repeated; a cached instance reused across anchored → unanchored → anchored searches still returns correct captures each time (added while investigating the lazy-allocation idea above; independent of it).
- Standard JMH benchmarks above, including the `--long` confirmation.

Not run:

- The full `safere` module test suite and the `safere-crosscheck` public-API crosscheck reactor.
- Fuzz targets (`safere-fuzz`).

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -298,10 +298,21 @@ at most once.
 
 - **When used:** Text length ≤ `MAX_BITMAP_BITS / prog.size()` (roughly
   256 KB / program size).
-- **Epsilon-cycle optimization:** Only ALT instructions involved in
-  epsilon cycles are visit-tracked.  Non-cycle ALTs can be revisited,
-  which is essential for nested quantifiers where an outer repetition must
-  re-enter a shared ALT entry point.
+- **Epsilon-cycle optimization:** Within one start position, only ALT
+  instructions involved in epsilon cycles are blocked on revisit.
+  Non-cycle ALTs can be revisited, which is essential for nested
+  quantifiers where an outer repetition must re-enter a shared ALT entry
+  point.
+- **Cross-start pruning:** Every ALT reached by an unanchored start
+  position whose search fails is folded into a second bitmap, and later
+  start positions skip those `(ALT, position)` pairs.  Whether a match is
+  reachable from such a pair does not depend on capture values, so this is
+  exact.  It keeps unanchored search linear instead of re-walking the same
+  greedy run from every start until the work budget trips and forces an
+  NFA fallback.  PROGRESS_CHECK loop registers only guard nullable bodies,
+  so a failed start still explores every future a later start could reach
+  through a pruned pair.  Off for anchored searches, which have a single
+  start.
 
 ### NFA / Pike VM (`Nfa`)
 
--- a/safere/src/main/java/org/safere/BitState.java
+++ b/safere/src/main/java/org/safere/BitState.java
@@ -270,12 +270,16 @@ final class BitState {
     matchResult =
         resultBuffer != null && resultBuffer.length >= ncap ? resultBuffer : new int[ncap];
     int limit = anchored ? startPos + 1 : Math.min(searchLimit + 1, textLen + 1);
+    pruneAcrossStarts = !anchored;
     for (int searchStart = startPos; searchStart < limit; searchStart++) {
       if (trySearch(prog.start(), searchStart)) {
         return bestMatch;
       }
       if (budgetExceeded) {
         return null;
+      }
+      if (pruneAcrossStarts) {
+        foldVisitedIntoDead();
       }
       if (searchStart < textLen) {
         searchStart = InputScanner.position(text.decodeForward(searchStart)) - 1;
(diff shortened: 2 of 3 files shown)


eaftan (approved):
Thanks for sending this and #832!

The cross-start pruning design looks sound. I looked closely at whether `PROGRESS_CHECK` registers make `(ALT, position)` an incomplete failure key, but targeted comparisons covering nested nullable loops, greedy and reluctant repetition, alternation, bounded repetitions, nonzero starts, and bounded search limits found no case where pruning hid a match or changed captures. The dirty-range reset and anchored-search handling also look correct.

I added benchmarks that match what you described and reproduced your reported speedups:

| No-match workload | Size | Before | After | Speedup |
|---|---:|---:|---:|---:|
| `a*b` | 1K | 160.7 µs | 24.9 µs | 6.46x |
| `a*b` | 5K | 1,308.5 µs | 135.8 µs | 9.64x |
| `a*b` | 10K | 2,604.6 µs | 264.7 µs | 9.84x |
| `a*b` | 20K | 5,233.5 µs | 546.4 µs | 9.58x |
| `[a-z]+@` | 1K | 164.1 µs | 31.3 µs | 5.25x |
| `[a-z]+@` | 5K | 1,374.6 µs | 157.0 µs | 8.76x |
| `[a-z]+@` | 10K | 2,761.7 µs | 317.8 µs | 8.69x |
| `[a-z]+@` | 20K | 5,601.0 µs | 643.9 µs | 8.70x |

The additional bitmap costs at most approximately 32 KiB per live or cached `BitState`. Given the reproduced speedup, elimination of work-budget fallback, and strong correctness evidence, that memory tradeoff looks justified.

With the benchmarks added to the PR, this LGTM.

PR #832, opened 2026-09-10
Precompute case-insensitive search state for multi-anchor Alternation

wendigo:
Precomputes per-literal case-insensitive search state for the multi-anchor engine's `Alternation` anchor, instead of rebuilding it on every `findNext()` probe.

## Problem

`MultiAnchorDescriptor.Anchor.Alternation.findNext(String, int)` (the multi-anchor engine's literal-set prefilter for `foo|bar|baz`-shaped alternations) calls `Matcher.indexOfIgnoreCase(text, lit, fromIndex)` per literal on every probe. That 3-arg convenience overload recomputes, from scratch, on every call:

- the "rarest ASCII anchor character" for the literal (`RarityOracle.rarestAsciiOffset`, an O(literal length) scan), and
- a `ClassHashChain` (a ~1KB shift table plus a nested Unicode case-fold expansion; `ClassHashChain.compileCaseInsensitive`), used both to accelerate false-anchor-hit verification on the ASCII path and to drive the search entirely on the non-ASCII (Unicode-fold) path.

Every other call site that needs this state (`LiteralPreparedRunner`, the descriptor's own single-literal path at `Matcher.java:3520-3544`) precomputes it once and threads it through the 7-arg overload. `Alternation` was the one path that didn't — it was presumably missed when those other call sites were built out, since `Alternation.create()` already precomputes the analogous non-fold-case state (`MultiLiteralInfo`, `TeddyModel`).

This matters most exactly where the multi-anchor engine calls `findNext` a lot: `MultiAnchorExecutor`'s dense per-gap probing, where each individual call only scans a short remaining range, so the previously-per-call rebuild cost is a large fraction of total time instead of being amortized over a long scan.

## Change

`Alternation` gains four parallel arrays (`anchorOffsets`, `anchorLows`, `anchorHighs`, `classHashChains`), populated once in `create()` using the same per-literal logic `LiteralPreparedRunner`'s constructor already uses, only when `foldCase` is true. `findNext(String, int)` now calls the 7-arg `Matcher.indexOfIgnoreCase` overload with the precomputed values instead of the 3-arg one. `findNext(Utf8InputScanner, int)` is unaffected — it never called `indexOfIgnoreCase` in the first place.

## Benchmark

Ad-hoc micro-harness (not part of the repo), best-of-7 over 400k reps, JIT-warmed, calling `Alternation.findNext` over a pool of 8192 distinct 40-char random-text gaps (to prevent the JIT from proving the loop body redundant across iterations — an earlier same-input-every-call version of this benchmark showed no difference for exactly that reason) with 4 literals of length 7-8, `foldCase=true`, no match present:

| | Before | After | Change |
|---|---|---|---|
| `Alternation.findNext` (foldCase, short-gap, no match) | ~1250 ns/op | ~140 ns/op | ~9x faster |

Reproduced across 3 alternating before/after runs on an idle machine (1189-1272 ns/op baseline vs. 123-153 ns/op fixed each time), not noise.

An earlier version of this same benchmark using one long (50k-char) text with a single `findNext(text, 0)` call per rep showed no measurable difference — the O(n) scan cost dominates the O(1) setup cost at that scale, and for all-ASCII literals the anchor-char occurrences during that scan also dominate. The win specifically shows up in the dense/many-short-probes shape, which is what the real caller (`MultiAnchorExecutor`) does.

## Verification

Ran:
- `mvn -pl safere test -Dtest=AlternationFindNextTest,MultiAnchorGapEngineTest,MultiAnchorCompilerTest,MatcherTest,RE2RegressionTest,CrossEngineExhaustiveTest,PatternSetTest` — pass.
- `mvn -pl safere spotless:check` — clean.
- New `AlternationFindNextTest`: 2-literal and N(>2)-literal alternations; ASCII and non-ASCII (mixed in the same `Alternation`, to exercise both the ASCII-with-ClassHashChain and pure-Unicode-fold branches of `indexOfIgnoreCase`) literals; single-character literals (bypass `ClassHashChain` entirely); `fromIndex` advancing past an earlier match to find a later one of a different literal (catches per-literal-index bookkeeping bugs); non-folded alternation unaffected.

Not run:
- The full `safere` module test suite and the `safere-crosscheck` public-API crosscheck reactor.
- Fuzz targets (`safere-fuzz`).
- Standard JMH benchmark suite (`run-java-benchmarks.sh`) — the change is internal to an acceleration heuristic with no dedicated JMH trial; the ad-hoc harness above targets the mechanism directly.

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -3174,6 +3174,16 @@
         "length": "{size}"
       },
       "shared": true
+    },
+    {
+      "id": "alternationFindNext.shortGapPool",
+      "recipe": {
+        "kind": "randomChars",
+        "alphabet": "abcdefghijklmnopqrstuvwxyz",
+        "length": 327680,
+        "seed": 20260910
+      },
+      "shared": false
     }
   ],
   "workloads": [
@@ -13059,6 +13069,34 @@
       "expected": {
         "type": "boolean",
         "value": true
+      }
+    },
+    {
+      "id": "AlternationFindNextBenchmark.foldedShortGapNoMatch",
+      "operation": "alternationFindNext",
+      "patterns": [
+        "NETWORK",
+        "STORAGE",
+        "DATABASE",
+        "SECURITY"
+      ],
+      "inputs": [
+        "alternationFindNext.shortGapPool"
+      ],
+      "arguments": {
+        "inputLength": 40
+      },
+      "resultConsumption": "integer",
+      "expected": {
+        "type": "integer",
+        "value": -1
+      },
+      "measurement": {
+        "mode": "averageTime",
+        "timingUnit": "nanoseconds",
+        "constraints": [
+          "prematerializedInput"
+        ]
       }
     }
   ]
--- a/safere-benchmarks/src/main/java/org/safere/AlternationFindNextBenchmarkRunner.java
+++ b/safere-benchmarks/src/main/java/org/safere/AlternationFindNextBenchmarkRunner.java
@@ -1,0 +1,42 @@
+// This file is part of a Java port of RE2 (https://github.com/google/re2).
+// Original RE2 code is Copyright (c) 2009 The RE2 Authors.
+// Modifications and Java port Copyright (c) 2026 Eddie Aftandilian.
+// Licensed under the BSD 3-Clause License (see LICENSE file).
+
+package org.safere;
+
+import java.util.List;
+
+/** Benchmark-only adapter for measuring package-private alternation anchor searches directly. */
+public final class AlternationFindNextBenchmarkRunner {
+  private final MultiAnchorDescriptor.Anchor.Alternation alternation;
+  private final String[] inputs;
+  private int nextInput;
+
+  /** Creates a runner over fixed-width slices of a materialized input pool. */
+  public AlternationFindNextBenchmarkRunner(
+      List<String> literals, String inputPool, int inputLength) {
+    if (inputLength <= 0 || inputPool.length() % inputLength != 0) {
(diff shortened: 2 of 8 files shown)


cushon:
This LGTM. I think there are additional opportunities to avoid recompilation of ClassHashChain, I opened #837 for some related changes

eaftan (approved):
The implementation looks correct. The per-literal arrays stay aligned, and the precomputed path preserves the existing case-insensitive search behavior for empty, single-character, ASCII, and Unicode literals. This also complements #837 rather than duplicating it: this PR covers alternation anchors, while #837 covers single anchors and repeated literal splitting.

I've pushed a commit adding a declarative JMH benchmark for the workload described in the PR: 8,192 distinct rotating 40-character inputs, four case-insensitive literals of length 7-8, and no matches. Lower is better, and `PR/base` is elapsed PR time divided by elapsed baseline time.

| Workload | Base | PR | PR/base | Interpretation |
|---|---:|---:|---:|---|
| Folded alternation, rotating short-gap no-match pool | 2,398.799 ± 72.000 ns/op | 282.764 ± 5.135 ns/op | 0.118x | PR is 8.48x faster |

That closely reproduces the reported approximately 9x improvement. Each eligible folded literal can retain a roughly 1 KiB `ClassHashChain` for the pattern's lifetime, but this result justifies that tradeoff for the repeated short-probe path. LGTM.

eaftan:
PR Shepherd is now monitoring this PR at `d3e16e3c798c31fe920ec2c042c425d900edb3f1`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `d3e16e3c798c31fe920ec2c042c425d900edb3f1`.

PR #834, opened 2026-09-10
Decouple Multi-Anchor Gap Engine from preferCaptureEngine in Matcher

cushon:
In `Matcher.java`, `shouldPreferCaptureEngine(prog, scanner)` evaluates to `true` when all of the following conditions are met:
1. `parentPattern.innerCapturesObserved()` is `true` (callers have accessed `matcher.group(...)` on the compiled `Pattern`).
2. `scanner.length() <= CAPTURE_DEMAND_TEXT_LIMIT` (the input is $\le 512$ bytes).
3. `options.bitState()` is enabled.

Historically, this check was designed to avoid running the full DFA for group 0 on short inputs only to immediately replay the entire search range through `BitState` once capture consumption was proven.

However, line 1709 previously gated the Multi-Anchor Gap Engine behind `!preferCaptureEngine`:

As a result, as soon as a caller read any capturing group on a pattern, all subsequent searches on inputs <512 bytes bypassed the gap engine, falling back into unanchored `searchWithBitStateOrNfa(...)`.

---

* **Decouple the Gap Engine**: Remove `!preferCaptureEngine` from the Multi-Anchor Gap Engine dispatch check in `Matcher.java`
* **Defer Capture Evaluation**: Move the `preferCaptureEngine` calculation down to line 1800 so it only executes when falling through to `BitState` / `Nfa`.
* **Preserve Deferred Capture Semantics**: The Multi-Anchor Gap Engine rapidly proves and bounds match limits $[S, E]$ (~150 ns) with deferred capture extraction. When `matcher.group(i)` is subsequently accessed, `resolveCaptures()` evaluates captures strictly over the tiny 15-byte matched slice $[S, E]$ (~90 ns) rather than rescanning the unanchored input from index 0.

---

```
./safere-benchmarks/scripts/compare-branch.sh --baseline decouple-gap-engine-capture-demand-baseline --current decouple-gap-engine-capture-demand 'RealWorldRegexBenchmark\.runBenchmark\.fixedAnchorLogCaptures.*@safere-string'
```

| Benchmark                           | baseline (ns/op) | current (ns/op) | Speedup |
| ----------------------------------- | ---------------- | --------------- | ------- |
| fixedAnchorLogCaptures.match.256    |        6023 ± 43 |       249 ± 8.9 |   24.2x |
| fixedAnchorLogCaptures.match.1000   |        489 ± 5.6 |        490 ± 19 |   1.00x |
[...]

PR #835, opened 2026-09-10
Direct Gap Capture Extraction for Multi-Anchor Engine

cushon:
Builds upon #833 by enabling direct capture group extraction within `MultiAnchorExecutor` when all user-defined capture groups in a pattern map to boundary points of the multi-anchor chain. This completely eliminates secondary capture engine replay (`BitState` / `OnePass`) on match confirmation and group access for these patterns.

In #833, the multi-anchor gap engine was decoupled from `preferCaptureEngine`, allowing patterns with capture groups (e.g., `^.*\[([A-Z]+)\]: ([0-9]+).*$`) to match via `MultiAnchorExecutor` and resolve captures lazily on demand.

However, resolving captures on demand previously required falling back to a secondary capture engine (`BitState` or `OnePass`), which re-scanned the matched substring with an NFA. For unanchored or fixed-anchor multi-anchor chains where the capture groups enclose whole anchors or gaps, the anchor and gap boundaries are already definitively located during the multi-anchor scan. Re-scanning with `BitState` is redundant work.

---

1. **Boundary Point Coordinate System**:
   A multi-anchor chain with $K$ segments defines $2K + 2$ deterministic boundary points:
   - $P_0 = \text{matchStart}$ (start of leading gap $G_0$)
   - $P_1 = p_0$ (start of Anchor 0)
   - $P_2 = p_0 + \text{len}_0$ (end of Anchor 0 / start of Gap 1)
   - $P_{2k+1} = p_k$ (start of Anchor $k$)
   - $P_{2k+2} = p_k + \text{len}_k$ (end of Anchor $k$)
   - $P_{2K} = p_{K-1} + \text{len}_{K-1}$ (start of trailing gap $G_K$)
   - $P_{2K+1} = \text{matchEnd}$ (end of trailing gap $G_K$)

2. **Compile-Time Capture Span Mapping (`MultiAnchorCompiler`)**:
   - Analyzes AST capture groups.
   - Any outer capture spanning the chain maps to $[0, 2K + 1]$.
   - Any captured anchor $k$ maps to $[2k + 1, 2k + 2]$.
   - Any captured gap $k$ maps to $[2k, 2k + 1]$.
   - If **all** user capture groups map cleanly to boundary point pairs, `MultiAnchorDescriptor` records the mapping in `captureGroupSpans`.
   - If any capture cannot be mapped (e.g., inner captures inside an anchor or alternation), `captureGroupSpans` is `null` and execution safely falls back to standard deferred capture resolution via `BitState`/`OnePass`.

3. **Execution & Direct Extraction (`MultiAnchorExecutor`)**:
   - When `groups != null && descriptor.canExtractAllCaptures()`, anchor coordinates are recorded into an `int[numSegments * 2]` array during the upstream/downstream anchor scan.
   - For non-capturing searches (`groups == null` or boolean matches), **zero array allocations** occur.
   - Upon match confirmation, `groups` are populated directly from the boundary points via `extractCaptures()`.

4. **Matcher Integration (`Matcher`)**:
   - When `canExtractAllCaptures()` is true, `applyDirectCaptureMatchResult(start, end)` marks `capturesResolved = true` immediately, avoiding any deferred resolution overhead when `group(i)` is called.

---

```
  --baseline decouple-gap-engine-capture-demand \
  'RealWorldRegexBenchmark\.runBenchmark\.fixedAnchorLogCaptures.*@safere-string'
```

| Benchmark                           | baseline (ns/op) | current (ns/op) | Speedup |
| ----------------------------------- | ---------------- | --------------- | ------- |
| fixedAnchorLogCaptures.match.256    |         249 ± 10 |       184 ± 2.7 |   1.35x |
| fixedAnchorLogCaptures.match.1000   |         507 ± 18 |        429 ± 12 |   1.18x |
| fixedAnchorLogCaptures.noMatch.256  |         63 ± 1.8 |        63 ± 1.3 |   1.00x |
| fixedAnchorLogCaptures.noMatch.1000 |        185 ± 2.7 |       184 ± 1.7 |   1.00x |


safere/src/main/java/org/safere/MultiAnchorExecutor.java
```diff
@@ -119,6 +124,8 @@ static Result find(MultiAnchorDescriptor descriptor, Utf8InputScanner scanner, i
     }
 
     boolean hasLeadingAnyStar = leadingGap.kind() == MultiAnchorDescriptor.GapKind.ANY_STAR;
+    boolean trackCaptures = groups != null && descriptor.canExtractAllCaptures();
+    int[] anchorPositions = trackCaptures ? new int[numSegments * 2] : null;
```
eaftan: This introduces a fresh `int[numSegments * 2]` allocation for every search of an extractable capturing pattern, including unsuccessful searches. The String overload has the same allocation below. Since the gap engine has otherwise aimed for zero per-search allocation, could this scratch space be retained and reused by Matcher? Alternatively, could we add an allocation benchmark showing that the resulting allocation rate and GC overhead are acceptable?

safere/src/test/java/org/safere/MultiAnchorGapEngineTest.java
```diff
@@ -952,6 +952,70 @@ void unixLinesDotMatchesCarriageReturnAndStopsAtNewline() {
...
+        "(AAA)[0-9]{2}(BBB)([a-z]{2})",
+      })
+  void directGapCaptureExtractionMatchesJdk(String regex) {
+    Pattern pattern = Pattern.compile(regex);
+    java.util.regex.Pattern jdkPattern = java.util.regex.Pattern.compile(regex);
+
+    assertThat(pattern.multiAnchor().isExecutableChain()).isTrue();
+    assertThat(pattern.multiAnchor().canExtractAllCaptures()).isTrue();
+
+    String[] testInputs = {
```
eaftan, line 1096: Could we include inputs containing multibyte UTF-8 characters, both before and inside captured regions? This test exercises `Utf8Matcher`, but every current input is ASCII, so it does not directly verify that the newly extracted capture coordinates remain byte offsets when character and byte positions differ.

eaftan (changes requested):
This boundary-point design looks sound and appropriately fails closed when a capture cannot be represented exactly. I reproduced the reported benchmark results:

| Workload | #834 ns/op | This PR ns/op | Speedup |
|---|---:|---:|---:|
| Match, 256 bytes | 179 ± 7.6 | 133 ± 2.8 | 1.35x |
| Match, 1,000 bytes | 357 ± 4.8 | 312 ± 3.6 | 1.15x |
| No match, 256 bytes | 46 ± 0.46 | 47 ± 0.55 | 1.00x |
| No match, 1,000 bytes | 146 ± 38 | 137 ± 4.8 | 1.06x |

My main concern is the new per-search anchor-coordinate allocation noted inline. I’d like either reusable scratch storage or allocation evidence before merging this optimization. I’d also like committed multibyte UTF-8 coverage for the new coordinate-extraction path.  See my line comments.

safere/src/main/java/org/safere/MultiAnchorExecutor.java
cushon: Done, I added a reuseable scratch array in `Matcher` that is passed in.

safere/src/test/java/org/safere/MultiAnchorGapEngineTest.java
cushon: Thanks, I added more UTF-8 test coverage.

PR #837, opened 2026-09-10
Precompute and reuse case-insensitive literal search state in Anchor.Single and Matcher.findNextMatchPacked

cushon:
This change avoids some runtime recompilation of `ClassHashChain`. This was inspired by #832.

When searching for case-insensitive literals of length $\ge 4$, `Matcher.indexOfIgnoreCase(...)` compiles a 2-gram `ClassHashChain` (allocating code-point fold variant arrays, stream pipelines, and a 1 KB shift table). Two literal search paths called un-cached overloads of `indexOfIgnoreCase`, triggering repeated dynamic recompilation during matching:
1. `Anchor.Single.findNext`: called the 6-argument overload on every candidate probe during multi-anchor scanning.
2. `Matcher.findNextMatchPacked`: called the 3-argument overload on every match iteration during `Pattern.split` and `Pattern.splitAsStream`.

---

1. **`MultiAnchorDescriptor.Anchor.Single`**:
   - Added `ClassHashChain classHashChain` component to `record Single`.
   - Precompute `classHashChain` in `Single.create(String literal, boolean foldCase)` when `foldCase && literal.length() >= 4`.
   - Forward `classHashChain` to the 7-argument `Matcher.indexOfIgnoreCase` overload in `Single.findNext`.

2. **`Matcher.findNextMatchPacked`**:
   - Replaced un-cached 3-arg `indexOfIgnoreCase` call.
   - Extract precomputed `anchorOffset`, `anchorLow`, `anchorHigh`, and `classHashChain` from `parentPattern.preparedMatchRunner(false)` when it is a `LiteralPreparedRunner`.
   - When not a `LiteralPreparedRunner` (e.g. when capturing groups exist on the literal pattern), fall back to `parentPattern.matchDescriptor().classHashChain()` and compute anchor offsets with empty-literal guards, matching `Matcher.replaceAllCore`.

### DIFF
--- a/safere-benchmarks/benchmark-data.json
+++ b/safere-benchmarks/benchmark-data.json
@@ -1213,6 +1213,22 @@
       "shared": true
     },
     {
+      "id": "realWorldRegex.caseInsensitiveSplit.{size}",
+      "axes": {
+        "size": [
+          1000,
+          10000,
+          100000
+        ]
+      },
+      "recipe": {
+        "kind": "repeatToLength",
+        "unit": "event payload record item with status pending and metadata\n----delimiter----\n",
+        "length": "{size}"
+      },
+      "shared": true
+    },
+    {
       "id": "realWorldRegex.fixedAnchorLog.match.{size}",
       "axes": {
         "size": [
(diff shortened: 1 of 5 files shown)


wendigo:
Any benchmarks to share?

cushon:
> Any benchmarks to share?

Good call, I added a benchmark.

I expected this to mostly show up in microbenchmarks and to save allocations, I don't think the benefit here is very general, but I added an e2e benchmark showing an example with `split` where this makes a significant difference.

```
./safere-benchmarks/scripts/compare-branch.sh   --baseline baseline-case-insensitive-benchmark   'caseInsensitiveLiteralSplit.*@safere-string'
```

| Benchmark                          | baseline (ns/op) | current (ns/op) | Speedup |
| ---------------------------------- | ---------------- | --------------- | ------- |
| caseInsensitiveLiteralSplit.1000   |     18966 ± 2710 |        919 ± 70 |   20.6x |
| caseInsensitiveLiteralSplit.10000  |   177341 ± 26092 |      9047 ± 195 |   19.6x |
| caseInsensitiveLiteralSplit.100000 | 1757498 ± 243664 |    84256 ± 1904 |   20.9x |

wendigo:
Thanks! Nice improvement

eaftan (approved):
This looks good. The cached `ClassHashChain` objects are immutable and scoped to the compiled pattern or anchor, and the split fallback with capture groups reuses the table for the same literal. Short and non-ASCII literals continue through the existing fallback paths. This complements #832: that PR caches alternation-anchor state, while this one covers single anchors and repeated literal splitting.

I reproduced the checked-in split benchmark with a controlled main-based baseline containing the identical workload declarations. `Speedup` is baseline time divided by PR time.

| Workload | Baseline ns/op | PR ns/op | Speedup |
|---|---:|---:|---:|
| Case-insensitive literal split, 1,000 bytes | 12,039 ± 265 | 628 ± 9.9 | 19.2x |
| Case-insensitive literal split, 10,000 bytes | 113,893 ± 2,863 | 6,993 ± 138 | 16.3x |
| Case-insensitive literal split, 100,000 bytes | 1,132,201 ± 42,077 | 63,066 ± 752 | 18.0x |

The result is close to the reported approximately 20x improvement. 🎉 LGTM

PR #852, opened 2026-09-13
Fix multi-anchor variable-gap matching invariants

eaftan:
Part of #805. This draft targets `variable-gap-chains` directly; it does not change that branch. It offers a tested correctness repair for Liam to review, adapt, or use while deciding how to proceed with the optimization. **This is not a performance endorsement or a request to merge the draft as-is.**

The submitted multi-anchor path has several matching problems when variable gaps interact:

- After an upstream gap backtracks, a descendant can reuse a guarded endpoint or reluctant-search watermark established for a different gap start. That can respectively match across a forbidden delimiter or skip a valid match.
- Bounded `.` gaps compare UTF-16 units or UTF-8 bytes with quantifier bounds, rather than counting Unicode code points.
- Variable *leading* gaps can change the JDK-compatible leftmost match chosen by the accelerated path.
- Repeated fixed, reverse, leading, and trailing scans need to be charged to the work budget when backtracking retries them.

This patch invalidates descendant scan state when its start changes, counts bounded wildcard gaps in code points, charges the repeated scans, and falls back to the general engine on budget exhaustion. For now, it also routes variable-leading-gap patterns through the general engine. Interior and trailing variable-gap eligibility remains. The new tests cover both String and UTF-8, greedy and reluctant gaps, supplementary characters, leftmost priority, and budget exhaustion.

### Performance tradeoff

The correctness change removes two of #805's headline String speedups. Standard JMH comparisons used the same benchmark data and harness on `main` (`fa62a17c3192b3f07ad19af9583fcca3255c022c`) and this patch's source tree (identical to measured snapshot `1f873e03951be03d981727f0ae3fc88c17398218`), on JDK 26.0.2. Each ratio is **patched time / main time**, geometrically averaged over the 1K, 10K, and 100K rows with equal weight; below 1 means faster. Standard mode used two forks, two 500 ms warmups, and five 500 ms measurements. UTF-8 has no capture-variant workload in this plan.

| API | Workload and outcome | Paired rows | Patched/main geomean |
|---|---|---:|---:|
| String | `multiInfixLog` match | 3 | 0.991 (near parity) |
| String | `multiInfixLogCaptures` match | 3 | 1.000 (near parity) |
| String | `multiClauseSequence` match | 3 | 2.134 (takes about 2.13× as long) |
| String | `structuredJsonPath` match | 3 | 0.987 (near parity) |
| String | All four no-match families | 12 | 1.008 (near parity) |
| UTF-8 | `multiInfixLog` match | 3 | 1.020 (near parity) |
| UTF-8 | `multiClauseSequence` match | 3 | 1.003 (near parity) |
| UTF-8 | `structuredJsonPath` match | 3 | 0.985 (near parity) |
| UTF-8 | All three no-match families | 9 | 1.007 (near parity) |

The important 100K matching comparisons were also run with the longer schedule (two forks, three 1 s warmups, five 1 s measurements). The table shows JMH point estimates and reported errors; near-parity differences have overlapping uncertainty.

| API | 100K matching workload | Main ns/op | Patched ns/op | Patched/main |
|---|---|---:|---:|---:|
| String | `multiInfixLog` | 343,184 ± 11,111 | 336,277 ± 9,503 | 0.980 |
| String | `multiClauseSequence` | 121,912 ± 1,364 | 275,978 ± 6,173 | 2.264 |
| String | `structuredJsonPath` | 543,981 ± 11,058 | 548,095 ± 12,075 | 1.008 |
| UTF-8 | `multiInfixLog` | 375,574 ± 13,508 | 373,469 ± 7,831 | 0.994 |
| UTF-8 | `multiClauseSequence` | 96,259 ± 1,435 | 91,863 ± 6,571 | 0.954 |
| UTF-8 | `structuredJsonPath` | 397,861 ± 26,131 | 408,966 ± 3,351 | 1.028 |

A separate standard-mode ablation compared the submitted #805 head (`925a027289d896c6cf990fb0cb3f4ed16ea49cbd`) directly with the patched tree on matching 100K String inputs. These ratios are **patched/submitted-head**:

| Workload | Submitted head ns/op | Patched ns/op | Patched/submitted head |
|---|---:|---:|---:|
| `multiInfixLog` | 109,766 ± 3,052 | 337,349 ± 8,999 | 3.073 |
| `multiClauseSequence` | 273,798 ± 2,334 | 274,408 ± 4,079 | 1.002 |
| `structuredJsonPath` | 85,639 ± 910 | 561,255 ± 44,266 | 6.554 |

The two variable-leading-gap workloads lose their acceleration because they now take the general path. The `multiClauseSequence` String slowdown relative to `main` was already present at the submitted head; this patch does not measurably add to it. I have not profiled that interior-gap path, so I cannot attribute its cost more specifically. The external OpenJDK-derived suite and full benchmark collection were not run.

### Verification

The full SafeRE suite and generated public-API crosscheck passed with `mvn verify -pl safere,safere-crosscheck -am -Pcrosscheck-public-api-tests -q`. Focused regression and diagnostics tests, the updated `MatchFuzzer` target, work-counter checks, `spotless:check`, `pmd:check`, and `git diff --check` also passed. A read-only review-fix-loop pass found no actionable P2-or-higher defect in this patch. Those checks establish the local repair, not a recovered performance benefit.

This is a draft because #805's optimization case needs another decision: recover a useful speedup while preserving these semantics, narrow eligibility further, or revise the performance claim. Liam can take the approach from here.

### DIFF
--- a/safere/src/main/java/org/safere/GapScanner.java
+++ b/safere/src/main/java/org/safere/GapScanner.java
@@ -243,7 +243,7 @@ final class GapScanner {
       return cur;
     }
     if (gap.maxLength() != Integer.MAX_VALUE) {
-      return (int) Math.min(maxPos, (long) fromPos + (long) gap.maxLength() * 2);
+      return (int) Math.min(maxPos, fromPos + (long) gap.maxLength() * 2);
     }
     return maxPos;
   }
@@ -303,7 +303,7 @@ final class GapScanner {
       return cur;
     }
     if (gap.maxLength() != Integer.MAX_VALUE) {
-      return (int) Math.min(maxPos, (long) fromPos + (long) gap.maxLength() * 4);
+      return (int) Math.min(maxPos, fromPos + (long) gap.maxLength() * 4);
     }
     return maxPos;
   }
--- a/safere/src/main/java/org/safere/GapScanner.java
+++ b/safere/src/main/java/org/safere/GapScanner.java
@@ -539,6 +539,12 @@ final class GapScanner {
             yield false;
           }
           if (gap.guardBytes().length == 1 && gap.isPureComplement()) {
+            if (len < gap.minLength()) {
+              yield false;
+            }
+            if (gap.maxLength() == Integer.MAX_VALUE && gap.minLength() == 0) {
+              yield true;
+            }
             int count = Character.codePointCount(text, from, to);
             yield count >= gap.minLength() && count <= gap.maxLength();
           }
@@ -612,6 +618,12 @@ final class GapScanner {
             yield false;
           }
           if (gap.guardBytes().length == 1 && gap.isPureComplement()) {
+            if (len < gap.minLength()) {
+              yield false;
+            }
+            if (gap.maxLength() == Integer.MAX_VALUE && gap.minLength() == 0) {
+              yield true;
+            }
             int count = 0;
             for (int i = from; i < to; ) {
               long decoded = scanner.decodeForward(i);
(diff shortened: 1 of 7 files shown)


cushon:
Most of this is obsolete now that the gap engine is removed, but the fuzzing improvements look worthwhile, I created https://github.com/eaftan/safere/pull/916 based on that if you want to fold that back in here or just merge the other PR if that's easier.

eaftan:
Makes sense.  Closing this in favor of #916.

PR #856, opened 2026-09-13
Derive single-line gap guard bytes from the gap, not the flags

cushon:
Fixes an issue in `MultiAnchorCompiler` where guard bytes for `SINGLE_LINE_ANY_STAR` gaps incorrectly included `\r` for gaps that are permitted to match it, such as `[^\n]` and `.` under `Pattern.UNIX_LINES`. 

Guard bytes specify characters that a gap cannot cross, and `isPureComplement` assumes every character except the guard bytes is valid. Naming `\r` on a gap that can legally cross `\r` causes accelerated scans to halt at carriage returns, introducing latent false negatives.

---

`Gap.extractGuardBytes` previously returned `{'\n', '\r'}` unconditionally for every `SINGLE_LINE_ANY_STAR` gap, and `classifyGap` consulted `UNIX_LINES` in only 1 of its 10 construction sites:

1. **Character classes vs. `UNIX_LINES`**: `[^\n]` is an explicit character class that excludes only `\n` and matches `\r`. Unlike `.`, character classes are unaffected by `UNIX_LINES`. However, `classifyGap` checked `UNIX_LINES` for `[^\n]*` and, under default flags, assigned it `{'\n', '\r'}`.
2. **Omitted `UNIX_LINES` checks**: For `+`, `?`, `{n,m}`, and bare `.`, `classifyGap` never inspected `UNIX_LINES`. As a result, expressions like `.+` or `.?` under `UNIX_LINES` (which legally match `\r`) were also given `{'\n', '\r'}`.
3. **Latent defect**: While `SINGLE_LINE_ANY_STAR` gaps are not yet eligible as executable leading/trailing/interior gaps at HEAD, this defect would cause silent false negatives as soon as executor eligibility is expanded to single-line gaps.

---

* **Safe Fallback in `extractGuardBytes`**: `Gap.extractGuardBytes` now returns `null` for `SINGLE_LINE_ANY_STAR` rather than guessing `{'\n', '\r'}`. If a producer cannot soundly prove whether `\r` is a barrier (e.g., compound homogeneous mixture paths), it receives no guard bytes, safely foregoing vector skipping without risking false negatives.
* **Explicit Factory Methods**: Added `Gap.singleLineAnyStar(min, max, greedy, crossesCarriageReturn)` and `Gap.anyStar(min, max, greedy)`. `crossesCarriageReturn` selects between `{'\n'}` and `{'\n', '\r'}`.
* **Unified Quantifier Handling**: Collapsed repetitive quantifier branches (`*`, `+`, `?`, `{n,m}`) in `classifyGap` into a single `repeatedGap(...)` helper:
  * For `ANY_CHAR` (`.`): derives `crossesCarriageReturn` from `isUnixLines(re, sub, flags)`.
  * For `CHAR_CLASS` (e.g. `[^\n]` vs `[^\n\r]`): derives `crossesCarriageReturn` from `sub.charClass.contains('\r')` directly on the character class, decoupling it from `UNIX_LINES`.
* **Interning Parity**: Ensured `.{0,}` and `.*` intern to the same canonical `Gap` instances.


eaftan (approved):
LGTM, thanks for the fix!

PR #860, opened 2026-09-13
Document intentional quoting divergence inside comments

### LINKED ISSUE #858: CrosscheckException in fuzzing - Comments mode `(?x)` discards `\Q` inside `#` comments, failing to quote subsequent lines
This was found in fuzzing, not in real world code.

There are some intentional divergences with `(?x)`, but I think this might be distinct. I'm not sure if this should be fixed in SafeRE or if it's another `java.util.regex` quirk.

---

When comments mode is enabled (`(?x)` or `Pattern.COMMENTS`), `#` begins a comment that runs to the end of the line.

If `\Q` occurs within a `#` comment line without a matching `\E` on the same line, `java.util.regex.Pattern` (and Perl) treats `\Q` as a quoting construct that continues onto subsequent lines. Metacharacters on subsequent lines are escaped and treated as literal characters, and any subsequent `\E` closes the quote.

SafeRE's `Parser.skipCommentsAndWhitespace()` skips characters from `#` to the line terminator as raw comment trivia without recognizing `\Q`. Consequently, SafeRE does not activate quoting for subsequent lines. Metacharacters following the comment are parsed as regular regex syntax (leading to unexpected `missing closing )`, `missing argument to repetition operator`, etc.), and any eventual `\E` throws `PatternSyntaxException: invalid escape sequence`.

```java

public class Repro {

    Pattern pattern = Pattern.compile("(?x)#\\Q\n(");
    pattern.matcher("(").matches();
  }
}
```

```
Exception in thread "main" org.safere.crosscheck.UnsupportedPatternException: SafeRE does not support this pattern (linear-time constraint): missing closing ) near index 9
(?x)#\Q
(
	at org.safere.crosscheck.UnsupportedPatternException.fromCause(UnsupportedPatternException.java:43)
	at org.safere.crosscheck.Pattern.compile(Pattern.java:99)
	at org.safere.crosscheck.Pattern.compile(Pattern.java:64)
	at org.safere.crosscheck.Repro.main(Repro.java:6)
Caused by: java.util.regex.PatternSyntaxException: missing closing ) near index 9
(?x)#\Q
(
```

eaftan:
A `\Q` inside a comments-mode `#` comment causes the JDK to quote syntax on subsequent lines. SafeRE preserves its existing interpretation that ignored comment text cannot affect later syntax: the reported pattern remains an unclosed-group error, and replacing the final `(` with `.` leaves a wildcard.

Document this as an intentional divergence, including the unspecified precedence in the JDK documentation, supporting PCRE2/Rust/Python/.NET observations, and Perl's distinction between regex literals and patterns supplied through variables. Link to the maintainer's compatibility rationale in the issue.

Add `CommentQuotingModelTest` to preserve this decision across compile/inline/scoped flags, LF/CR/CRLF, wildcards, captures, character classes, subsequent quoting, and syntax errors. Expectations run against SafeRE and are disabled only in generated crosscheck tests using `@DisabledForCrosscheck`. No implementation behavior changes.

Verification passed:
- `mvn -pl safere -Dtest=CommentQuotingModelTest,CommentsTest test -q`
- `mvn test -pl safere,safere-crosscheck -am -Pcrosscheck-public-api-tests -Dtest=CommentQuotingModelTest,CommentsTest -Dsurefire.failIfNoSpecifiedTests=false -q` — confirmed model tests execute in SafeRE and are skipped in generated crosscheck.
- `git diff --check`

The focused tests ran before the final documentation/Javadoc clarification; formatting was checked afterward. Full SafeRE and full public API crosscheck suites, fuzzing, and benchmarks were not run.

Fixes #858.


cushon (approved):

safere/src/test/java/org/safere/CommentQuotingModelTest.java
```diff
@@ -0,0 +1,75 @@
...
+    for (String invalid : new String[] {"(", "*", ".\\E"}) {
+      String regex = prefix + "#\\Q" + newline + invalid + suffix;
+      assertThatThrownBy(() -> Pattern.compile(regex, flags))
+          .as("%s: syntax after ignored comment: %s", description, invalid)
+          .isInstanceOf(PatternSyntaxException.class);
+    }
+  }
+
+  private static Stream<Arguments> commentContexts() {
+    return Stream.of("\n", "\r", "\r\n")
```
cushon, line 65: I think Unicode line terminators are also supposed to terminate comments, if you wanted to be exhaustive I think this could also include `\u0085` (NEL), `\u2028` (LS), and `\u2029` (PS)

cushon (commented):
Thanks, this LGTM.

Do you think it makes sense to try to handle this in `FuzzSupport.compileCompatibleOrSkip()` to avoid false positives fuzzing on this?

PR #861, opened 2026-09-13
Dispatch every vector scan on its own search window

cushon:
This replaces #845, but targets main, since the other series is on hold.

This consolidates the six `VectorScanProviders#providerFor*Length` accessors
and two availability predicates into a unified API:

    VectorScanProviders.providerFor(ScanKind kind, int windowLength)
    VectorScanProviders.vectorProviderAvailable()

The old `providerForLength(int length)` did not say whether `length` was the
whole input or the window the call would actually examine. Making that
explicit turned up four places that were dispatching on the input size:

- `Utf8InputScanner` cached one provider at construction, chosen from the
  whole input length, and reused it for every scan on that scanner.
  `lastIndexOfAscii`, the code-point class scans and the ignore-case scans
  all inherited a decision made for a window they weren't scanning. Some
  applied a windowed `minimumInputLength()` check, but against a provider
  already selected by the wrong length. The field is gone; each site now
  selects per window.

- `Utf8StartAccelerator.Teddy.findCandidate` and `MultiLiteral.findCandidate`
  passed `scanner.length()` but scan `[fromIndex, length)`. As the driver
  advances the window shrinks and the input does not.

- `MultiAnchorDescriptor.Anchor.Alternation.findUtf8` had the same bug.

- `Utf8InputScanner.indexOfCodePointClass` chose between the SWAR multi-range
  kernel and the general class scan based on whether a provider existed for
  the whole input, so a narrow multi-range ASCII class in a long input
  declined SWAR and fell through to the scalar code-point loop.


eaftan (approved):
This is a nice cleanup. Replacing the separate provider selectors and cached whole-input choice with a single scan-kind and window-length decision makes the dispatch policy clearer and easier to maintain. The revised call sites use each scan's actual window, reducing the chance of selecting a provider based on the wrong length, and the SWAR and scalar fallbacks remain reachable.

eaftan:
PR Shepherd is now monitoring this PR at `0512b12966f266a905dbe1026045d7344027834c`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `0512b12966f266a905dbe1026045d7344027834c`.

PR #862, opened 2026-09-13
Make vector scan dispatch observable in tests

cushon:
This replaces #847, since the other series is on hold.

This change adds a `ScanAudit` class, somewhat like `WorkCounter`, to record
per scan call which kind, direction, window length and implementation tier a
dispatch site ran. This allows adding test coverage for dispatch and window
handling.

    record ScanEvent(ScanKind kind, ScanDirection direction,
                     int windowLength, ScanPath path) {}

`ScanPath` is `VECTOR`, `SWAR`, `SCALAR`, `DECLINED` (a provider was consulted,
nothing ran), or `CONSULTED`. The last is not a path at all: `providerFor`
records it automatically, so a dispatch site that never records a path of its
own shows up as an unpaired consultation instead of going silently unaudited.

#861 fixes an example of the kind of issue this coverage is meant to catch.

`ScanAudit.ENABLED` is a `static final boolean` read once from
`org.safere.testing.scanAudit`, so C2 folds every `record` call away in a
normal run. The guard lives inside `record` rather than at each call site,
which keeps a ladder to one statement per exit.

This deliberately does not follow `WorkCounterConfig.ENABLED`, which is
generated at build time behind a Maven profile. That machinery exists because
a work counter fires once per byte scanned, where even a folded branch in the
interpreter matters; a dispatch event only fires once per scan call. The two
are also mutually exclusive observers `captureForTesting` throws under
`-Pwork-counters`, which bypasses every ladder the audit is trying to watch.

The one-line `safere-crosscheck/pom.xml` change excludes the audit test from
public API generation: it depends on package-private engine types and cannot
compile once crosscheck relocates test sources.


eaftan (approved):
I found two event-accounting issues: routing probes could appear as scan consultations, and the alternation route could omit its scalar fallback. I pushed a commit to fix both.

I ran a quick before/after JMH check of six short UTF-8 workloads and found no meaningful changes to performance. LGTM

eaftan:
PR Shepherd is now monitoring this PR at `0a9fe3092f260db27575f0fad35b81392fec8e50`.

eaftan:
This PR is no longer tracked by PR Shepherd because its head moved to `7f985f1586c454e043db72a2dd54278b9a2f7dab` without a verified PR Shepherd push. PR Shepherd cannot merge this head. To resume tracking, post a new comment containing only `/shepherd merge 7f985f1586c454e043db72a2dd54278b9a2f7dab`.

PR #864, opened 2026-09-13
Handle intentional comment quoting differences in fuzzing

eaftan:
Differential fuzzing still reported the intentional comment/quoting divergence documented in #860: `(?x)#\Q\n(` can differ at compilation, while the dot variant compiles in both engines but matches differently. Exclude this family in `FuzzSupport.compileCompatibleOrSkip()` after compiling both patterns, so unexpected non-syntax failures still surface.

The test-only scanner tracks comment and quote context, backslash pairing, scoped COMMENTS/UNIX_LINES flags, nested character classes, and control-escape operands. It conservatively excludes quote openers closed within the same comment too. Regression tests verify that ordinary quoting, escaped hashes, literal mode, and disabled comments remain eligible for comparison. Add comment-quote variants to the parser fuzz generator.

Extend `CommentQuotingModelTest` to NEL, LS, and PS across compile-time, inline, and scoped flags. These characters terminate comments but remain literal regex atoms, unlike LF/CR, so matching and character-class expectations account for that distinction. No library implementation behavior changes.

Addresses the specific post-merge feedback on #860. Refs #858; the broader continuous-fuzzing/false-positive policy discussion is deferred.

Validation on OpenJDK 26.0.1:
- Confirmed the new fuzz-helper regression tests failed before the exclusion, then passed with it.
- Passed `mvn -pl safere-fuzz -am -Dtest=CommentQuotingModelTest,CommentsTest,FuzzSupportCommentQuotingTest,FuzzSupportCaptureWaiverTest,FuzzSupportOracleTimeoutTest,ParserCompatibilityFuzzer,EscapeSyntaxFuzzer -Dsurefire.failIfNoSpecifiedTests=false test -q` (fuzz targets in regression mode).
- Passed `mvn test -pl safere,safere-crosscheck -am -Pcrosscheck-public-api-tests -Dtest=CommentQuotingModelTest,CommentsTest -Dsurefire.failIfNoSpecifiedTests=false -q`; confirmed model tests execute in SafeRE and are skipped in generated crosscheck. This ran before the final fuzz-scanner-only refinement; the focused test/fuzzer command above ran afterward.
- Passed `mvn -pl safere,safere-fuzz spotless:check -q` and `git diff --check`.
- Full SafeRE and full public API crosscheck suites, sustained fuzzing, and benchmarks were not run.


eaftan:
Honestly, I'm not super satisfied with this approach.  There's a lot of machinery in this PR to detect the cases that we want to intentionally diverge on.  Let me take a step back and see if I can come up with a more maintainable way to do this.

Issue #461, opened 2026-06-02
Single-character classes diverge when region ends inside surrogate pair


eaftan:
SafeRE currently matches single-character classes when an opaque matcher region ends inside a valid surrogate pair, while `java.util.regex` does not.

This was noticed while reviewing whether draft PR #390 can be abandoned. The original `\S`/unpaired-surrogate repro from #375 is fixed on current `main`, but PR #390 also covered this broader region-boundary case, which still diverges.

## Repro

```java
String text = "\uD83D\uDE00";

for (String regex : new String[] {"[^a]", ".", "\\P{Cs}", "[^\\p{Cs}]"}) {
  var jdk =
      java.util.regex.Pattern.compile(regex, java.util.regex.Pattern.DOTALL)
          .matcher(text)
          .region(0, 1);
  var safere =
      org.safere.Pattern.compile(regex, org.safere.Pattern.DOTALL)
          .matcher(text)
          .region(0, 1);

  System.out.println(regex + " jdk=" + jdk.find() + " safere=" + safere.find());
}
```

## Observed on current main

```text
[^a] jdk=false safere=true
. jdk=false safere=true
\P{Cs} jdk=false safere=true
[^\p{Cs}] jdk=false safere=true
```

## Notes

- `origin/main` was at `23c614131606e54e9f590da437fd671d38ca41c4`.
- The exact #375 repro now matches JDK on `main`: both find `\S` at index 34.
- PR #390 contains draft coverage for this broader behavior in `MatcherTest` and `UnicodeMatchBoundsTest`, but the PR is old and merge status is dirty.

PR #869, opened 2026-09-14
Decode scalar atoms within split-surrogate matcher regions

### LINKED ISSUE #867: Opaque-region character matching depends on a surrogate outside the region
## Problem

Ordinary character matching in an opaque region depends on whether a character **outside** the region completes a surrogate pair. Revisit the intentional divergence [“Quantified Scalar Atoms at Split Surrogate Region Ends”](https://github.com/eaftan/safere/blob/eccf4eeffca9bf63c01029e861cf3c83051f5d1d/INTENTIONAL_DIVERGENCES.md#quantified-scalar-atoms-at-split-surrogate-region-ends).

For a region containing only a high surrogate, SafeRE's `.` matches if the input ends there or the next character is `x`, but fails if the next character is a low surrogate—even though that next character is outside the opaque region. ICU consistently treats the exposed high surrogate as unpaired within the region and matches it.

The existing intentional divergence correctly seeks consistent behavior between an atom and its quantified form. However, consistency alone does not establish that both should reject the region. Both accepting is also coherent and appears more consistent with Java's bounded-text APIs.

Related history: #461 previously treated SafeRE accepting single-character classes at a split surrogate region end as a compatibility bug because JDK atoms rejected them. This issue revisits that premise using the broader evidence below; it should not simply restore all previous class behavior.

## Reproducer and observations

Tested with SafeRE checkout `eccf4eeffca9bf63c01029e861cf3c83051f5d1d`, OpenJDK `26.0.1+8-34`, and ICU 78.3. All Java cases use default flags, default opaque/anchoring bounds, and `matches()` over `[0, 1)`.

```java
public final class SplitSurrogateRegionRepro {
    String[] inputs = {"\uD83D", "\uD83D\uDC4D", "\uD83Dx"};
    String[] names = {"isolated high", "paired low outside region", "x outside region"};
    for (int i = 0; i < inputs.length; i++) {
      for (String regex : new String[] {".", ".*", "\\p{Cs}"}) {
        boolean safe = org.safere.Pattern.compile(regex)
            .matcher(inputs[i]).region(0, 1).matches();
        boolean jdk = java.util.regex.Pattern.compile(regex)
            .matcher(inputs[i]).region(0, 1).matches();
        System.out.printf("%s / %s: SafeRE=%s JDK=%s%n", names[i], regex, safe, jdk);
      }
    }
  }
}
```

| Input / region | Pattern | SafeRE | JDK | ICU |
|---|---|---:|---:|---:|
| High surrogate only / entire input | `.` | true | true | true |
| High surrogate only / entire input | `.*` | true | true | true |
| High surrogate only / entire input | `\p{Cs}` | true | true | true |
| Full emoji / region containing only high surrogate | `.` | false | false | true |
| Full emoji / region containing only high surrogate | `.*` | false | true | true |
| Full emoji / region containing only high surrogate | `\p{Cs}` | false | false | true |
| High surrogate followed by `x` / region containing only high surrogate | `.` | true | true | Not probed |
| High surrogate followed by `x` / region containing only high surrogate | `.*` | true | true | Not probed |
| High surrogate followed by `x` / region containing only high surrogate | `\p{Cs}` | true | true | Not probed |

The no-region standalone-high-surrogate case was also tested: all three engines accept `.` / `.*` / `\p{Cs}`. ICU probes constructed the input directly as UTF-16, with no UTF-8 conversion or replacement of the surrogate. Its successful matches span `[0, 1)` with no error. The split case used native `RegexMatcher::region(0, 1, status)` on a two-unit `UnicodeString` containing `{0xD83D, 0xDC4D}`.

Joni and PCRE2 results from the earlier comparison do not resolve this API question: the Joni UTF-8 adapter replaced an isolated surrogate, and PCRE2 UTF mode rejected a truncated sequence. Neither was a comparable native Java-region test.

## Specification assessment

The proposed expectation is **region-local decoding for ordinary consuming atoms**, so `.` and `.*` both match the exposed high surrogate. This is an interpretation of the API, not an explicit surrogate-boundary requirement in the regex Javadoc:

- [`Matcher.region()`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/regex/Matcher.html#region(int,int)) defines the region as the portion searched and allows indices that split surrogate pairs. [`matches()`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/regex/Matcher.html#matches()) matches the entire region.
- [`useTransparentBounds()`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/regex/Matcher.html#useTransparentBounds(boolean)) explicitly governs lookahead, lookbehind, and boundary constructs. It does **not** expressly specify character decoding or promise universal equivalence between regions and substrings. Do not overstate the specification argument as an unambiguous prohibition covering every operation.
- [`Character.codePointAt(char[], index, limit)`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/lang/Character.html#codePointAt(char%5B%5D,int,int)) supplies a strong Java precedent: it returns the high surrogate by itself when the low surrogate is beyond `limit`. This supports region-local code-point decoding without making the engine operate on code units generally. It is an analogy, not a requirement that Matcher call that method.
- [UTS #18 supplementary-character handling](https://www.unicode.org/reports/tr18/#Supplementary_Characters) requires treating surrogate pairs as code points, but does not settle whether pairing at a matcher-region boundary uses the full input or only the selected region.

SafeRE's full-input pairing model is internally coherent. Its weakness is that an ordinary atom's result inside an opaque region changes with hidden suffix text. Our assessment favors ICU's region-local model, with the above specification limits acknowledged.

## Work items

- Reassess the intentional divergence and adopt/document the desired region-decoding rule. The proposed direction is region-local decoding for consuming atoms, while retaining code-point treatment of complete pairs within the region.
- Before changing implementation, add systematic regression coverage for split region starts and ends, isolated surrogates, ordinary and quantified atoms, positive and negated classes, and relevant matching operations. Include controls that vary only text outside the region and compare against the corresponding substring where appropriate.
- Keep character-category semantics correct: if a split high surrogate is treated as unpaired, `\p{Cs}` should recognize it and `\P{Cs}` should exclude it. Do not indiscriminately accept every previously rejected class.
- Assess transparent-boundary assertions separately; their permission to inspect outside text should not accidentally redefine the consuming atom's policy. Preserve linear-time matching across execution engines.
- Update intentional-divergence documentation and affected crosscheck/fuzz expectations. Reassess `QUANTIFIED_SPLIT_SURROGATE_SCALAR_COMPOSITION`; separately evaluate the related `BOUNDARY_ANY_CLASS_SPLIT_SURROGATE_SCALAR_COMPOSITION` family rather than assuming all of it is resolved.

Refs #461. Related to the oracle investigation in #858. This issue records targeted probes and source/specification analysis; no production fix or full test-suite validation was performed.

eaftan:
A matcher region ending between a surrogate pair exposes only the high surrogate, but ordinary atoms could decode the complete pair using text outside the region. This change makes scalar consumption respect the logical region: `.` and `.*` consume the exposed surrogate, `\p{Cs}` recognizes it, and `\P{Cs}` excludes it. Complete pairs inside the region remain single code points.

The rule also applies to scalar alternatives in programs containing `\X`. Full matches remain region-end strict, alternative ordering and captures are preserved, and Unicode word boundaries use the code-point category visible within opaque bounds. The JDK contract leaves surrogate decoding at region edges unspecified, and observed JDK behavior varies by atom/quantifier; the intentional-divergence documentation records SafeRE's consistent region-local rule and the remaining transparent-boundary distinction.

Eligible scalar regions use BitState with explicit consumption, boundary, and anchor bounds shared with NFA. Decoding and search-candidate advancement respect those bounds without copying the region. Existing bitmap and work-budget limits retain NFA fallback for larger or more expensive searches. Grapheme programs use NFA; explicit cross-queue priority tracking is enabled only when scalar and grapheme consumption limits differ. Ordinary grapheme searches retain queue ordering without maintaining those extra links.

Regression coverage includes both surrogate halves, captures and alternative priority, String and UTF-8 engine equivalence, bounds settings, matcher reuse, allocation scaling, and work scaling across BitState-sized and NFA-sized ranges. The region fuzzer covers scalar captures and mixed grapheme programs. Scalar divergence-sweep classification recognizes simple atom wrappers; more complex wrappers remain unclassified for review. The unrelated existing NFA optional-lazy capture defect is tracked in #880 with a disabled regression.

### Performance

Sequential targeted JMH runs used identical workload declarations through `./run-java-benchmarks.sh --fastbuild --long`: 2 forks, 3 × 1 s warmup, and 5 × 1 s measurement. The comparison below is the branch base `595ebc5dd659e07f064e610b6064baa353c1b5a6` against the production change in `3590a2d374cde82df2ffc1a61c7466141f0088ad`. Measurements preceded the final commit; subsequent production edits changed comments only.

Times are microseconds per retained-matcher `.*` region `find()`, with the region ending between the final surrogate pair:

| Input UTF-16 units | Base | This PR | Base / PR |
| --- | ---: | ---: | ---: |
| 256 | 11.097 | 2.581 | 4.30× |
| 4096 | 172.283 | 39.813 | 4.33× |

The operation also changes its result: the base omits the exposed high surrogate, while this PR consumes it. These timings therefore compare the old and corrected behavior, not identical match results. Warm allocation in the PR is approximately 64 B/op at both sizes; BitState retains reusable bitmap/job-stack storage. These results apply to the targeted eligible scalar workloads, not all region matching.

A direct branch-base-versus-final-head grapheme benchmark comparison has not been collected. The six declared String/ASCII grapheme find-all workloads cover `\X`, `(?:\X|.)`, and `(?:.|\X)` at lengths 256 and 4096; they do not establish performance across Unicode clusters, split regions, or UTF-8. No whole-PR grapheme performance-neutrality claim is made.

### Validation

Passed:

- Focused region engine-equivalence, diagnostics, BitState, boundary, surrogate-region, and grapheme tests.
- Full SafeRE tests in the library phase of `mvn verify -pl safere,safere-crosscheck -am -Pcrosscheck-public-api-tests -q`.
- Public API crosscheck: `mvn verify -pl safere-crosscheck -Pcrosscheck-public-api-tests -q`.
- Work scaling: `mvn -pl safere -Pwork-counters -Dtest=MatcherLinearTimeTest#splitEndRegionFailuresStayLinearAcrossEngineFallback test -q`.
- Region fuzzer regression mode: `mvn -pl safere-fuzz -am -Dtest=RegionBoundsFuzzer -Dsurefire.failIfNoSpecifiedTests=false test -q`.
- Targeted standard/long JMH and allocation profiling; CPU profiling informed the implementation.
- Full-branch read-only review: no actionable findings.

Not run: full divergence sweeps, extended fuzz campaigns, full benchmark collections, UTF-8 timing comparisons, or the external OpenJDK benchmark suite. Region work-scaling tests cover size-based NFA fallback but do not explicitly assert work-budget-triggered fallback.

Fixes #867
Refs #461


eaftan:
Targeted JMH follow-up on the review fixes: baseline `d27c1f0fc089bdf82967e511ffc382c2833e9dc7` versus current PR head `83b97fae0242f92d77ff0e8ace898994077de70c`. Both revisions used identical workload declarations and sequential runs through `./run-java-benchmarks.sh`. These results isolate the review fixes within this PR.

The `--long` run (2 forks, 3 × 1 s warmup, 5 × 1 s measurement) found:

| SafeRE String workload | Before | After | After/before |
| --- | ---: | ---: | ---: |
| `.*` region `find`, split surrogate at end, 256 UTF-16 units | 3,147.5 ns/op | 10,812.4 ns/op | 3.44× |
| `.*` region `find`, split surrogate at end, 4096 UTF-16 units | 48,389.9 ns/op | 183,058.0 ns/op | 3.78× |

Across six grapheme `findAllCount` cases (`\X`, `(?:\X|.)`, and `(?:.|\X)` on 256 and 4096 ASCII characters), the equal-weight geometric mean of after/before time was **1.105** (about **10% more time**); individual ratios ranged from 1.07× to 1.15×. Standard runs showed similar ratios.

There is an allocation benefit in the split-end case: the standard run with JMH `-prof gc` measured **864.04 → 88.15 B/op** at 256 units and **12,384.69 → 92.14 B/op** at 4096 units.

I'm rethinking this implementation in light of the latency slowdowns, especially the split-end path. The next step is to profile it and look for a way to preserve the bounded-allocation behavior at lower cost.

PR #875, opened 2026-09-14
Run the random differential fuzzer through the UTF-8 domain

cushon:
`RandomTest` generates 200 patterns and 100 strings per method and compares SafeRE against
`java.util.regex`, but it only ever ran the String API — the file had no reference to
`Utf8Input`, `Utf8Matcher` or `getBytes`.

There is a String↔UTF-8 oracle already, at `Utf8EnginePathEquivalenceTest:36–38`, but it
runs about a dozen hand-written fixtures and is `@DisabledForCrosscheck`;
`Utf8CrosscheckTest` is fixed strings. So the oracle and the randomized population were
disjoint, and no random input ever reached the byte side.

That matters because the two domains are mirrored implementations —
`StringStartAccelerator`/`Utf8StartAccelerator`, `StringInputScanner`/`Utf8InputScanner`,
and both `RejectPrefilter` domains — and `UTF8.md` says only the input representation and
match coordinates change.

## Changes

* Each generated `(pattern, text)` now also runs through `Utf8Input`. The String side is
  the oracle rather than the JDK, since the JDK has no byte-domain API and the existing
  checks already pin the String side to it.
* The comparison walks the whole `find()` sequence and every capture group, not just the
  first match. A byte-domain implementation can agree on the first match and still drift on
  iteration, on empty-match advancement, or on group bounds. String offsets are mapped to
  byte offsets so coordinates are compared directly rather than only matched text.
  `matches()` and `lookingAt()` are compared on fresh matchers.
* Adds `complicatedLongInput`, an arm with longer inputs and fewer strings.

## Why the long-input arm

The existing corpus caps text at 20 characters, which cannot dispatch any wide kernel:
`IncubatorVectorScanProvider` requires a 1024-byte window for `CLASS`, `IGNORE_CASE` and
`TEDDY`, and `ByteSwarScan.MIN_FILTER_LENGTH` is 64. Measured with `ScanAudit` over the
UTF-8 domain:

| corpus | dispatches | max window | kinds | vector path |
| :--- | ---: | ---: | :--- | ---: |
| existing, ≤20 chars | 1,066 | 30 B | `BYTE`, `PAIR` | 0 |
| `complicatedLongInput` | 41,960 | 2,313 B | `BYTE`, `PAIR`, `CLASS`, `MULTI_LITERAL` | 18,218 |

Without it the byte-domain comparison only covers scalar and SWAR paths.

`TEDDY` and `IGNORE_CASE` are still unreached — they need case-insensitive patterns and
multi-literal alternations this generator does not produce. Supplementary characters are
out of scope: every alphabet here is BMP-only, so the documented divergence in
`Utf8CrosscheckTest.emptyFindSkipsJdkPositionsInsideSupplementaryScalars` cannot arise, and
covering it would mean the comparator has to model that divergence rather than flag it.

## Validation

Mutating `Utf8StartAccelerator.Literal.findCandidate` to resume one byte late on a non-zero
`fromIndex` — a drift invisible on the first match — drove three of the four short-input
methods straight to the 50-failure cap:

```
String/UTF-8 disagree: pat="bb*?" text="bbabaacbbcc"
  match #1 group 0: string=[1,2) -> expected utf8=[1,2) but got [3,4)
String/UTF-8 disagree: pat="(b)." text="cabaccabbbc"
  find() #2: string=true utf8=false
```


eaftan (approved):
This is a great addition and gives us useful randomized coverage of the UTF-8 matching paths. 

I found that `matches()` and `lookingAt()` compared only success or failure, so capture-bound differences could escape detection. I pushed a fix that compares all group
bounds for those operations as well, with a focused regression test. 

LGTM

eaftan:
PR Shepherd is now monitoring this PR at `dad205b3face092035199775075708bd4b1b49d7`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `dad205b3face092035199775075708bd4b1b49d7`.

PR #865, opened 2026-09-13
Retire the multi-anchor gap executor

cushon:
I don't think this optimization is pulling its weight. I propose dropping it for now, I have some planned follow-ups that will improve existing paths and recover some performance the gap executor was partially working around, and then explore something close to regex-automata's reverse-inner.

As I mentioned in https://github.com/eaftan/safere/pull/805#issuecomment-5655377243

> I think the multi-anchor stuff makes sense as a general approach to prefilters and start/end acceleration.
>
> I think the part that was trying to characterize kinds of gaps, and the MultiAnchorExecutor, was flawed.
>
> Something close to regex-automata's reverse-inner would get most of the same benefits and cover some patterns the gap executor doesn't, but without the complexity of trying to handle arbitrary interior gaps and the surface area of another executor.
>
> Most of the benchmarks where the gap executor is winning were added in PRs that implemented the gap engine, partly with the goal of confirming it was actually working. But benchmarking with the gap executor disabled shows it doesn't apply to very many cases, and it's a wash on the cases that predate it. One of the ones where it wins (fixedAnchorLog) looks like an anchor selection issue more than a case where the gap executor is really needed.

## Changes

Deletes:

* `MultiAnchorExecutor` and `GapScanner`, the `Matcher.doFindCore` block that speculatively invoked them, the 14 `Gap` methods that existed only to delegate into `GapScanner`, and the `multiAnchorGapEngine` engine-path option.
* The `Chain` / `Segment` / `Anchor` model (`Anchor.Alternation`, `selectDriver`, `computeDriverIndices`, `isExecutableChain`)
* The reverse literal-scan layer (`Utf8InputScanner.lastIndexOfAscii`, the `ByteVectorScan`/`ByteSwarScan` reverse kernels)

## Notes

* Some corpus analysis found this is accelerating ~2-4% of patterns.
* Over the 51 reachable trials the executor helps 26, hurts 9, and is indistinguishable on 16. On the suite below the geomean cost of deletion is 18.6%, but that is almost entirely two families: 5.2% excluding uuidExtraction, and 0.6% excluding uuidExtraction and fixedAnchorLog.match. Whole-suite over all 1,259 differential trials, measured separately on x86 with the executor disabled by flag, it was 1.3%.
* Its best workload is `fixedAnchorLog`, `error:\[[A-Z]\] code:500`. The compiler picks the prefix `error:[`, which occurs frequently in the haystack, over `] code:500`, which occurs once. So better anchor selection would solve that regression.
[...]

PR #876, opened 2026-09-14
Compare start-literal candidates on selectivity alone

cushon:
This recovers performance for one of the cases regressed by #865, but in a more general way.

---

`extractStartPlan` arbitrates between two start-acceleration candidates: the
leading literal prefix, and a fixed-offset literal found later in the pattern.
Two suppressors stood in front of that comparison, and between them they kept
the worse candidate on most patterns that had a choice:

- a length gate, `prefix.length() <= 2`, which skipped the comparison entirely
  unless the leading prefix was a single character or a pair;
- a 2× margin, which then discarded most of what the gate admitted.

The gate was the binding constraint. On `error:\[[A-Z]\] code:500` the leading
prefix is `error:[` — seven characters, so the two candidates were never
compared at all. `error:[` occurs on *every line* of a log; the fixed-offset
`] code:500` occurs once.

```diff
-      if (prefix.length() <= 2) {
-        int prefixScore = RarityOracle.literalSelectivityScore(prefix, prefixFoldCase);
-        int folScore = RarityOracle.literalSelectivityScore(fol.literal());
-        if (folScore > prefixScore * 2) {
+      int prefixScore = RarityOracle.literalSelectivityScore(prefix, prefixFoldCase);
+      int folScore = RarityOracle.literalSelectivityScore(fol.literal());
+      if (folScore > prefixScore) {
```

Ties go to the leading prefix, which needs no offset arithmetic to verify.

## Benchmarks

JMH, 100k inputs, `safere-string`, AMD EPYC 7B13 / JDK 26, `-f 1 -wi 3 -i 5`:

| trial | before | after | |
|---|---:|---:|---:|
| `fixedAnchorLog.match` | 457,836 ns | 13,582 ns | **33.7×** |
| `fixedAnchorLog.noMatch` | 13,456 ns | 13,438 ns | 1.00× |
| `uuidExtraction.match` | 11,915 ns | 10,035 ns | **1.19×** |
| `uuidExtraction.noMatch` | 11,722 ns | 9,811 ns | **1.19×** |

These are two of the four families that regressed when the multi-anchor gap
executor was retired in #865.

## Corpus impact

Swept over 46,866 distinct patterns drawn from 61,534 call sites:

- **1,110 patterns (2.38%)** change their chosen start literal, every one of
  them `Literal` → `FixedOffset`.
- Mean chosen-literal length **7.43 → 14.95**.
- `RejectPlan` class changes on 640 patterns.

## Accepted risk

276 of the 1,110 flips land on a **shorter** literal, and **91 on a literal of
one or two characters** — most often `-` (11), `.` (9), `@` (6), `]` (6). None
of those has been measured on a workload.

I deliberately did not add a guard, because the two obvious guards are provably
wrong here:

- **"require the new literal to be at least as long"** would reject
  `uuidExtraction`'s `-` (1 char, versus `id:` at 3) — and that single-character
  literal is where its entire 1.19× comes from. Measured: with that predicate,
  `uuidExtraction` returns to 0.99×.
- **recalibrating `literalSelectivityScore`** so length outweighs single rare
  characters does the same thing for the same reason (see below).

Short is not the same as worse. A one-byte needle is the one shape that scans at
memchr speed — `indexOf('-')` measures 50.7 GB/s against 8.6 GB/s for
`indexOf("id:")` — so the score is implicitly trading selectivity against scan
throughput, and whether `"` beats `category` depends on whether the input is
JSON. No static score can know that.

The durable answer is the one regex-automata uses: abandon a prefilter observed
not to filter. That is a separate change.

## What I tried and rejected

The plan called for adding a length term to `literalSelectivityScore` *first*,
on the grounds that removing the gate alone makes 276 patterns worse. I built
that and measured it, and it does not hold up:

- `literalSelectivityScore` already has a length term (`score += r + 1`). Its
  real defect is that it sums an *ordinal* rarity rank as if it were a
  log-probability, so `'@'` (rank 135) is charged 67× more than `'r'` (rank 2)
  when the true information ratio is nearer 3×. I implemented the principled
  correction, `w(c) = 16·log2(rank + 2)`.
- It moves **47 patterns (0.10%)**, because the score is only ever consulted
  inside `if (prefix.length() <= 2)` — the gate returns first on the rest.
- It moves **no benchmark family**: 1.00× on all four trials above.
- It **loses** `uuidExtraction` (1.19× → 1.00×), for the reason in the previous
  section.

It is worth doing after literal scans anchor on their rarest byte, since the
throughput term the current miscalibration compensates for will be gone. It is
not worth doing before that, and it is not a prerequisite for this change.


eaftan (approved):
The simpler start-literal comparison makes sense as a follow-up to #865. I added a test for String and UTF-8 `find()` bounds with the newly selected fixed-offset plan, including repeated matches after a multibyte prefix. I've pushed that test addition.

I reproduced the reported workloads against #865 and measured the capture and guarded-record cases that matter to the stack decision in a second paired run. `#876/#865` is elapsed time, so lower is better.

| Workload | #865 ns/op | #876 ns/op | #876/#865 | Result |
|---|---:|---:|---:|---|
| fixedAnchorLog.match, 100000 | 294,755 | 12,075 | 0.041 | 24.4× faster |
| fixedAnchorLog.noMatch, 100000 | 11,846 | 11,842 | 1.000 | Flat |
| uuidExtraction.match, 100000 | 11,060 | 8,300 | 0.750 | 1.33× faster |
| uuidExtraction.noMatch, 100000 | 10,848 | 8,869 | 0.818 | 1.22× faster |
| fixedAnchorLogCaptures.match, 1000 | 1,613 | 266 | 0.165 | 6.07× faster |
| guardByteDelimitedRecord.match, 100000 | 284,533 | 314,379 | 1.105 | No clear recovery |

This recovers the fixed-anchor capture regression along with the intended log and UUID cases. It does not recover the guarded-record workload, but given the agreed-upon removal of the gap executor, this is a clear improvement over the previous state.

LGTM

eaftan:
PR Shepherd is now monitoring this PR at `e9baa261647bec14224d0196a4ff821f13fa616a`.

eaftan:
This PR is no longer tracked by PR Shepherd because its head moved to `f8f8a4f2e69c46fd26dde9f3478da9a4274015d6` without a verified PR Shepherd push. PR Shepherd cannot merge this head. To resume tracking, post a new comment containing only `/shepherd merge f8f8a4f2e69c46fd26dde9f3478da9a4274015d6`.

PR #877, opened 2026-09-14
Fix incorrect scm URLs in released module POMs

wendigo:
## Problem

Released module POMs (e.g. `safere-*.pom` published to Maven Central)
have malformed `<scm>` URLs with a duplicated path segment, for
example:

```xml
<scm>
  <connection>scm:git:https://github.com/eaftan/safere.git/safere</connection>
  <developerConnection>scm:git:git@github.com:eaftan/safere.git/safere</developerConnection>
  <url>https://github.com/eaftan/safere/safere</url>
</scm>
```

## Root cause

The parent POM's `<scm>` block is correct:

```xml
<scm>
  <connection>scm:git:https://github.com/eaftan/safere.git</connection>
  <developerConnection>scm:git:git@github.com:eaftan/safere.git</developerConnection>
  <url>https://github.com/eaftan/safere</url>
</scm>
```

But none of the 7 child modules declare their own `<scm>`. Maven
applies "URL inheritance" to `scm/connection`, `scm/developerConnection`,
and `scm/url` (also `project/url` and `distributionManagement/site/url`):
a child that doesn't set its own value doesn't simply copy the
parent's value verbatim — Maven appends the child module's own
artifactId/path segment onto the parent's URL. That's intentional for
projects laid out with the module directory matching a real subpath of
the SCM URL, but wrong here since the parent's `<scm>` already points
at the top-level repo.

Verified with `mvn help:effective-pom -pl <module>` — every module's
effective `<scm>` had its own artifactId appended, e.g.:

| module | effective scm/url before |
|---|---|
| `safere` | `https://github.com/eaftan/safere/safere` |
| `safere-unicode` (artifactId `safere-unicode-generator`) | `https://github.com/eaftan/safere/safere-unicode-generator` |
| `safere-crosscheck` | `https://github.com/eaftan/safere/safere-crosscheck` |
| `safere-fuzz` | `https://github.com/eaftan/safere/safere-fuzz` |
| `safere-exhaustive` | `https://github.com/eaftan/safere/safere-exhaustive` |
| `safere-ffm-re2` | `https://github.com/eaftan/safere/safere-ffm-re2` |
| `safere-benchmarks` | `https://github.com/eaftan/safere/safere-benchmarks` |

This matches exactly what's visible in the currently released POM.

## Fix

Since Maven 3.6.1 ([MNG-6159](https://issues.apache.org/jira/browse/MNG-6159)),
this default path-appending behavior can be disabled per element via
XML attributes on the parent's `<scm>` element itself:

```xml
<scm child.scm.connection.inherit.append.path="false"
     child.scm.developerConnection.inherit.append.path="false"
     child.scm.url.inherit.append.path="false">
  <connection>scm:git:https://github.com/eaftan/safere.git</connection>
  <developerConnection>scm:git:git@github.com:eaftan/safere.git</developerConnection>
  <url>https://github.com/eaftan/safere</url>
</scm>
```

This is a single, one-time change in the parent POM — no need to add
a redundant `<scm>` override to every child module. Re-ran
`mvn help:effective-pom` for all 7 modules after the change: every
module now resolves to the correct, unsuffixed
`https://github.com/eaftan/safere` URLs.

`flatten-maven-plugin` (used for the `release` profile) resolves `scm`
from the effective model at build time, so this fix at the source
(rather than post-processing in flatten config) ensures the flattened
and released POM gets the corrected value too.

## Test plan

- [x] `mvn help:effective-pom -pl <module>` for all 7 modules, before
      and after: confirmed the duplicated path segment is gone.
- [x] `mvn validate` across the whole reactor still passes.


wendigo:
cc @eaftan

eaftan (changes requested):
Thanks @wendigo for catching this!  I'm a little worried about duplicating the same `<scm>` block in all of the child POMs.  Maven supports disabling the inherited path adjustment from the parent POM:

```
<scm
    child.scm.connection.inherit.append.path="false"
    child.scm.developerConnection.inherit.append.path="false"
    child.scm.url.inherit.append.path="false">
  <connection>scm:git:https://github.com/eaftan/safere.git</connection>
  <developerConnection>scm:git:git@github.com:eaftan/safere.git</developerConnection>
  <url>https://github.com/eaftan/safere</url>
</scm>
```

I verified that this produces the correct SCM values in all seven modules’ effective POMs and in the flattened safere release POM. Would you mind switching to this approach?

wendigo:
@eaftan done

eaftan (approved):

PR #878, opened 2026-09-14
Bound the capture-demand speculation to the accelerated start

cushon:
Once a `Pattern` has been observed to consume inner captures, `find()` on an input of 512
characters or fewer runs the capture-aware engine directly instead of locating group 0 with the
DFA and replaying the range through BitState. That attempt was unbounded —
`searchWithBitStateOrNfa(effectiveStart, scanner.length(), …)` — and `BitState.doSearch` walks
candidate starts one code point at a time with no literal acceleration of its own. So it displaced
the forward DFA's vectorised `O(n)` scan, which re-accelerates from its start state inside the
loop, with `O(n × |prog|)` scalar work. It only paid off when the match was at or near the start
the prefilter had already chosen.

This bounds the attempt to that start and falls through to the DFA pipeline when it fails.

```java
if (shouldPreferCaptureEngine(prog, scanner)) {
  int captureSearchLimit = startPositionPreselected ? effectiveStart : scanner.length();
  int[] result = searchWithBitStateOrNfa(prog, scanner, effectiveStart, captureSearchLimit, …);
  if (result != null || captureSearchLimit >= scanner.length()) {
    return applyFullMatchResult(result);
  }
  diagnosticDiscardBoundary(MatchStrategy.BIT_STATE);
  diagnosticDiscardBoundary(MatchStrategy.NFA);
  // fall through to the reverse-DFA / forward-DFA pipeline
}
```

`BitState` and the Pike NFA already accept a bound on where new candidate starts are tried, so only
the argument changes — no engine change was needed. Patterns with no start accelerator keep the
previous behaviour. Falling through is semantics-preserving: no start beyond the accelerated
candidate was tried, and the DFA path searches the full range from `effectiveStart`.

`matches()` and `lookingAt()` are deliberately untouched. Both are anchored at 0 with exactly one
candidate start, so the scan explosion cannot arise there.

Boundary attribution in `DiagnosticAccumulator` is first-wins, so an abandoned attempt would have
permanently claimed credit from the engine that actually decided the match. Added
`discardBoundary(strategy)`, which withdraws the claim only if that strategy still holds it.

## This is a robustness fix, not a speedup — the earlier numbers are retracted

The original description claimed **8.91×** on `fixedAnchorLogCaptures.match.256` (6,513.9 → 730.9
ns). That was measured against #865's head `ad874a7b`, which predates #876. **#876 fixes the same
trial independently and better**, taking it to 224 ns — 3.3× past where this change reached from
the old base.

Re-measured on the rebased branch against its actual base, x86_64 (AMD EPYC 7B13), JDK 26,
`-f 2 -wi 2 -i 5`:

| trial | `main` `7c389986` (ns) | this PR (ns) | |
| :--- | ---: | ---: | ---: |
| `fixedAnchorLogCaptures.match.256` | 224.0 ± 3.5 | 224.5 ± 3.4 | 0.998× |
| `fixedAnchorLogCaptures.match.1000` | 377.7 ± 5.2 | 377.5 ± 4.3 | 1.001× |
| `multiInfixLogCaptures.match.1000` | 6,273.9 ± 88.5 | 6,289.1 ± 128.8 | 0.998× |
| `unanchoredLogCaptures.match.1000` | 1,347.7 ± 26.6 | 1,342.0 ± 24.6 | 1.004× |
| `stackTraceExtract` | 6,576.9 ± 70.4 | 6,609.8 ± 48.9 | 0.995× |
| `apiRouteMatch` *(control, no captures)* | 582.2 ± 10.3 | 579.7 ± 2.8 | 1.004× |

Flat everywhere, inside the ±1% noise floor on this box.

The likely mechanism, not directly instrumented: with the selective start literal #876 now picks,
the accelerator lands on or beside the true match, so BitState succeeds on its first candidate and
never takes the walk this PR bounds. The walk is still reachable; these workloads just stopped
reaching it.

So what is left is the part that never depended on a benchmark:

* The unbounded `searchWithBitStateOrNfa(…, scanner.length(), …)` is `O(n × |prog|)` scalar work
  reachable from an ordinary `find()`. #876 removed it from the workloads we measure. It did not
  remove it from the code, and nothing pins it there.
* The bound costs nothing measurable, and the fall-through is semantics-preserving.
* `SearchScalingRegressionTest.observedCaptureDemandDoesNotMakeFindWalkTheInputWithTheCaptureEngine`
  pins it with `WorkCounter` totals rather than wall time, so it keeps holding regardless of what
  literal selection does next. That test is the durable part of this PR.

If you would rather not carry a change with no benchmark behind it, I am fine closing it — but I
would want the regression test to land either way, since it is what would have caught the
interaction between literal selection and engine routing in the first place.

## Why not one of the simpler changes

* **Scope the preference to callers that actually demand captures.** The latch
  (`Pattern.innerCapturesObserved`) is sticky and `Pattern`-global, which is a real wart — but it is
  not this cost. The behaviour reproduces on a `Pattern` whose caller reads `group(1)` on every
  match, so narrowing who the preference applies to moves nothing.
* **Delete the preference.** It was introduced deliberately in #587 with measured wins:
  `logLineParse` 1.85×, `stackTraceExtract` 1.51×, `csvFieldScan` neutral.
* **Restrict it to anchored operations.** `logLineParse` is `matches()` and would survive, but
  `stackTraceExtract` is a `find()` loop — this gives up half of #587's win.

## Tests

The new `SearchScalingRegressionTest` case compares `WorkCounter` totals for the same pattern and
input with and without the latch set, and requires the latched path to stay within 2× of the
deferred one. It was checked against the unfixed sources first, where it fails (`deferredWork=478`,
`eagerWork=1108`); the input was then widened for margin. The class is `@Tag("work-counter")`, so it
only runs under `-Pwork-counters`.

## Not addressed here

* Patterns with **no start accelerator** keep the unbounded attempt. No benchmark exercises the bad
  case there, and narrowing it would surrender #587's win on that population with nothing measured
  either way.
* BitState still has no literal acceleration *between* candidate starts. Driving its candidates from
  the accelerator the way the DFA's accelerated states do is the principled fix, but it is not a
  small one: the unanchored loop is linear only because of the cross-start `dead` pruning added in
  #830, which #830 deliberately disables in anchored mode, so a naive "anchored BitState per
  accelerator candidate" loop would be quadratic.
* **`CAPTURE_DEMAND_TEXT_LIMIT = 512` is unchanged, and I measured raising it rather than leaving it
  as a suggestion.** At 4096, with bytecode parity verified (`sipush 512` → `sipush 4096`, a single
  operand, the class otherwise identical): `fixedAnchorLogCaptures.match.1000` improves 1.107×, but
  `multiInfixLogCaptures.match.1000` regresses **7.02×** (6,289 → 44,136 ns) and
  `unanchoredLogCaptures.match.1000` 1.26×. All three are 1,000-byte capture workloads, so text size
  cannot be the right gate — the split tracks whether the pattern has unbounded repetition before
  the first capture group. Worth a separate change; 512 should stay for now.

## Relationship to #830, #834 and #835

* **#830** (merged) made this same walk linear with its cross-start `dead` bitmap, and its problem
  statement names this exact call site. It fixed the asymptotics; this avoids taking the walk.
* **#834** (merged) targeted the same trial by routing the workload to the gap executor instead.
  #865's deletion of the executor gave that back; #876 has since recovered it without an executor.
* **#835** is step 3 of that ladder and is now unmergeable: `MultiAnchorExecutor.java` accounts for
  +264/−54 of its diff and no longer exists.


eaftan (commented):
This bounded speculation seems like a reasonable robustness improvement. As the PR description notes, #876 eliminated the measurable improvement in the existing benchmark workloads by choosing a more selective start literal, but the expensive path remains reachable when that literal occurs at rejected candidates.

During review, I fixed a diagnostics bug where abandoned capture speculation could discard the boundary strategy from an earlier successful replacement match. I also corrected the work-counter regression test so its decoys contain the selected literal and actually exercise rejected accelerated candidates. The corrected test fails on main with `deferredWork=456, eagerWork=1385`, and passes here with `deferredWork=456, eagerWork=467`.

LGTM

eaftan (approved):

eaftan:
PR Shepherd is now monitoring this PR at `d417ede436827f399feff644777bfdfa3e22c11d`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `d417ede436827f399feff644777bfdfa3e22c11d`.

PR #883, opened 2026-09-15
Scan start literals on their rarest byte

cushon:
`String.indexOf` is intrinsified, but its two kernels are not the same speed. On x86_64 a
multi-character needle scans at 0.119 ns/byte where `indexOf(char)` reaches 0.020; on
aarch64 the gap is wider still, 0.225 against 0.028. Every exact-case literal scan in the
String domain — the Tier 1 start accelerators for `StartPlan.Literal` and
`StartPlan.FixedOffset`, and the Tier 0 reject prefilter for `RejectPlan.RequiredLiteral` —
issued a bare `text.indexOf(literal)` and so took the slow one.

Each of those three sites already had a UTF-8 twin going through a shift-table scan, and the
case-insensitive String path already anchored on a rare byte via `Matcher.indexOfIgnoreCase`.
The exact-case path, much the more common one, was the one left out.

This scans for one rare character of the literal and verifies the hits. New class
`StringLiteralSearch` is the single entry point for all three call sites.

### Bounding a guess

The anchor is chosen at compile time from a corpus-wide frequency table, so it is a guess
about *this* haystack, and static rarity provably cannot separate the winners from the
losers: `id:` and `error:` both anchor on `:` at rarity 29, and one wins several times over
where the other loses. `indexOf(String)` is not a slow scan that verification improves on —
it is a fused filter and verify that emits no false positives — so a dense anchor loses to it
badly.

Three defences bound the cost of a wrong guess, all within a single call:

- Windows shorter than `strikeBudget * minDensityStride` are not anchored at all. That is the
  shortest window in which the density test below can reach a sparse verdict, and it
  independently matches the figure the cost model gives.
- A false candidate is rejected on the literal's last character before `startsWith` is
  called. The first character is left to `startsWith`, which compares it first anyway.
- Candidates arriving closer together than `minDensityStride` are counted as strikes; once
  the budget is spent the search reverts to `indexOf(String)` for the rest of the scan.

The decision deliberately does not outlive a single call. Carrying it in an object shared
across the calls of one search was measured at 4.2× the cost of the entire search on a
candidate-dense input, while never once changing a decision — the strike constants stop being
compile-time constants and the reference threads a field into the DFA's innermost loop.
Backing off *across* calls is already handled a layer up by `Dfa`'s quarantine, which wraps
these calls; what was missing was protection within one.

The adaptive-defeat parameters move to `AcceleratorPolicy`, which gives the previously dead
`strikeBudget` component a meaning and gives `Dfa`'s quarantine a shared source of truth for
constants it had hard-coded. Values are unchanged, so `Dfa`'s behaviour is identical.

### Reach

Over the 46,866-pattern corpus the anchor applies to 88.93% of exact-case `StartPlan.Literal`
sites, 50.12% of `StartPlan.FixedOffset`, and 99.76% of `RejectPlan.RequiredLiteral`.

### What was split out

A fourth commit routed one-character literals to `String.indexOf(char)`, on the grounds that
nothing between `StringStartAccelerator` and the intrinsic looks at the needle's length.
That is true on x86_64, where it is worth **6.37×** on `uuidExtraction` — whose start literal
since #876 is the single character `-`.

It is **not** true on aarch64. Measured directly there, `indexOf(String)` with a
one-character needle and `indexOf(char)` both run at **0.0282 ns/byte** — identical to four
digits. The routing already happens, and the commit can only buy the ~0.5 ns difference in
per-call setup. What it *costs* there is codegen: growing `indexOfDirect` from 8 to 40
bytecodes to host the branch measured **−6.7%** on a workload whose literal is ten characters
long and which never takes the branch at all.

The fix is to make the decision where the literal's length is already known —
`StringStartAccelerator`, at plan time — so there is no per-call branch on either ISA. That
is tracked as a separate change and is not needed for anything here.

### Tests

`StringLiteralSearchTest` is a differential against `String.indexOf` over random inputs at
every start position, plus the work-budget fallback, non-ASCII haystacks and the boundary
indices. Validated by mutation: an off-by-one on the last valid start fails 4 of 8 cases, and
dropping the anchor-offset bias on the scan fails 3.

### Benchmarks

```
./safere-benchmarks/scripts/compare-branch.sh   --baseline origin/main   --current origin/adaptive-accelerator-policy   '(ApplicationBenchmark\.(apiRouteMatch|csvFieldScan|logLineParse|secretRedaction|stackTraceExtract|urlExtraction)|FixedOffsetAccelerationBenchmark\.findSlashNoise|RealWorldRegexBenchmark\.runBenchmark\.(fixedAnchorLog\.(match|noMatch)\.100000|fixedAnchorLogCaptures\.(match|noMatch)\.(1000|256)|guardByteDelimitedRecord\.(match|noMatch)\.100000|isoTimestampLog\.(match|noMatch)\.100000|uuidExtraction\.(match|noMatch)\.(100000|10000)))@safere-string'
```
### Benchmarks

21 trials from the cross-engine harness, `@safere-string`, `-f 5 -wi 5 -i 10 -r 1 -w 1`:

```
java -Dsafere.benchmark.corpus=<corpus> -jar safere-benchmarks/target/benchmarks.jar \
  CrossEngineBenchmark.run -f 5 -wi 5 -i 10 -r 1 -w 1 -p crossEngineTrial="$TRIALS"
```

Speedup is baseline/current, so >1 is faster. Two machines: x86_64 (AMD EPYC 7B13,
Linux) and aarch64 (Apple M-series, macOS). Run-to-run noise measured from the five
`ApplicationBenchmark` controls, none of which touch this code: ±1% on x86_64, ±3%
on aarch64.

| trial | x86 base (ns) | x86 | arm base (ns) | arm |
| --- | ---: | ---: | ---: | ---: |
| `isoTimestampLog.noMatch.100000` | 9,843 | **6.21** | 24,424 | **7.66** |
| `isoTimestampLog.match.100000` | 10,005 | **5.68** | 25,218 | **7.07** |
| `guardByteDelimitedRecord.match.100000` | 469,289 | 0.99 | 295,646 | 1.10 |
| `guardByteDelimitedRecord.noMatch.100000` | 9,861 | 0.99 | 24,366 | 1.01 |
| `fixedAnchorLog.match.100000` | 13,639 | 0.99 | 31,264 | 0.98 |
| `fixedAnchorLog.noMatch.100000` | 13,457 | 0.99 | 30,770 | 1.00 |
| `uuidExtraction.match.100000` | 10,073 | 1.00 | 3,384 | 0.97 |
| `uuidExtraction.noMatch.100000` | 9,828 | 1.00 | 3,165 | 0.96 |
| `uuidExtraction.match.10000` | 1,216 | 1.00 | 468 | 1.00 |
| `uuidExtraction.noMatch.10000` | 1,016 | 1.00 | 339 | 0.95 |
| `fixedAnchorLogCaptures.match.1000` | 383 | 1.01 | 437 | 0.99 |
| `fixedAnchorLogCaptures.noMatch.1000` | 176 | 1.00 | 328 | 0.98 |
| `fixedAnchorLogCaptures.match.256` | 225 | 1.00 | 246 | 0.98 |
| `fixedAnchorLogCaptures.noMatch.256` | 60 | 0.99 | 85 | 1.06 |
| `findSlashNoise` | 16,101 | 0.98 | 10,892 | 0.97 |
| `apiRouteMatch` | 587 | 1.01 | 357 | 1.03 |
| `urlExtraction` | 1,166 | 1.02 | 723 | 0.97 |
| `csvFieldScan` | 3,722 | 1.01 | 1,838 | 1.01 |
| `logLineParse` | 2,637 | 1.00 | 1,843 | 0.97 |
| `stackTraceExtract` | 6,605 | 1.00 | 2,577 | 1.00 |
| `secretRedaction` | 1,593 | 1.00 | 752 | 1.02 |

Geomean 1.18 on x86_64 and 1.21 on aarch64. Excluding the two `isoTimestampLog`
trials the geomean is 1.00 on both, with everything inside the measured noise band:
0.98–1.02 on x86_64 and 0.95–1.10 on aarch64.

`isoTimestampLog` is the pattern this change targets. Its start literal is
`Z error`, and the scan now runs on `Z` (rarity 127) instead of the whole
seven-character literal, which the corpus never contains.

`findSlashNoise` is the only trial that is slightly down on both machines, at 0.98
and 0.97. Its literal is a single `/`, so it is its own rarest byte and the scan is
unchanged; the delta is the cost of the extra dispatch. On aarch64 this trial is
bimodal across JVM sessions — repeated runs of the same build measured 9,613 and
11,190 ns with ±1% within-run error — so treat its aarch64 number as indicative only.


eaftan (approved):
Thanks for splitting the one-character routing back out. I found one linear-time issue in the rare-anchor loop: the adaptive defeat counted only candidate spacing, while each candidate could perform a full `startsWith`. A periodic input can place candidates exactly at the non-strike spacing and make each comparison fail near the end of an arbitrarily long literal, so the submitted loop's work grows quadratically.

I've pushed a fix and a work-scaling regression. The fix caps rare-anchor eligibility at twice the density stride, so candidate verification has a constant bound; longer literals use the existing direct String search. This keeps the hot runtime loop unchanged, which matters because a first repair using a runtime work budget made the wrapper large enough to hurt JIT inlining.

The final exact comparison used the PR's full 21-trial filter against current `main`. Speedup is `main/final elapsed time`, so values above 1.0 are faster.

| Benchmark | main ns/op | final ns/op | Speedup | Result |
|---|---:|---:|---:|---|
| `isoTimestampLog.match.100000` | 7,953 | 2,230 | 3.57× | Central gain reproduced |
| `isoTimestampLog.noMatch.100000` | 8,701 | 2,018 | 4.31× | Central gain reproduced |
| `urlExtraction` | 828 | 771 | 1.07× | Short-input control is healthy |
| `guardByteDelimitedRecord.match.100000` | 303,666 | 286,008 | 1.06× | Directionally faster; baseline uncertainty is wide |
| `uuidExtraction.noMatch.100000` | 8,418 | 7,672 | 1.10× | Faster |
| `findSlashNoise` | 12,190 | 12,565 | 0.97× | Slight loss with overlapping uncertainty |

The remaining controls were broadly neutral. The targeted win is clear, and the added verification work is now linearly bounded. LGTM

eaftan:
PR Shepherd is now monitoring this PR at `541595c284e9444b86fefc292bb564f26aaa9fca`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `541595c284e9444b86fefc292bb564f26aaa9fca`.

PR #885, opened 2026-09-16
Fix case-insensitive byte prefix scan bounds

### LINKED ISSUE #881: fuzz: ArrayIndexOutOfBoundsException in ByteSwarScan.indexOfPairIgnoreCase when anchor pair matches near buffer end
This was found by Utf8InputFuzzer.

---

### Summary
In `ByteSwarScan.indexOfPairIgnoreCase`, an `ArrayIndexOutOfBoundsException` is thrown when an anchor pair candidate matches near `wordEnd` and `prefixLen > maxAnchorOffset + Long.BYTES`.

### Root Cause
In [`ByteSwarScan.java:240-247`](https://github.com/eaftan/safere/blob/main/safere/src/main/java/org/safere/ByteSwarScan.java#L240-L247):
```java
int maxAnchorOffset = Math.max(offset1, offset2);
int wordEnd = length - maxAnchorOffset - Long.BYTES;
...
while (pos <= wordEnd) {
  byte c1 = bytes[offset + pos + offset1];
  if (c1 == low1 || c1 == high1) {
    byte c2 = bytes[offset + pos + offset2];
    if (c2 == low2 || c2 == high2) {
      if (Ascii.regionMatchesIgnoreCase(bytes, offset + pos, prefix, prefixLen)) {
        return pos;
      }
```
`wordEnd` only guarantees that `pos + maxAnchorOffset + Long.BYTES <= length`. If `prefixLen > maxAnchorOffset + Long.BYTES`, `pos + prefixLen` can exceed `length`.

When `pos + prefixLen > length`, if `bytes[offset + pos + offset1]` and `bytes[offset + pos + offset2]` match the pair candidates, line 245 calls `Ascii.regionMatchesIgnoreCase(bytes, offset + pos, prefix, prefixLen)` without checking `pos <= length - prefixLen`. If the subsequent bytes match the prefix up to the buffer end, `regionMatchesIgnoreCase` reads past the logical slice (or past `bytes.length`, throwing `ArrayIndexOutOfBoundsException: Index N out of bounds for length N`).

Note that other candidate verification sites in `indexOfPairIgnoreCase` already include this check:
- Line 224: `pos <= length - prefixLen && Ascii.regionMatchesIgnoreCase(...)`
- Line 282: `if (candidatePos <= length - prefixLen)`
- Line 300: `limitScalar = length - prefixLen; for (; pos <= limitScalar; pos++)`

### Proposed Fix
Add `pos <= length - prefixLen` to the guard on line 245:
```diff
--- a/safere/src/main/java/org/safere/ByteSwarScan.java
+++ b/safere/src/main/java/org/safere/ByteSwarScan.java
@@ -242,7 +242,8 @@ final class ByteSwarScan {
       if (c1 == low1 || c1 == high1) {
         byte c2 = bytes[offset + pos + offset2];
         if (c2 == low2 || c2 == high2) {
-          if (Ascii.regionMatchesIgnoreCase(bytes, offset + pos, prefix, prefixLen)) {
+          if (pos <= length - prefixLen
+              && Ascii.regionMatchesIgnoreCase(bytes, offset + pos, prefix, prefixLen)) {
             return pos;
           }
           verificationWork += prefixLen;
```
(Alternatively, cap `wordEnd = Math.min(length - maxAnchorOffset - Long.BYTES, length - prefixLen);`).

---

```java
package org.safere;

import static com.google.common.truth.Truth.assertThat;

import java.nio.charset.StandardCharsets;
import org.junit.jupiter.api.Test;

class ByteSwarScanTest {

  @Test
  void indexOfPairIgnoreCase_nearBufferEnd_doesNotOverrun() {
    String prefix = "abcdefghij"; // prefixLen = 10
    byte[] bytes = new byte[20];

    // offset1 = 0 ('a'), offset2 = 1 ('b').
    // maxAnchorOffset = 1, Long.BYTES = 8 => wordEnd = 20 - 1 - 8 = 11.
    // Place a 9-byte partial match starting at pos = 11:
    System.arraycopy("abcdefghi".getBytes(StandardCharsets.UTF_8), 0, bytes, 11, 9);

    // At pos = 11, pos <= wordEnd (11 <= 11) is true and the anchor pair matches at 11 and 12.
    // However, pos + prefixLen = 11 + 10 = 21 > bytes.length (20).
    // Missing bounds check at ByteSwarScan.java:245 causes an ArrayIndexOutOfBoundsException.
    int idx =
        ByteSwarScan.indexOfPairIgnoreCase(
            bytes,
            /* offset= */ 0,
            /* length= */ bytes.length,
            prefix,
            prefix.length(),
            /* offset1= */ 0,
            (byte) 'a',
            (byte) 'A',
            /* offset2= */ 1,
            (byte) 'b',
            (byte) 'B',
            /* start= */ 11);

    assertThat(idx).isEqualTo(-1);
  }
}
```

```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: Index 20 out of bounds for length 20
	at org.safere.Ascii.regionMatchesIgnoreCase(Ascii.java:236)
	at org.safere.ByteSwarScan.indexOfPairIgnoreCase(ByteSwarScan.java:245)
```

eaftan:
Case-insensitive byte pair scanning could verify a prefix beyond the logical UTF-8 slice when its anchors were near the start of a longer prefix. This could throw at the physical array end or return a false match using bytes outside the slice.

Cap the SWAR scan loop so both anchor word loads and full-prefix verification fit within the slice. This preserves candidate coverage and the existing linear-work limit. Shared byte scan bounds helpers distinguish full-match limits, anchor-load limits, and the combined limit; the refactor preserves the other loop bounds and per-candidate checks.

The regression matrix runs against both SWAR and vector implementations, covering physical and logical ends, slice offsets, prefix lengths, anchor positions, and search starts. Longer prefixes and positions exercise vector loads as well as scalar tails. A UTF-8 fuzz generator axis covers truncated case-insensitive prefixes. The internal regression class carries both its disabling annotation and the structural crosscheck exclusion required for compilation.

Fixes #881.

## Validation

- Confirmed the original regression tests fail before the fix with both out-of-bounds exceptions and false matches.
- Passed the full `mvn -pl safere test -q` suite and `mvn -pl safere-fuzz -am -Dtest=Utf8InputFuzzer -Dsurefire.failIfNoSpecifiedTests=false test -q` in regression mode before the review-follow-up refactor.
- On the final changes, passed `mvn verify -pl safere,safere-crosscheck -am -Pcrosscheck-public-api-tests -Dtest=ByteScanTest,Utf8InputScannerTest,StringLiteralSearchTest,CrosscheckGenerationPolicyTest -Dsurefire.failIfNoSpecifiedTests=false -q`. This compiles all generated crosscheck sources and runs the selected scanner, literal-search, and policy tests; vector-only cases are skipped in this scalar run.
- On the final changes, passed `mvn test -pl safere -Pvector-tests -Dtest=ByteScanTest,Utf8InputScannerTest,Utf8VectorPairTripleTest,VectorSpeciesSimulationTest,StringLiteralSearchTest,MultiLiteralTest -q`. All shared SWAR/vector bounds cases ran successfully.
- Passed formatting checks through scalar `verify`, and `git diff --check`.
- Running the vector command through `verify` passed its selected tests but failed at the subsequent JAR packaging step; the existing profile attachment conflict is tracked separately in #891.
- The full SafeRE suite and fuzz regression were not rerun after the helper refactor. The full generated public API crosscheck suite, sustained fuzzing, and benchmarks were not run locally.


safere-crosscheck/pom.xml
```diff
@@ -106,6 +106,7 @@
                         <exclude name="AsciiTest.java"/>
                         <exclude name="AstAnalysisTest.java"/>
                         <exclude name="BitStateTest.java"/>
+                        <exclude name="ByteSwarScanTest.java"/>
```
cushon: Idly non-blocking thought: would it be possible for the crosscheck generator to check the annotation to avoid having to use both `@DisabledForCrosscheck` and the `pom.xml` list?

safere/src/test/java/org/safere/ByteSwarScanTest.java
```diff
@@ -0,0 +1,63 @@
...
+            System.arraycopy(
+                upper,
+                0,
+                bytes,
+                offset + position,
+                Math.min(prefixLength, bytes.length - offset - position));
+            for (int start : new int[] {0, position, position + 1}) {
+              int expected = available >= prefixLength && start <= position ? position : -1;
+              assertThat(
+                      ByteSwarScan.indexOfPairIgnoreCase(
```
cushon: It might be nice to parameterize this and also run it over `ByteVectorScan` to make sure they're both correct and stay that way. (As far as I can tell `ByteVectorScan` is already correct and this bug was only in `ByteSwarScan`, but parameterizing the test could prevent regressions.)

safere/src/main/java/org/safere/ByteSwarScan.java
```diff
@@ -201,7 +201,8 @@ static int indexOfPairIgnoreCase(
     long workLimit = WorkLimit.forRemaining(length - pos);
 
     int maxAnchorOffset = Math.max(offset1, offset2);
-    int wordEnd = length - maxAnchorOffset - Long.BYTES;
+    // Both anchor word loads and the immediate full-prefix check must fit in the slice.
+    int wordEnd = Math.min(length - maxAnchorOffset - Long.BYTES, length - prefixLen);
```
cushon: I think maybe there's an opportunity to add some helpers here and in `ByteVectorScan` to make this sort of mistake harder, basically for all of the arithmetic doing `length - prefixLen` in both files.

WDYT about adding something like

```java
/** Last start at which a match of {@code matchLen} still fits in the slice. */
static int lastMatchStart(int length, int matchLen) {
  return length - matchLen;
}

/** Last start at which a {@code loadBytes}-wide load at {@code +anchorOffset} fits. */
static int lastLoadStart(int length, int anchorOffset, int loadBytes) {
  return length - anchorOffset - loadBytes;
}

/** Last start at which both the widest anchor load and the full match fit. */
static int lastWideScanStart(int length, int matchLen, int maxAnchorOffset, int loadBytes) {
  return Math.min(lastLoadStart(length, maxAnchorOffset, loadBytes),
                  lastMatchStart(length, matchLen));
}
```

cushon (commented):
Thanks! LGTM overall, I left a few non-blocking comments.

safere/src/test/java/org/safere/ByteSwarScanTest.java
eaftan: Done

safere/src/main/java/org/safere/ByteSwarScan.java
eaftan: Done

safere-crosscheck/pom.xml
eaftan: Filed issue #892 for this

eaftan:
PR Shepherd is now monitoring this PR at `d3f12c7aea4d1fe75b0e1d4b62da5aafd3630da1`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `d3f12c7aea4d1fe75b0e1d4b62da5aafd3630da1`.

PR #886, opened 2026-09-16
Fix stale DFA arrays after accelerated start-state allocation

### LINKED ISSUE #882: fuzz: `NullPointerException: Cannot read field "next" because "s" is null` in `Dfa.doSearchInternal` after accelerator fast-forward
### Summary
A `NullPointerException` occurs in `Dfa.doSearchInternal` during DFA search when an accelerator fast-forwards to a candidate position where `startState()` allocates a new state that triggers a capacity expansion of the DFA's flat arrays (`transitions`, `offsetToState`, `isAcceleratedStateOffset`).

```
java.lang.NullPointerException: Cannot read field "next" because "s" is null
	at org.safere.Dfa.doSearchInternal(Dfa.java:1691)
	at org.safere.Dfa.doSearch(Dfa.java:1415)
	at org.safere.Matcher.searchForwardDfa(Matcher.java:259)
	at org.safere.Matcher.doFindCore(Matcher.java:1923)
	...
```

### Root Cause
In `Dfa.doSearchInternal`, local references to `transitions`, `offsetToState`, and `isAcceleratedStateOffset` are cached before the search loop (lines 1532–1534).

When an accelerator fast-forwards past non-matching text (lines 1544–1567):
```java
pos = nextPos;
if (pos >= textLen) {
  break;
}
s = startState(text, pos, anchored, false);
if (s == null) {
  return null;
}
```
If `startState` builds a new state not previously in the cache, `addStateToFlatArrays` may reallocate `this.transitions`, `this.offsetToState`, and `this.isAcceleratedStateOffset` to double their previous capacity and store the new state `s` in the new array.

However, `doSearchInternal` does not refresh its local `offsetToState` and `transitions` references after `startState()`. When the search continues with `sId = s.id * numClasses`, the subsequent lookup:
```java
s = offsetToState[sId];
```
reads from the old, unexpanded array where `offsetToState[sId]` is `null`. At line 1691:
```java
State ns = s.next[cls];
```
this dereferences `null`, throwing `NullPointerException`.

### Fix
Refresh the local array variables after calling `startState` in `doSearchInternal`:
```diff
--- a/safere/src/main/java/org/safere/Dfa.java
+++ b/safere/src/main/java/org/safere/Dfa.java
@@ -1568,6 +1568,9 @@ final class Dfa {
           if (s == null) {
             return null;
           }
+          transitions = this.transitions;
+          offsetToState = this.offsetToState;
+          isAcceleratedStateOffset = this.isAcceleratedStateOffset;
           if (s.isMatch()) {
             if (isRequiredEndMatch(pos, needEndMatch, textLen, trailingTermStart)) {
               matched = true;
```

---

#### Minimal Reproducer

```java
package org.safere;

import static com.google.common.truth.Truth.assertThat;

import java.nio.charset.StandardCharsets;
import java.util.HexFormat;
import org.junit.jupiter.api.Test;

class DfaStateArrayStalenessTest {

  @Test
  void fastForwardStartStateAllocation_doesNotStaleOffsetToState() {
    // Hex representations of the fuzzer regex and input triggering fastForward
    // startState reallocation.
    String regexHex =
        "7862777c00000000000000006eebba80e78182ef8095d48437f28d9393e3acbddfaf0dd294efbfbf022c271854df802e2f696173d494d4bf80dfa5dfbfdfb37a24e3a081d99ede9e";
    String inputHex =
        "9e00000000000000002a0a00242a015e0a242a00a6188c5d7224a2247c9f45737a24";

    String regex = new String(HexFormat.of().parseHex(regexHex), StandardCharsets.UTF_8);
    String input = new String(HexFormat.of().parseHex(inputHex), StandardCharsets.UTF_8);

    Pattern pattern =
        Pattern.compile(regex, Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE);
    Matcher matcher = pattern.matcher(input);

    // Before the fix, throws NullPointerException: Cannot read field "next" because "s" is null
    // at org.safere.Dfa.doSearchInternal(Dfa.java:1691)
    assertThat(matcher.find()).isFalse();
  }
}
```

eaftan:
Accelerated DFA search could throw `NullPointerException` when restarting at a candidate created a new start state and resized the flat arrays. The ASCII search loop retained references to the old arrays. Refresh all three cached references after start-state creation, preserving the same invariant already maintained after transition-state creation.

Add regression coverage for the original fuzz input and generated variants across state/alphabet sizes, matching and nonmatching inputs, and repeated searches. Extend `MatchFuzzer` to generate nearby restart/context-change cases. Both new regression tests reproduced the crash before the fix and pass afterward.

Fixes #882.

Validation passed:
- `mvn -pl safere -Dtest=DfaArrayGrowthTest,DfaTest test -q`
- `mvn verify -pl safere,safere-crosscheck -am -Pcrosscheck-public-api-tests -q` (full SafeRE suite and generated JDK public API crosscheck)
- `mvn -pl safere-fuzz -am -Dtest=MatchFuzzer -Dsurefire.failIfNoSpecifiedTests=false test -q` (regression mode)
- `git diff --check`
- Read-only review of the complete branch/worktree: no findings.

Validation used JDK 26.0.2. Continuous fuzzing, benchmarks, and other JDK versions were not run.


cushon (approved):
LGTM

PR #887, opened 2026-09-16
Fix deferred capture coordinates for opaque regions

### LINKED ISSUE #880: NFA full-region match loses capture of an optional lazy group
At commit `0eb00f72`, `(a+?)?` can successfully match a region containing `a` through the NFA but report group 1 as null. The default BitState path reports `a`.

Package-local repro:
```java
Pattern p = Pattern.compile("(a+?)?", 0,
    EnginePathOptions.builder().bitState(false).build());
Matcher m = p.matcher("\uD801\uDC00a").region(2, 3).useAnchoringBounds(false);
System.out.println(m.matches()); // true
System.out.println(m.group(1));  // null; expected "a"
```

With `bitState(true)`, the same operation reports `a`. The region contains only the literal `a`, so the supplementary prefix is outside the consumed region. An optional outer group cannot be absent while the inner `a+?` supplies the entire successful nonempty match.

Found before production edits while building region engine-equivalence coverage for PR #869's performance follow-up. The new matrix compares matches/lookingAt/find traces, captures, region offsets, and bounds settings. This capture issue is separate from the split-end engine-selection cost. A standalone package-local Java probe reproduced it with fresh matchers.

Investigate full-match capture handling in the NFA for optional lazy consuming groups on region paths, including pooled/reused engines and related quantifier combinations. Preserve the linear-time guarantee.

eaftan:
An opaque region can match successfully through the DFA and then lose its inner captures when extraction falls back to the NFA. For example, `(a+?)?` on `"xa"` with region `[1, 2)` reports a successful full match but a null group 1 when BitState is disabled. The same failure occurs with default engine selection once the region exceeds BitState's input limit.

Normalize the region bounds to the temporary input view during deferred capture extraction, then restore the original bounds before translating captures back to input offsets. This keeps the NFA assertion context and scanner in the same coordinate system, with constant additional work and no additional matching passes.

Add JDK differential coverage for related quantifiers, nested and named groups, assertions, bounds settings, matches/lookingAt/find sequences, and pooled/reused engines. Also cover UTF-8 byte offsets and default engine fallback. Add a grammar-based deferred-capture fuzzer and include its package in the existing fuzz runner's discovery. Exclude the engine-control test from generated crosscheck compilation, consistent with existing internal tests; it directly compares against the JDK.

Fixes #880.

Validation on OpenJDK 26.0.2:

- Confirmed the new differential matrix fails before the production fix.
- Passed `mvn -pl safere -Dtest=DeferredRegionCaptureTest,MatcherDeferredCaptureStateTest test -q`.
- Passed `mvn -pl safere test -q`.
- Passed `mvn -pl safere-fuzz -am -Dtest=DeferredRegionCaptureFuzzer,DeferredRegionCaptureTest -Dsurefire.failIfNoSpecifiedTests=false test -q`, including the added default-engine fallback regression.
- Passed `safere-fuzz/scripts/run-fuzz-test.sh --max-duration 30s --keep-going 1 DeferredRegionCaptureFuzzer` with no findings.
- Passed `mvn verify -pl safere-crosscheck -Pcrosscheck-public-api-tests -q` against this checkout freshly installed with `mvn -pl safere -am install -DskipTests -q`.
- Passed formatting, shell syntax, and whitespace checks.

Not run: benchmarks, extended fuzzing, or validation on other JDK versions.


eaftan:
PR Shepherd is now monitoring this PR at `e53e0206b76e94481ecd765e588327c9acc0b0b3`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `e53e0206b76e94481ecd765e588327c9acc0b0b3`.

PR #888, opened 2026-09-16
Name the byte-pair and byte-triple scan kernels for what they compare

cushon:
This is a pure refactoring to rename some methods. I had started looking at this as part of the gap exector when there were more kernels, where consistent naming would have made it easier to have test coverage asserting that the full cross product of accelerators existed across string/byte[], start and end-acceleration, at one point `lastIndexOfAsciiPair` and `lastIndexOfBytePair` both existed. The removal of the gap executor in #865 has made that cross product smaller, but I think these names are more accurate.

---

`VectorScanProvider.indexOfAsciiPair` and `indexOfAsciiTriple` advertise an ASCII
precondition they do not have. They compare for equality:

```java
// ByteVectorScan:79
values.compare(EQ, v0).or(values.compare(EQ, v1))
```

Equality is exact for all 256 byte values, and the SWAR fallback is byte-generic
too (`(first & 0xFFL) * BYTE_ONES`). The `Ascii` in the name describes today's
only caller, which happens to gate on `c < 128`, not the kernel.

That the two are one operation is visible at the call site — `Utf8InputScanner`
dispatches to the provider and falls back to `ByteSwarScan` with the same
arguments, under two different names:

```java
int hit = pairProvider.indexOfAsciiPair(bytes, offset, scanLen, b0, b1, start);  // :194
...
return ByteSwarScan.indexOfBytePair(bytes, offset, scanLen, b0, b1, start);      // :201
```

`ByteSwarScan` had it right. This renames the provider side to match, and writes
the rule down on the interface:

> A kernel is named `Byte` when it compares for equality, which is exact for all
> 256 byte values, and `Ascii` when it compares for *order*, which is only
> correct below `0x80` because `byte` is signed.

`indexOfAsciiClass` keeps its name: it takes `int[] ranges` and compares ordering,
so signed-byte comparison genuinely breaks above `0x80`. The scanner-layer
`indexOfAsciiPair(int c1, int c2, ...)` also keeps its name — it takes code points
and its callers do gate on ASCII.

Beyond tidiness: a future caller wanting a non-ASCII byte pair — a UTF-8 lead-byte
pair, say — reads the current name as a prohibition, and would either gate on
ASCII for no reason or add a redundant kernel.

No behaviour change. `javap -c -p` on `Utf8InputScanner`, `ByteVectorScan` and
`IncubatorVectorScanProvider` is identical before and after, modulo constant-pool
indices: same instructions, same bytecode offsets, same code-array sizes. That
matters here because `StringLiteralSearch.indexOf` and its neighbours have crossed
a JIT inlining threshold six times in this repo; a rename should be free, and this
confirms it is.

Tests: 23,281 default, 23,281 with `-Pvector-tests`, 104 with `-Pwork-counters`.
No failures, no errors.


eaftan (approved):
LGTM, thanks for the cleanup!

PR #893, opened 2026-09-17
Prevent anchored DFA cache states from skipping split delimiters

### LINKED ISSUE #889: Unanchored DFA search falsely accelerates on cached anchored start states
This was found by fuzzing.

`Pattern.split()` can miss intermediate matches and produce fewer tokens than JDK `java.util.regex.Pattern.split()` when unanchored DFA search visits a cached state that was previously marked as an anchored start state.

### Root Cause

`Dfa.startState(InputScanner, int, boolean, boolean)` unconditionally marks the returned state with:
```java
s.isStartState = true;
isAcceleratedStateOffset[s.id * numClasses] = true;
```
even when `anchored == true` or `reverseContext == true`.

During `Matcher.findSplitPositions()` / `findNextMatchPacked()`, matches are verified using an anchored forward DFA search (`dfa(false).doSearch(scanner, matchStart, true, false)`). This caches the anchored start state of the pattern with `isStartState = true`.

For patterns containing repeated character classes followed by a literal that belongs to the class (e.g., `[0-9]*0` or `[0-y]*\0`), consuming a character that continues the repetition transitions the DFA into an instruction set identical to the anchored start state. During a subsequent unanchored search attempt, the DFA loop encounters `s.isStartState == true`, incorrectly infers that the search has returned to the idle start state, and invokes `fastForward(...)`. This start acceleration skips ahead to a later candidate position and discards the valid in-progress match.

### Minimal Reproducer

```java

public class Repro {
        String regex = "[0-9]*0";
        String input = "0xxx011a----------------0";
        String[] js = java.util.regex.Pattern.compile(regex).split(input);
        String[] ss = org.safere.Pattern.compile(regex).split(input);
        System.out.println("JDK:    " + java.util.Arrays.toString(js));
        System.out.println("SafeRE: " + java.util.Arrays.toString(ss));
        assert java.util.Arrays.equals(js, ss) : "Divergence between JDK and SafeRE";
    }
}
```

```
JDK:    [, xxx, 11a----------------]
SafeRE: [, xxx011a----------------]
```

### Proposed Remediation

In `Dfa.startState(...)`, only set `s.isStartState = true` and `isAcceleratedStateOffset[...] = true` when `!anchored && !reverseContext`.

eaftan:
`Pattern.split()` could skip intermediate delimiters after anchored verification marked a shared DFA state as eligible for unanchored start acceleration. For example, splitting `0xxx011a----------------0` with `[0-9]*0` omitted the middle delimiter.

Only forward, unanchored start states now receive the start-acceleration marker. Anchored and reverse-context starts remain cached without granting that permission, preserving in-progress matches and the linear-time guarantee.

Adds systematic coverage for split limits, greedy and reluctant repetition, captures, Unicode, and cache reuse across APIs; direct String/UTF-8 DFA regressions; and a dedicated split fuzz generator. The new public API tests reproduced 73 failures before the fix and pass afterward. A read-only review found no actionable findings.

Validation passed:
- `mvn -pl safere -Dtest=SplitDfaCachingTest,DfaTest test -q`
- `mvn verify -pl safere,safere-crosscheck -am -Pcrosscheck-public-api-tests -q` (full library suite and generated public API crosscheck)
- `mvn -pl safere-fuzz -am -Dtest=SplitFuzzer -Dsurefire.failIfNoSpecifiedTests=false test -q` (regression mode)
- `git diff --check`

Validation used JDK 26.0.2. Other JDK versions, sustained fuzzing, and benchmarks were not run.

Fixes #889


cushon (approved):
LGTM

Issue #890, opened 2026-09-16
find() ignores a known-selective required literal and leads with an unselective start literal (guardByteDelimitedRecord, 2.4-3.6x since #865)


cushon:
`RealWorldRegexBenchmark.runBenchmark.guardByteDelimitedRecord.match.100000` is 2.4–3.6× slower than it was before #865 retired the multi-anchor gap executor. It is the one family that regressed with that change and has not recovered since.

This issue records the measurement, rules out the obvious explanations, and narrows what actually needs to be done.

### Measurement

Three arms, 21 trials, three interleaved cycles, short protocol (`-f 2 -wi 2 -i 5`), per-arm corpora with the shared inputs verified byte-identical by SHA-256. Ratios are `A / arm`, so **< 1 means slower**.

| arm | commit | |
| :--- | :--- | :--- |
| `A_pre` | `024e03c9` | before the executor retirement |
| `B_865` | `7d773d53` | #865, executor retired |
| `C_main` | `35199cfb` | plus #875 #876 #877 #878 #883 #887 |

| | `A→B` | `B→C` | `A→C` |
| :--- | ---: | ---: | ---: |
| x86_64 (EPYC 7B13) | 0.293 | 0.951 | **0.279 — 3.58× slower** |
| aarch64 (Apple M) | 0.394 | 1.062 | **0.419 — 2.39× slower** |

Per-cycle `A→C` on x86 is 0.279 / 0.294 / 0.278, spreads of 2.0% / 2.3% / 3.3% against a ±3% control band. Raw x86 medians, ns/op:

```text
A_pre    137,753   140,368   137,579
B_865    470,191   479,526   468,811     the retirement
C_main   494,177   478,245   494,589     the follow-ups
```

Two things worth noting:

*   **The follow-up PRs did not recover any of it.** They moved it 5% the wrong way on x86 and 6% the right way on aarch64 — which is to say, nothing.
*   **`noMatch` is completely unaffected**: 0.984× on x86, 1.020× on aarch64. That asymmetry is the main clue, see below.

Independently reproduced outside JMH, one pattern per JVM, 20k warmup iterations and the best of five 2k-iteration blocks — no shared lambda call site, so no megamorphic dispatch:

```text
A_pre    137,225 ns/op
C_main   463,516 ns/op     3.38x
```

### It is not a planning regression — the plan is unchanged

The compiled plan is identical on both sides:

```text
pattern    record:[^\r\n;]*val=200

[...]

PR #896, opened 2026-09-17
Base DFA accelerator backoff on actual skipped work

eaftan:
Dense start candidates can trigger DFA accelerator backoff even when each scan bypasses useful work. Charge each accelerator call its existing estimated cost and credit only positions actually skipped, so candidate spacing and scalar progress during quarantine cannot make an unprofitable scan appear useful.

The bounded deficit preserves partial losses, grows quarantine for persistently unprofitable scans, and restores the initial window after surplus skips repay the deficit. Both DFA loops use the same accounting; matching transitions and cost constants are unchanged. Add compile-time-disabled work counters and deterministic String/UTF-8 tests for profitable dense starts, negligible skips, partial losses, break-even calls, and recovery after quarantine.

Refs #890. This addresses the backoff component; it does not implement reverse-inner search or establish complete recovery against historical revisions.

### Performance

Before: `eb948032ef3795b778f5fd2d43e7884b5df16fdc` (2026-09-16T23:49:17Z). After: the implementation in this PR. Sequential runs through `run-java-benchmarks.sh` on an Intel Core i7-11700K, WSL2, OpenJDK 26.0.2, with identical input manifests. Standard comparison covered 26 targeted String/UTF-8 rows (2 forks, 2 × 500 ms warmup, 5 × 500 ms measurement); nine important/control rows received `--long` confirmation (2 forks, 3 × 1 s warmup, 5 × 1 s measurement), with candidate/baseline order reversed.

Long confirmation, µs/op ± JMH 99.9% confidence half-width:

| --- | ---: | ---: | ---: |
| guardByteDelimitedRecord.match.100000, String | 325.187 ± 7.756 | 126.576 ± 4.212 | 2.57× faster |
| guardByteDelimitedRecord.match.100000, UTF-8 | 317.605 ± 29.075 | 150.011 ± 2.068 | 2.12× faster |
| multiInfixLog.match.100000, String | 339.647 ± 19.160 | 96.256 ± 3.532 | 3.53× faster |

Separate untimed tracing shows skipped positions rising from 11,219 to 49,979 on the record String workload and quarantines falling from 38 to zero, with JDK-equivalent match bounds. Profiling before implementation identified forward DFA execution as the dominant cost.

Tradeoffs: no standard row had a mean regression above 5%. Long-run String short16 and long32 controls had noisy point estimates 2.0% and 6.6% slower, respectively, with overlapping confidence intervals; UTF-8 long32 was 1.5% slower. Existing cost estimates remain conservative, and these targeted measurements do not establish improvement on every workload or architecture. Raw evidence remains in the local ignored `benchmark-results/issue-890-skip-policy/` directory; this PR does not publish a benchmark collection or change BENCHMARKS.md.

### Validation

Passed on the final implementation:
- `mvn -pl safere test -q`
- `mvn -pl safere -Pwork-counters test -q`
- `mvn -pl safere -Dtest=DfaTest,StartAcceleratorTest,StartAcceleratorDifferentialTest,GapPatternMatchingTest,PrefixAccelerationTest,MatcherTest test -q`
- `mvn -pl safere spotless:apply -q` and `git diff --check`
- Read-only review of the complete change: no findings.

The new tests fail against the original policy; the break-even regression also rejects an intermediate policy that discarded partial losses. New deterministic recovery coverage focuses on literal acceleration; other policies rely on existing coverage. Public API crosscheck, sustained fuzzing, the full benchmark collection, and the external OpenJDK-derived suite were not run.

### CI correction

Add the compile-time crosscheck exclusion for `DfaStartAccelerationWorkTest`, which uses package-private SafeRE internals and already carries `@DisabledForCrosscheck`. That annotation disables execution but cannot prevent compilation of inaccessible types. Both failed CI jobs had this same compilation error.

Reproduced locally with `mvn -pl safere-crosscheck -am -Pcrosscheck-public-api-tests test-compile -DskipTests -q`. After the exclusion, generated public API crosscheck sources compile and `CrosscheckGenerationPolicyTest` passes via `mvn -pl safere-crosscheck -am -Pcrosscheck-public-api-tests test -Dtest=CrosscheckGenerationPolicyTest -Dsurefire.failIfNoSpecifiedTests=false -q`. This correction changes only test-generation configuration; it does not rerun the full public API crosscheck suite or alter production code.


cushon:
I had an agent review this, it had some notes but overall this seems like a good direction.

---

I measured this against its base, and it does what it says: **2.75× on
`guardByteDelimitedRecord.match.100000`**, with everything else flat.

Short protocol (`-f 2 -wi 2 -i 5`), three interleaved cycles, per-arm corpora,
medians in ns/op:

| trial | `eb948032` | `d8dd8232` | ratio |
| :--- | ---: | ---: | ---: |
| **`guardByteDelimitedRecord.match.100000`** | **534,687** | **194,152** | **2.754×** |
| `guardByteDelimitedRecord.noMatch.100000` | 10,004 | 9,957 | 1.005× |
| `guardByteHeader.match.100000` | 3,277 | 3,278 | 1.000× |
| `isoTimestampLog.match.100000` | 1,762 | 1,762 | 1.000× |
| `logLineParse` | 2,657 | 2,624 | 1.013× |
| `urlExtraction` | 1,139 | 1,144 | 0.996× |

Two separate runs an hour apart gave 2.750× and 2.754×, so the headline is
solid. Per-cycle it is 2.67 / 3.62 / 2.65, so **"at least 2.65×"** is the
conservative claim.

I included `guardByteHeader` deliberately: same pattern shape, but its input is
built so the start literal is selective (1 hit against
`guardByteDelimitedRecord`'s 2,941). It measures 1.000×, so this is not trading
one guard-byte shape against the other, which was the thing I most expected to
go wrong.

### How much of #890 this recovers

Adding the pre-executor-retirement commit `024e03c9` as a third interleaved arm:

| | `024e03c9` | `eb948032` | `d8dd8232` |
| :--- | ---: | ---: | ---: |
| `guardByteDelimitedRecord.match.100000` | 137,219 | 534,687 | 194,152 |
| vs `024e03c9` | — | **3.90× slower** | **1.41× slower** |

**That is 86% of the absolute regression gone** (+397,468 ns → +56,932 ns).
#890 should probably stay open for the remaining 1.41×, but it stops being a
headline number, and the reverse-inner work it is filed against is no longer
the only thing that can move it.

One more thing that shows up in the per-cycle data and is worth having on the
record:

```
eb948032   518,347   699,004   534,687     35% spread
d8dd8232   194,152   193,314   201,559      4% spread
```

The old policy is not just slower here, it is **erratic** — which is what you
would expect from a backoff whose trigger depends on candidate spacing that
includes DFA-consumed positions. The new accounting is stable across cycles.
That is a real second benefit and worth a line in the description.

### Codegen — both concerns cleared, no action needed

I checked these rather than asking you to, because this file has a history of
surprising us:

*   **`doSearchInternal` is 83 bytecodes *smaller*** (2,223 → 2,140). The
    deficit accounting is simpler than the strike/stride scheme it replaces.
    It is far past `FreqInlineSize` either way, so nothing changes about
    inlining, but it is not growing.
*   **The compiled `Dfa` contains zero `WorkCounter` references on both arms.**
    `WorkCounterConfig.ENABLED` is a build-time constant, so javac eliminates
    the two new telemetry branches outright. They cost nothing in the shipped
    jar. Worth stating explicitly somewhere, since "telemetry in the innermost
    loop" reads alarming until you know that.

### The one thing I would still like to understand

Repayment got much more aggressive, and it resets the exponential backoff.

For `AcceleratorPolicy.LITERAL` (`minProfitableSkip = 16`, `strikeBudget = 16`,
so `lossLimit = 256`), consider the state right after a quarantine fires, where
`skippedWorkDeficit = lossLimit >>> 1 = 128`:

| | what it takes for one scan to clear the deficit and reset `quarantineWindow` to 2048 |
| :--- | :--- |
| before | a single **stride of 2,048** (`stride >>> 8` must repay 8 strikes) |
| after | a single **skip of 144** (`max(0, 128 + 16 - skipped)` reaches 0) |

So roughly 14× less sparse evidence now un-does a quarantine, *and* it discards
the accumulated backoff window rather than decaying it. The shape I cannot rule
out is *(dense region)(~150-unit gap)(dense region)(~150-unit gap)…*, where
every gap resets the window and backoff never compounds, so the accelerator
keeps paying per-call cost across every dense region for the whole input.

I do not have a workload that does this — it may well be unreachable in
practice. But it is the adversarial-input question, and
`DfaStartAccelerationWorkTest` is already the right place to answer it: a case
on an alternating input asserting that `largestQuarantine()` actually escalates
would settle it in either direction.

If it does reproduce, the smallest fix is probably to decouple the two effects —
let a profitable call repay the deficit, but keep `quarantineWindow` where
backoff left it (or halve it) instead of snapping back to
`initialQuarantineWindow`. That preserves the memory of repeated failure while
still letting a genuinely-good region resume acceleration.

Relatedly: nothing currently exercises quarantine-then-repay at all. Both test
cases assert steady states — `quarantines() == 0` on profitable input, backoff
on dense input. If the repayment branch is worth having, something should show
it recovering.

### Small

`strikeBudget` now carries two unit systems: `Dfa` reads it as a multiplier
(`strikeBudget() * minSkip`, input units) while `StringLiteralSearch` still
reads it as a literal count of strikes. The javadoc is explicit about the
split so this is not a bug, but would a derived `lossLimit()` accessor on the
policy be worth it? It names the quantity, keeps the multiply out of a hot
method, and leaves `strikeBudget` meaning one thing per caller.

cushon:
Some more code feedback from the agent

---

Two follow-ups

**`minProfitableSkip` displaced the one constant in this file that had a derivation.** The old path read `minDensityStride`, and `AcceleratorPolicy` records where 64 came from: *"anchoring on a single character saves roughly 0.1 ns/byte over `String.indexOf(String, int)` and spends one verification per anchor occurrence, so it stops paying at roughly one candidate per 64–90 bytes."*

This PR drops `minDensityStride` from `doSearchInternal` and puts `minProfitableSkip` in its place, where it now does three jobs at once: the profit test (`skipped < minSkip`), the loss unit (`deficit += minSkip - skipped`), and the quarantine trigger (`lossLimit = strikeBudget * minSkip`). `LITERAL` is 16, and the record still carries `// TODO: Conduct systematic empirical micro-benchmarking ... to precisely tune minimum profitable skip thresholds`.

I'm not claiming 16 is wrong — the 2.75× says it isn't wrong here. But three uses of one constant compound, and the number now carrying them is the one the file itself flags as unmeasured. A sweep of 8 / 16 / 24 / 32 for `LITERAL` on the #890 workload would settle it, and would say whether `CHAR_CLASS`'s 24 and `VECTOR_MULTI_LITERAL`'s 64 are in the right place too.

**Coverage is `LITERAL`-only.** All six tests run `record:[^\r\n;]*val=200`. `VECTOR_MULTI_LITERAL` (`minSkip` 64, so `lossLimit` 1024) and `CHAR_CLASS` (24) are where the new arithmetic diverges most from the old, and neither is exercised.

Two smaller notes, neither blocking:

- `accelerationResumePos = nextPos + Math.min(quarantineWindow, textLen - nextPos)` quietly fixes an overflow the old `nextPos + quarantineWindow` had on inputs near `Integer.MAX_VALUE`. Worth a line in the description rather than passing as incidental.
- The quarantine state is still per-search — `skippedWorkDeficit`, `quarantineWindow` and `accelerationResumePos` are locals in `doSearchInternal`, so a `find()` loop forgets the quarantine and re-pays up to `lossLimit` on each call. Not a regression, the old code was the same. But #908 landed cross-search accounting for the DFA cache in this exact file, on the argument that per-search state loses information across a `find()` loop. The counter-argument — a quarantine is cheap to relearn, a state cache is not — is probably right, but worth writing down rather than inheriting.

eaftan:
> I had an agent review this, it had some notes but overall this seems like a good direction.
> 
> I measured this against its base, and it does what it says: **2.75× on `guardByteDelimitedRecord.match.100000`**, with everything else flat.
> 
> Short protocol (`-f 2 -wi 2 -i 5`), three interleaved cycles, per-arm corpora, medians in ns/op:
> 
> trial	`eb948032`	`d8dd8232`	ratio
> **`guardByteDelimitedRecord.match.100000`**	**534,687**	**194,152**	**2.754×**
> `guardByteDelimitedRecord.noMatch.100000`	10,004	9,957	1.005×
> `guardByteHeader.match.100000`	3,277	3,278	1.000×
> `isoTimestampLog.match.100000`	1,762	1,762	1.000×
> `logLineParse`	2,657	2,624	1.013×
> `urlExtraction`	1,139	1,144	0.996×
> Two separate runs an hour apart gave 2.750× and 2.754×, so the headline is solid. Per-cycle it is 2.67 / 3.62 / 2.65, so **"at least 2.65×"** is the conservative claim.
> 
> I included `guardByteHeader` deliberately: same pattern shape, but its input is built so the start literal is selective (1 hit against `guardByteDelimitedRecord`'s 2,941). It measures 1.000×, so this is not trading one guard-byte shape against the other, which was the thing I most expected to go wrong.
> 
> ### How much of #890 this recovers
> Adding the pre-executor-retirement commit `024e03c9` as a third interleaved arm:
> 
> `024e03c9`	`eb948032`	`d8dd8232`
> `guardByteDelimitedRecord.match.100000`	137,219	534,687	194,152
> vs `024e03c9`	—	**3.90× slower**	**1.41× slower**
> **That is 86% of the absolute regression gone** (+397,468 ns → +56,932 ns). #890 should probably stay open for the remaining 1.41×, but it stops being a headline number, and the reverse-inner work it is filed against is no longer the only thing that can move it.
> 
> One more thing that shows up in the per-cycle data and is worth having on the record:
> 
> ```
> eb948032   518,347   699,004   534,687     35% spread
> d8dd8232   194,152   193,314   201,559      4% spread
> ```
> 
> The old policy is not just slower here, it is **erratic** — which is what you would expect from a backoff whose trigger depends on candidate spacing that includes DFA-consumed positions. The new accounting is stable across cycles. That is a real second benefit and worth a line in the description.
> 
> ### Codegen — both concerns cleared, no action needed
> I checked these rather than asking you to, because this file has a history of surprising us:
> 
> * **`doSearchInternal` is 83 bytecodes _smaller_** (2,223 → 2,140). The
>   deficit accounting is simpler than the strike/stride scheme it replaces.
>   It is far past `FreqInlineSize` either way, so nothing changes about
>   inlining, but it is not growing.
> * **The compiled `Dfa` contains zero `WorkCounter` references on both arms.**
>   `WorkCounterConfig.ENABLED` is a build-time constant, so javac eliminates
>   the two new telemetry branches outright. They cost nothing in the shipped
>   jar. Worth stating explicitly somewhere, since "telemetry in the innermost
>   loop" reads alarming until you know that.
> 
> ### The one thing I would still like to understand
> Repayment got much more aggressive, and it resets the exponential backoff.
> 
> For `AcceleratorPolicy.LITERAL` (`minProfitableSkip = 16`, `strikeBudget = 16`, so `lossLimit = 256`), consider the state right after a quarantine fires, where `skippedWorkDeficit = lossLimit >>> 1 = 128`:
> 
> what it takes for one scan to clear the deficit and reset `quarantineWindow` to 2048
> before	a single **stride of 2,048** (`stride >>> 8` must repay 8 strikes)
> after	a single **skip of 144** (`max(0, 128 + 16 - skipped)` reaches 0)
> So roughly 14× less sparse evidence now un-does a quarantine, _and_ it discards the accumulated backoff window rather than decaying it. The shape I cannot rule out is _(dense region)(~150-unit gap)(dense region)(~150-unit gap)…_, where every gap resets the window and backoff never compounds, so the accelerator keeps paying per-call cost across every dense region for the whole input.
> 
> I do not have a workload that does this — it may well be unreachable in practice. But it is the adversarial-input question, and `DfaStartAccelerationWorkTest` is already the right place to answer it: a case on an alternating input asserting that `largestQuarantine()` actually escalates would settle it in either direction.
> 
> If it does reproduce, the smallest fix is probably to decouple the two effects — let a profitable call repay the deficit, but keep `quarantineWindow` where backoff left it (or halve it) instead of snapping back to `initialQuarantineWindow`. That preserves the memory of repeated failure while still letting a genuinely-good region resume acceleration.
> 
> Relatedly: nothing currently exercises quarantine-then-repay at all. Both test cases assert steady states — `quarantines() == 0` on profitable input, backoff on dense input. If the repayment branch is worth having, something should show it recovering.
> 
> ### Small
> `strikeBudget` now carries two unit systems: `Dfa` reads it as a multiplier (`strikeBudget() * minSkip`, input units) while `StringLiteralSearch` still reads it as a literal count of strikes. The javadoc is explicit about the split so this is not a bug, but would a derived `lossLimit()` accessor on the policy be worth it? It names the quantity, keeps the multiply out of a hot method, and leaves `strikeBudget` meaning one thing per caller.

Thanks for the independent measurements and the alternating-input example. I added tests showing that repeated dense regions with 200-character gaps do escalate quarantine on String and ASCII UTF-8 input, while 500-character gaps allow acceleration to recover after quarantine.

I also tried keeping the full deficit after quarantine, retaining the quarantine window while repaying the deficit, and probing during quarantine. Keeping the full deficit made the 200-gap workload 29.5% slower; retaining the window made it 14.4% slower; probes were essentially flat. Escalating sooner saves accelerator calls but can cost more in scalar DFA scanning, so I’m keeping the current reset policy.

I agree #890 should remain open for the performance gap this PR does not recover. On `lossLimit()`, the multiplication currently happens once per search, outside the scan loop; I’d leave the accessor change out of this PR.

eaftan:
> Some more code feedback from the agent
> 
> Two follow-ups
> 
> **`minProfitableSkip` displaced the one constant in this file that had a derivation.** The old path read `minDensityStride`, and `AcceleratorPolicy` records where 64 came from: _"anchoring on a single character saves roughly 0.1 ns/byte over `String.indexOf(String, int)` and spends one verification per anchor occurrence, so it stops paying at roughly one candidate per 64–90 bytes."_
> 
> This PR drops `minDensityStride` from `doSearchInternal` and puts `minProfitableSkip` in its place, where it now does three jobs at once: the profit test (`skipped < minSkip`), the loss unit (`deficit += minSkip - skipped`), and the quarantine trigger (`lossLimit = strikeBudget * minSkip`). `LITERAL` is 16, and the record still carries `// TODO: Conduct systematic empirical micro-benchmarking ... to precisely tune minimum profitable skip thresholds`.
> 
> I'm not claiming 16 is wrong — the 2.75× says it isn't wrong here. But three uses of one constant compound, and the number now carrying them is the one the file itself flags as unmeasured. A sweep of 8 / 16 / 24 / 32 for `LITERAL` on the #890 workload would settle it, and would say whether `CHAR_CLASS`'s 24 and `VECTOR_MULTI_LITERAL`'s 64 are in the right place too.
> 
> **Coverage is `LITERAL`-only.** All six tests run `record:[^\r\n;]*val=200`. `VECTOR_MULTI_LITERAL` (`minSkip` 64, so `lossLimit` 1024) and `CHAR_CLASS` (24) are where the new arithmetic diverges most from the old, and neither is exercised.
> 
> Two smaller notes, neither blocking:
> 
> * `accelerationResumePos = nextPos + Math.min(quarantineWindow, textLen - nextPos)` quietly fixes an overflow the old `nextPos + quarantineWindow` had on inputs near `Integer.MAX_VALUE`. Worth a line in the description rather than passing as incidental.
> * The quarantine state is still per-search — `skippedWorkDeficit`, `quarantineWindow` and `accelerationResumePos` are locals in `doSearchInternal`, so a `find()` loop forgets the quarantine and re-pays up to `lossLimit` on each call. Not a regression, the old code was the same. But [Flush and rebuild the DFA state cache on budget exhaustion #908](https://github.com/eaftan/safere/pull/908) landed cross-search accounting for the DFA cache in this exact file, on the argument that per-search state loses information across a `find()` loop. The counter-argument — a quarantine is cheap to relearn, a state cache is not — is probably right, but worth writing down rather than inheriting.

I benchmarked `LITERAL` thresholds 8, 16, 24, and 32. The apparent difference between 8 and the current 16 did not persist in longer runs; 24 and 32 made the #890 String workload roughly 3× slower. I also tested `CHAR_CLASS` thresholds 12, 16, 24, and 32 on dense, sparse, and alternating String/UTF-8 inputs. A possible benefit at 12 did not repeat in longer runs, so I’m keeping the current thresholds.

The PR now includes String/UTF-8 tests for dense and sparse character-class inputs and a comment explaining why quarantine state is local to each search. I tried to exercise `VECTOR_MULTI_LITERAL`, but the patterns I tested selected earlier fast paths or another start plan; its threshold remains uncalibrated. I’ll add a note about the overflow improvement to the PR description.

eaftan:
@cushon I propose merging this PR and keeping #890 open for further work to recover the performance lost when we retired the multi-anchor gap executor. This change doesn’t restore all of it, but it does recover much of the `guardByteDelimitedRecord` regression.

cushon (approved):
Thanks, that addresses all of the points I raised, merging this SGTM.

The agent agrees, with a couple of notes about the details:

---

Verified locally against 14da257a — the new tests pass, and bracketing escalation at 200 against recovery at 500 is a better shape than what I suggested.

You're right on `lossLimit`; it's hoisted at `Dfa.java:1730`, outside the loop.

One note on the sweep: 24 and 32 being ~3× slower is less about profitability than about how fast the policy gives up. `minProfitableSkip` is the profit test, the deficit unit, and a factor in `lossLimit`, so raising it 16→24 raises the per-call deficit 50% and quarantines proportionally sooner. The number is validated, but the three roles can't be tuned independently — worth capturing on #890 rather than here.

On `VECTOR_MULTI_LITERAL`: I think the patterns weren't the problem. It has only two sites, both in `Utf8StartAccelerator`, and `create` only yields `Teddy`/`MultiLiteral` inside `if (vectorProviderAvailable())` (lines 47, 57). `loadSelected()` returns null unless `-Dorg.safere.experimental.vectorScanProvider=vector`, so without the incubator flags no pattern can reach it. Try `--add-modules=jdk.incubator.vector -Dorg.safere.experimental.vectorScanProvider=vector` with a `Utf8Matcher` and a `MultiLiteral` start plan with no word boundary.

Separately, `LINE_ANCHOR` is also at 16 and wasn't in the sweep. It's String-only (`StringStartAccelerator.LineAnchor`) and needs no flags, so it's more reachable than the vector one.

PR #899, opened 2026-09-17
Benchmark `fixedAnchorLogCaptures` above the anchored-scan window

cushon:
`StringLiteralSearch` declines its anchored scan below `MIN_ANCHORED_WINDOW`,
which is `STRIKE_BUDGET * MIN_DENSITY_STRIDE` = **1,024**. At or above that it
plays the anchored gamble: repeated `indexOf` on the literal's rarest character
with per-hit verification, charging a strike whenever the stride comes in under
`MIN_DENSITY_STRIDE`, and bailing out to `indexOfDirect` once the budget is
spent.

`fixedAnchorLogCaptures` ran at **256 and 1,000** only. Both sit below 1,024, so
the workload never exercised the branch the policy actually selects for this
pattern on any realistic input. It straddled the threshold without crossing it.

This adds **2,048**. The sizes are now 256 / 1,000 / 2,048, which brackets the
threshold instead of sitting entirely on one side of it.

## Why 2,048

The cost above the threshold is a fixed per-call charge, not a change in scan
rate, so the smallest size that clears the gate captures the whole effect —
there is nothing to gain from going further, and a larger input would just cost
suite time. A separate sweep put the step between 1,024 and 1,100 and measured
the same delta at 2,048 and at 4,096.

## What the new size sees

Three interleaved cycles on x86_64 (AMD EPYC 7B13, JDK 26), `-f 2 -wi 2 -i 5`,
median of cycles, per-arm corpora. Both arms carry this same added size, so the
comparison isolates the engine change rather than the grid:

*   `pre-883` = `12b7f3e7`, the parent of
    [#883](https://github.com/eaftan/safere/pull/883)
*   `main` = `d5690591`

| workload | pre-883 ns | main ns | Δ ns | main/pre |
| `match.256` | 222.4 | 226.8 | +4.4 | 1.020 |
| `match.1000` | 375.9 | 384.0 | +8.1 | 1.022 |
| **`match.2048`** | 517.9 | 635.7 | **+117.8** | **1.227** |
| `noMatch.256` | 60.2 | 60.8 | +0.6 | 1.010 |
| `noMatch.1000` | 176.7 | 175.0 | −1.7 | 0.991 |
| **`noMatch.2048`** | 314.1 | 433.0 | **+118.8** | **1.378** |

Per-cycle ratios agree tightly — 0.815 / 0.819 / 0.808 on `match.2048` and
0.725 / 0.725 / 0.731 on `noMatch.2048` — and the within-arm spread is at most
4.7%, mostly under 2%.

Two things to note:

*   **At the existing sizes the two arms are indistinguishable**, within 2.2%.
    That is the coverage hole stated plainly: #883 could not have been caught
    here.
*   **The two 2,048 deltas are the same number** — +117.8 and +118.8 ns — even
    though the arms differ by 1.6× in absolute time and do different amounts of
    work. That is the signature of a **fixed cost per `find()` call**, not a
    change in per-character rate, which is what the anchored gamble's fixed
    strike budget predicts.

The same shape shows up within `main` alone: extrapolating `noMatch` from its
own 256→1,000 slope of 0.1535 ns/char predicts 335.9 ns at 2,048, against 433.0
measured — **+97 ns** of nonlinearity that the old size grid could not resolve.

## Scope

Benchmark data only; no engine change, and nothing here is a fix.

The pattern is `(error:\[)[A-Z](\] code:500)`, whose required literal
`] code:500` anchors on `']'`. The rarity oracle ranks `']'` at 109, near the
top of its scale, but in a log line it recurs roughly every 46 characters —
below `MIN_DENSITY_STRIDE` of 64 — so every iteration strikes, the loop spends
its whole budget, and then falls back and does the full scan regardless. The
remedy is a separate change; this PR only makes the effect observable.

The sibling `fixedAnchorLog` workload already runs at 1,000 / 10,000 / 100,000
and needs nothing.


eaftan (approved):
Useful benchmark coverage that lays the foundation for #901. LGTM.

PR #900, opened 2026-09-17
Keep the longer start literal unless the shorter one scans faster

cushon:
`MultiAnchorCompiler.extractStartPlan` arbitrates between two start-acceleration
candidates: the leading literal prefix, and a fixed-offset literal found later in
the pattern. [#876](https://github.com/eaftan/safere/pull/876) removed the length
gate and the 2x margin standing in front of that comparison, leaving a direct
`folScore > prefixScore`. That was right for the patterns it measured, and it is
wrong whenever the fixed-offset literal is *shorter*.

`RarityOracle.literalSelectivityScore` sums per-character rarity, so it can rank
a shorter literal above a longer one whenever the shorter one is built from rarer
characters. On `https?://...` it ranks `://` above `http`, because `:` and `/`
are rarer than `h`, `t` and `p`. As a ranking of characters that is correct. As a
ranking of filters it is not: the two co-occur one-for-one in a URL, so `://`
filters exactly as much as `http` while giving back three characters and adding a
back-check at every candidate offset.

Shorter is not the same as worse, though, which is why this is narrower than a
length gate. A one-character literal declines anchoring in
`StringLiteralSearch.anchorOffset` and takes the single-character path of the
`indexOf` intrinsic, which scans faster per byte than the multi-character one. It
can win outright despite filtering less — that is where #876's 1.19x on
`uuidExtraction` comes from, taking `-` over `id:`.

So the flip is refused only when the fixed-offset literal is shorter than the
prefix **and** longer than one character.

### Measured

Four arms, `-f 6 -wi 3 -i 8 -r 1`, three interleaved cycles, x86_64. The third
arm is the second build with no property set, as an equivalence control; the
fourth forces `RejectPlan.None` to separate the start-plan change from the Tier 0
prefilter it re-enables.

| arm | Tier 1 start scan | Tier 0 reject | `urlExtraction` ns |
| :--- | :--- | :--- | ---: |
| A `main` | `FixedOffset "://"` | `None` | 1134.2 |
| B `guard` | `Literal "http"` | `RequiredLiteral "://"` | 1085.5 |
| C `guard`, second build | as B | as B | 1084.0 |
| D `guard`, Tier 0 forced off | `Literal "http"` | `None` | 1066.0 |

| | ratio | |
| :--- | ---: | :--- |
| A / D | **1.064x** | the start-plan change on its own |
| D / B | **0.982x** | what the re-enabled Tier 0 prefilter costs here |
| A / B | **1.045x** | net, as shipped |
| B / C | 1.001x | build-equivalence control |

The three groups do not overlap across any of the three cycles: A spans
1122.9-1139.1, B and C together span 1083.0-1091.1, D spans 1060.6-1066.8.

The 1.8% that Tier 0 costs is expected on this workload and is not an argument
against it. `urlExtraction`'s input is 153 characters containing three URLs under
`findAllLengthSum`, so `RequiredLiteral("://")` never rejects anything and only
ever adds a scan. On an input with no `://` in it the same prefilter is what
rejects the whole thing in one pass. Abandoning a prefilter observed not to
filter is tracked in [#890](https://github.com/eaftan/safere/issues/890); this is
another data point for it.

### Nothing else moves

The whole 21-trial cross-engine suite, two arms, three interleaved cycles:
geomean **0.995x**, every trial inside its own noise band. `urlExtraction` is the
only trial in that suite whose compile-time plan differs between the arms —
verified by dumping `startPlan` and `rejectPlan` for every pattern in
`benchmark-data.json` — so the other twenty are a noise floor rather than a test.
The largest deviation among them was 5.2%, on a trial whose plan is byte-identical
in both arms.

### Corpus sweep

46,866 distinct patterns from 61,534 call sites, comparing the compile-time plan
with and without the guard:

| | patterns | share | call sites |
| :--- | ---: | ---: | ---: |
| `startPlan` reverted to the leading prefix | 257 | 0.548% | 350 |
| reverted a single-character literal | 0 | | |

254 are `FixedOffset` -> `Literal`; the other three are the inner plan of a
`LeadingExpansion`. The dominant shape is `://` -> `http` at 130 patterns, half
the total, then `google` -> `support` 20, `(files` -> `contains` 3,
`-d/` -> `/cns/` 3, `type` -> `metric` 3, `;base64,` -> `data:image/` 3.

**116 of the 257 also go from `RejectPlan.None` to
`RejectPlan.RequiredLiteral`.** `extractRejectPlan` excludes `fol.literal()` and
everything `extractFixedPrefixLiterals` collects whenever the start plan is
`FixedOffset`, which on a fixed-width pattern is every literal in it, leaving
nothing to require. Reverting to `Literal` excludes only the prefix. So
`urlExtraction` ends up with the longer literal driving the start scan and the
rarer one rejecting the whole input, which neither #876 nor its predecessor had.

23,460 tests pass.


eaftan:
@cushon I reproduced the URL improvement, but found two counterexamples where the length rule loses useful filtering: frequent `error:A?? ` records followed by one `error:A!!` match for `error:[A-Z]!!`, and frequent `support0normal ` records followed by one `support0google` match for `support[0-9]google`.

| Pattern | prefixUnit | suffix | main ns/op | PR ns/op | PR/main |
|---|---|---|---:|---:|---:|
| `error:[A-Z]!!` | `error:A?? ` | `error:A!!` | 272 ± 9.7 | 22,419 ± 367 | 82.4 |
| `support[0-9]google` | `support0normal ` | `support0google` | 272 ± 18 | 28,614 ± 1545 | 105.2 |

I opened draft #903 with just those benchmark workloads, the paired results, and reproduction commands. In the earlier paired run against this PR's production base, they took about 82× and 105× as long with this change. The whole-input reject prefilter passes because the rare suffix occurs in the final matching record, so it doesn't recover the lost start filtering.

Would you take a look? Can we preserve the URL improvement without losing the shorter literal's filtering on these inputs?

cushon:
Thanks @eaftan, I have a variation of this in #906 that I'm not done validating yet, but that I think might have similar benefits without the regression in the cases you found.

cushon:
Superseded by #906

PR #901, opened 2026-09-17
Concede the anchored literal scan after four dense observations

cushon:
`StringLiteralSearch.indexOf` gambles that a literal's rarest character is also
rare in the haystack. When the gamble loses it pays for the discovery on every
call, forever, because nothing carries the verdict past the end of the call.
The strike counter that is supposed to bound that cost is a *net* count with a
budget of 16, and a net count answers the wrong question: an anchor that is
uniformly dense has shown everything it is going to show within a handful of
observations, but it still has to pay fifteen more before conceding.

Charging four per dense stride instead of one, against the same budget of 16,
makes a uniformly dense anchor concede on its fourth observation while leaving
repayment intact for the anchor that merely clumps.

### The cost, and what this recovers

`(error:\[)[A-Z](\] code:500)` anchors on `']'`, which the rarity oracle scores
at 109 — near the top of its scale — but which recurs every 46 characters in a
log line, inside `MIN_DENSITY_STRIDE` of 64. So every iteration strikes, the
loop burns the whole budget, and then scans the window again with
`String.indexOf`.

`fixedAnchorLogCaptures` at size 2048, x86_64, three interleaved cycles,
`-f 4 -wi 3 -i 6`:

| trial | main | this PR | Δ ns | ratio |
| `match.256` | 226.1 | 226.3 | −0.2 | 0.999 |
| `match.1000` | 379.0 | 381.0 | −2.0 | 0.995 |
| **`match.2048`** | **636.3** | **555.2** | **+81.1** | **1.146** |
| `noMatch.256` | 60.3 | 60.3 | 0.0 | 1.001 |
| `noMatch.1000` | 175.3 | 175.1 | +0.2 | 1.001 |
| **`noMatch.2048`** | **432.2** | **348.8** | **+83.3** | **1.239** |

The two 2048 groups do not overlap in any cycle: `noMatch` main
{430.5, 432.2, 434.2} against {348.8, 348.8, 352.5}, `match` main
{633.9, 636.3, 636.5} against {554.9, 555.2, 558.0}. Every sub-threshold trial
overlaps, so this costs nothing where it does not apply.

The `noMatch` 256→1000 slope is 0.1546 ns/char, which extrapolates size 2048 to
337.3 ns. Measured excess over that is **94.9 ns on main and 11.5 ns here — 88%
of the fixed cost recovered**. The absolute Δ is nearly identical on `match`
(81.1) and `noMatch` (83.3) despite baselines 200 ns apart, which is the
signature of a fixed per-call cost rather than a change in scan rate.

### Why weighting the charge rather than counting consecutive strikes

The obvious implementation is a second counter for consecutive strikes, reset on
every sparse stride. It was written, and reverted: a separate local with its own
increment, test and reset took `indexOf` from **201 bytecodes to 216**, and 219
is a size at which this method has already been measured to stop being inlined —
costing more than the fix recovers. The javadoc on `indexOf` warns about exactly
this, and it has bitten six times across this file's history, twice on aarch64
only.

`iinc` takes its constant as an operand, so `strikes += 4` is byte-for-byte the
same instruction as `++strikes`. The method is **201 bytecodes on this branch,
identical to main**, with `iinc 11, 4` in place of `iinc 11, 1`.

One subtlety worth recording: the charge has to be a literal. Writing it as
`STRIKE_BUDGET / 4` is not a compile-time constant, because `STRIKE_BUDGET` is
read from `AcceleratorPolicy`, and it compiled to `getstatic`/`iadd`/`dup` for
**205 bytecodes** instead of folding into the existing `iinc`.

### Why this is not the same as lowering the budget to 4

Charging 4 and repaying 1 against a budget of 16 makes a sparse stride worth a
quarter of a dense one. Setting the budget to 4 and keeping charge and repayment
at 1 makes them equal, and that is materially weaker: an anchor that is dense
for three strides and sparse for four, repeating — 43% dense — never concedes
under the equal-weight rule, because each sparse stride buys back a whole dense
one. Under the weighted rule it concedes on its second cycle, while a genuinely
good anchor (two dense, ten sparse) still never does.

Keeping `STRIKE_BUDGET` at 16 also leaves `MIN_ANCHORED_WINDOW` at exactly 1024,
so no window that previously delegated to `String.indexOf` now anchors. Only the
*rationale* for that constant changes: a verdict is now reachable four strides
in, which makes 1024 conservative rather than exact. Lowering it to match would
newly anchor every window between 256 and 1024, which is a much larger change
and wants its own measurement.

### Tests

Two differential tests in `StringLiteralSearchTest`, both against
`String.indexOf`:

*   `agreesWithJdkWhenTheAnchorRecursJustInsideTheDensityStride` — the 46-stride
    shape the charge is calibrated on, at every 137th start index.
*   `agreesWithJdkWhenTheAnchorClumpsAndThenThinsOut` — five dense observations
    followed by a long sparse run, which repayment has to keep alive.

They pin correctness under early concession, not the concession point itself.
**The concession point is not observable from the result, and the work counter
cannot see it either**: both rules charge the same total characters, since the
gap scans and the fallback scan partition the text either way and differ only by
`12 × literalLength` out of ~3,000. The real cost is per-intrinsic-call setup,
which `WorkCounter` does not model. Only the benchmark can see this policy — and
only above 1,024 characters, which is why
[#899](https://github.com/eaftan/safere/pull/899) is a prerequisite for the
suite being able to observe it. There is no code dependency.

### What this does not address

This is remedy 1 of three in Project 61 and it is the cheapest one. It does
nothing for the first limb of the problem — a **dense but correct** anchor,
where every candidate verifies, returns on the first iteration and the density
test never runs at all. `guardByteDelimitedRecord` has `record:` and `:` at the
same 2,941 positions, a stride of 16.8 against a break-even near 89, and is
invisible to any strike rule. Sizing the gate on the inter-anchor stride rather
than on the remaining input length (remedy 2) is the fix for that, and is
unaffected by this change.


eaftan (approved):
LGTM and nice win.  Thanks for finding this and for documenting why you chose to weight the existing counter rather than add a second one.

PR #903, opened 2026-09-18
Add start-literal filtering benchmark counterexamples

eaftan:
Adds two small benchmark workloads for the start-literal selection discussion in #900. Each has frequent leading prefixes in nonmatching records and a rare shorter required literal in one matching record at the end. Selecting the longer prefix loses the filtering provided by the shorter literal; the whole-input reject prefilter passes because the suffix occurs in that final match.

Only `benchmark-data.json` changes. No production policy changes or reverts. This is a draft to share reproducible counterexamples before deciding how #900 should handle them.

Refs #900.

| Workload | Pattern | Repeated decoy | Final matching suffix | Input length |
|---|---|---|---|---:|
| `StartLiteralSelectionBenchmark.error` | `error:[A-Z]!!` | `error:A?? ` | `error:A!!` | 10000 |
| `StartLiteralSelectionBenchmark.support` | `support[0-9]google` | `support0normal ` | `support0google` | 10000 |

Both inputs use `suffixToLength`; both workloads use `findAllLengthSum` and consume the integer result.

Earlier paired measurements used main production at `d5690591a11cfd550b3eeae9da00fc87b5b60af5` and submitted #900 at `fd48cd1a471658bc751775fef000771a2be5c41f`, with identical workload declarations added to each arm. These are the earlier review measurements, not a new comparison against the newer base of this draft. JDK 26.0.2; standard wrapper configuration: 2 forks, 2 warmup × 500 ms, 5 measurement × 500 ms. `PR/main` is elapsed time, so lower is better.

| Workload | main ns/op | #900 ns/op | PR/main | Result |
|---|---:|---:|---:|---|
| error | 272 ± 9.7 | 22,419 ± 367 | 82.4 | Takes 82.4× as long |
| support | 272 ± 18 | 28,614 ± 1545 | 105.2 | Takes 105.2× as long |

To run the String trials on this branch:

```bash
./run-java-benchmarks.sh '^org\.safere\.benchmark\.CrossEngineBenchmark\.run$' -- \
  -p 'crossEngineTrial=StartLiteralSelectionBenchmark.error@safere-string,StartLiteralSelectionBenchmark.support@safere-string'
```

For a paired reproduction, apply this benchmark-only commit unchanged to each production revision, then use the controlled refs:

```bash
  --baseline <baseline-with-workloads> --current <900-with-workloads> \
  'StartLiteralSelectionBenchmark\.(error|support)@safere-string'
```

Validation: existing JSON declarations verified unchanged; `git diff --check` and `mvn -pl safere-benchmarks -am spotless:check -q` passed. Focused smoke execution of both String trials passed using `./run-java-benchmarks.sh --smoke` with the trial selection shown above (build and input materialization included). Smoke timings are execution checks, not performance evidence. Full library tests, generated crosscheck, other engines, and a new paired performance run against today's main have not been run.


eaftan:
Closing this as I think we've dealt with the issue

PR #904, opened 2026-09-18
Benchmark and test the anchored scan against false candidates

cushon:
No input in the suite exercises the anchored start scan against a candidate it
has to reject.

Of the 122 benchmark patterns that compile to a literal start plan, 94 anchor on
a character, and all 25 runnable inputs among them contain exactly as many anchor
occurrences as literal occurrences:

| trial | anchor occurrences | literal occurrences |
| :--- | ---: | ---: |
| `guardByteDelimitedRecord` | 2,941 | 2,941 |
| `guardByteHeader` | 2,857 | 2,857 |
| `markupImageLink` | 11,111 | 11,111 |

Every candidate the scan produces is a match. So neither the candidate pre-check
in `StringLiteralSearch.indexOf` nor the strike counter that abandons the scan
has ever been measured doing the thing it is there for, and a change to either
would look free.

`guardByteRecordNoise` runs the existing `record:[^\r\n;]*val=200` over lines
that begin `other:` instead. The start plan is `Literal[record:]`, which anchors
on the `:`, so the scan stops 1,352 times per 100 KB and rejects all but at most
one of them.

Two details decide whether the trial measures that or something else, and both
were found by getting them wrong first.

**The line is 74 characters.** The strike counter abandons the anchored scan
below `MIN_DENSITY_STRIDE`, 64; at a shorter stride the trial quietly measures
the fallback instead. Both sizes are also above `MIN_ANCHORED_WINDOW`, 1,024.

**The `noMatch` input contains one `val=200`**, placed on an `other:` line so it
cannot complete a match. The reject prefilter for this pattern is `val=200`, so
without it the trial is one failed prefilter scan over 100 KB and never reaches
the start accelerator at all — **9,927 ns/op, against 25,734 with it**.

`StringLiteralSearchTest` covers the differential half of the same gap: literals
whose rarest character is also their last, over haystacks carrying false
candidates. Where the two coincide the pre-check re-reads the byte the anchor
scan just returned, so it always passes and every candidate takes a full
verification; a haystack whose anchor occurrences are all real matches never
reaches that verification with anything to reject.


eaftan (approved):
LGTM.  Thanks for adding this workload and the test.

Issue #789, opened 2026-08-30
Investigate intermittent Maven Central HTTP 429 failures in CI


eaftan:
## Summary

SafeRE CI has intermittently failed before compilation because Maven Central returned HTTP 429
while Maven was resolving dependencies. We believe we have seen a couple of other occurrences, but
only the example below has been verified and linked so far.

## Verified example

- 2026-08-30, `fuzz-regression` job for PR #776:
  https://github.com/eaftan/safere/actions/runs/33290772409/job/99201908427

The `Build SafeRE for fuzz tests` step failed while resolving
`org.junit:junit-bom:pom:6.1.3` from `https://repo.maven.apache.org/maven2`. Maven reported HTTP 429
`Too Many Requests`, then cached the failed transfer and could not build the reactor. Compilation
and fuzz regression execution never began.

## Impact

- Produces red CI unrelated to the proposed code change.
- Requires another commit or manual rerun to recover.
- Can affect early build/setup steps before project tests provide any signal.

## Investigation

- Review Sonatype's guidance on Maven Central 429 responses:
  https://central.sonatype.org/faq/429-error/
- Determine how frequently Maven Central 429 responses occur in SafeRE Actions jobs.
- Check whether the workflow's parallel jobs independently populate empty Maven caches and create a
  burst of identical dependency requests.
- Evaluate a bounded retry/backoff for transient dependency-resolution failures.
- Evaluate whether cache setup, dependency prefetching, or workflow fan-out can reduce redundant
  requests without serializing unrelated test execution.
- Add links here if the other remembered occurrences can be identified.

Any mitigation should remain bounded and should not hide persistent dependency or POM failures.

eaftan:
another one here: https://github.com/eaftan/safere/actions/runs/35334985529/job/105568303843?pr=905

PR #905, opened 2026-09-18
Search for a one-character start literal as a character

cushon:
`String.indexOf(String, int)` with a one-character needle does not reach the
JDK's character kernel on x86_64. Measured over the same 100,000-byte input:
0.098 ns/byte for the string form against 0.0155 for `String.indexOf(int, int)`,
a factor of 6.3. Nothing between `StringStartAccelerator` and the intrinsic
looks at the needle's length, so a one-character literal does not find the
faster kernel by itself.

Decide which kernel to use when the plan is built, and carry the answer in the
accelerator record beside the `anchorOffset` / `anchor` pair it already holds.
`Literal` and `FixedOffset` both gain a `singleAsciiChar` component and both
dispatch through `StringLiteralSearch.indexOfPlanned`.

## Where the branch lives, and why it matters more than the branch

The test is at plan time rather than in `indexOfDirect` because putting it there
was measured and rejected. It took that method from 8 bytecodes to 40 and
charged every caller for a branch only 8% of literals can take, including the
anchored path that reaches it as a fallback. On aarch64, where the two kernels
are the same speed to four digits, that cost alone moved two unrelated trials by
5–6% — `isoTimestampLog.noMatch.100000` −5.0% and
`fixedAnchorLogCaptures.match.1000` −6.3%, neither of which can take the branch.
Across 21 trials that version was 0.989× geomean on ARM.

This version leaves `indexOf` at 201 bytecodes and `indexOfDirect` at 8, both
byte-for-byte unchanged, and both of those trials come back clean.

| method | `d5690591` | this PR | |
| :--- | ---: | ---: | :--- |
| `StringLiteralSearch.indexOf` | 201 | **201** | unchanged |
| `StringLiteralSearch.indexOfDirect` | 8 | **8** | unchanged |
| `StringLiteralSearch.indexOfChar` | — | 16 | new |
| `StringLiteralSearch.indexOfPlanned` | — | 26 | new |
| `Literal.findCandidate` | 17 | 21 | +4 |
| `FixedOffset.findCandidate` | 21 | 25 | +4 |
| `FixedOffset.nextFixedOffsetCandidate` | 250 | 252 | +2 |
| `RejectPrefilter.Literal.canReject` | 34 | 34 | unchanged |

## x86_64

EPYC 7B13, JDK 26, three interleaved cycles per arm, `main`/PR ratios with
per-cycle ratios in brackets:

```
uuidExtraction.noMatch.100000    6.377x  [6.443 6.380 6.356]
uuidExtraction.match.100000      5.588x  [5.559 5.637 5.581]
schemeExtract.1048576            2.978x  [3.072 3.223 2.978]
findSlashNoise                   0.972   [0.990 0.966 0.971]
```

## aarch64

Apple M-series, JDK 26, 11 trials, three interleaved cycles per arm. Six trials
the change routes, five controls it cannot reach:

```
trial                                      kind     main       PR   main/PR   per-cycle
findSlashNoise                           routed  10601.7   9329.5     1.136*  1.142 1.142 1.134
fixedAnchorLogCaptures.match.1000       control    446.6    436.1     1.024*  1.024 1.005 1.029
guardByteDelimitedRecord.noMatch.100000 control  23445.6  23093.0     1.015   1.015 0.974 1.072
uuidExtraction.noMatch.100000            routed   3121.1   3088.8     1.010   1.043 0.972 0.909
uuidValidation                           routed    370.9    370.0     1.003   1.042 0.999 0.992
isoTimestampLog.noMatch.100000          control   3204.7   3207.8     0.999   0.999 1.008 0.882
urlExtraction                           control    734.0    738.4     0.994   0.953 1.002 1.007
uuidExtraction.match.100000              routed   3252.5   3272.8     0.994   0.974 1.020 1.025
quotedString.4096                        routed    367.2    372.1     0.987   1.000 0.996 0.971
logLineParse                            control   1879.1   1929.3     0.974   1.138 0.883 0.976
schemeExtract.1048576                    routed     73.1     75.5     0.968   1.015 0.968 0.978

geomean over 6 routed trials:  1.0150
geomean over 5 controls:       1.0011
```

`*` = the two arms' raw numbers are disjoint across all three cycles.

The two trials the earlier version taxed are the ones to read first:
`fixedAnchorLogCaptures.match.1000` goes from −6.3% to 1.024, and
`isoTimestampLog.noMatch.100000` from −5.0% to 0.999.

## The two ISAs trade places

`findSlashNoise` is −2.8% on x86 and **+13.6% on ARM**, disjoint and tight.
`schemeExtract` is the reverse: 2.978× on x86 and 0.968 on ARM.

The original premise was that ARM had no win available, because the two kernels
measure identically there — 0.0282 ns/byte either way. That measurement is
correct but it is a statement about *per-byte rate*, taken from a single call
over 100,000 bytes. It says nothing about per-call cost.

`findSlashNoise` is the workload that separates the two. Its input repeats
`service 42 completed path /api/v1/items with 7 records `, so consecutive
slashes are a median of 4 characters apart and it makes tens of thousands of
calls that each scan almost nothing. On ARM `indexOf(char)` is cheaper to
*enter* even though it scans at the same rate, and that is the whole +13.6%. On
x86 the ordering inverts: the character kernel scans 6.3× faster but costs more
to enter, so the same call-dense workload loses 2.8% while long scans gain
multiples.

So the change is worth something on both ISAs, for different reasons, and the
x86 `findSlashNoise` regression is specific to x86 rather than intrinsic to the
design. It is still a real trade there — 0.971 and 0.972 across two independent
runs, while several other sub-5% movements in the same runs did not reproduce at
all.

A length guard would not recover it. The quantity that decides the outcome is
how far the call actually scans, and the only quantity available at the call is
how much text remains — ~32,000 characters here, indistinguishable from a long
scan. Separating the two needs runtime feedback of the kind `indexOf` already
applies through `STRIKE_BUDGET`, which is a change of its own.

`schemeExtract` at 0.968 on ARM is inside the noise. It is a 1 MB single-scan
workload, so per-call cost is amortised across a million bytes and there is no
mechanism by which routing it to an equally fast kernel can cost 3%. Its
per-cycle ratios straddle 1.0, and the five control trials — which by
construction cannot be affected — themselves span 0.974 to 1.024 in this run.

## Blast radius

Over the 46,866-pattern corpus: 3,827 patterns (**8.17%**) and 5,672 of 61,534
call sites (**9.22%**) compile to a start plan this routes. 897 of those patterns
(23.4%) also carry a reject prefilter — 535 `RequiredLiteral`, 184
`EndAnchoredSuffix`, 171 `EndAnchoredCharClass`, 7 `Composite` — which can reject
before the start accelerator runs, so on non-matching input the reach is 2,930
patterns (6.25%) and 4,620 call sites (7.51%). The most common routed characters
by call site are `-`, `/`, `<`, `[`, `{` and `.`.

## What it declines

Non-ASCII, where `String.indexOf(int, int)` matches by code point and so is a
different search, and case-folded literals, which reach a different accelerator.

`RejectPrefilter.Literal` is the third caller of `indexOf` and is deliberately
left alone: `MultiAnchorCompiler.extractRequiredLiteral` will not return a
literal shorter than two characters, so the branch would be dead there.

`indexOfChar` charges the work a one-character `indexOfDirect` would, clamping
`fromIndex` the way `indexOf` does before it delegates. A search must not record
different work depending on which kernel its plan picked, or the scan-dispatch
audit reports the divergence as a bug somewhere else.


eaftan (approved):
Huh, this is fascinating.  How did you figure out that `indexOf(String)` is slower than `indexOf(int)`?

Your results repro for me.  LGTM

eaftan:
The CI failure is another instance of #789, which I thought I had fixed.  It's a transient failure.  I'll rerun the failed job but I expect it to pass.

eaftan:
PR Shepherd is now monitoring this PR at `00cdd98bf6a8fc22cbc0f3b80144b22d0960617e`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `00cdd98bf6a8fc22cbc0f3b80144b22d0960617e`.

cushon:
The difference between `indexOf(String)` and `indexOf(int)` jumped out comparing benchmark results that I happened to run on both x86 and aarch64. I also filed https://bugs.openjdk.org/browse/JDK-8392735 for that, it seems like hotspot is probably missing an optimization on x86 that's present in the aarch64 backend.

PR #906, opened 2026-09-18
Resolve multi-offset start candidates against the leading character class

cushon:
This is a different approach to improving the same area as #900, without the regressions discussed in the review for #900.

---

`StringStartAccelerator.FixedOffset` and `Utf8StartAccelerator.FixedOffset` only
consulted the leading character class when the fixed-offset literal could sit at
exactly one offset. When it could sit at several, both fell through to an
unconditional retreat by `maxOffset` — the leftmost position the literal *could*
imply, rather than the one it does imply.

That is sound but imprecise, and it costs in two ways. `Matcher` treats the
accelerator's answer as the match start to try, so a position the class could
have rejected in O(1) instead becomes a full engine wakeup. And when the class
admits *none* of the offsets, the occurrence could have been skipped outright;
today it is not.

The second case is the expensive one. It fires whenever the fixed-offset literal
is common in the text while the leading class is rare, which is exactly the
shape the plan is supposed to exploit.

This walks the offsets from the largest down, returns the leftmost start the
leading class admits, and skips the occurrence entirely when it admits none.

### Measurements

Two trials added to `FixedOffsetAccelerationBenchmark`, over both accelerators.
`\b(?:(?:BUG)|(?:FIXED))=\s*(\d+,?)+` puts `=` at offsets `[3,5]` behind the
class `{B,F}`; `(?:requisition|req):(\S*)` puts `:` at `[3,11]` behind `{r}`.
Both are corpus patterns. The inputs are 32 KB of configuration- and log-shaped
text dense in the literal and free of the class.

x86_64, AMD EPYC 7B13, JDK 26.0.2.1, 3 cycles, arms interleaved and alternated,
`-f 1 -wi 2 -i 5`, medians.

| trial | main (ns) | this PR (ns) | main/PR | per-cycle |
| :--- | ---: | ---: | ---: | :--- |
| `findMultiOffsetColonNoise@safere-string` | 251734.5 | 20184.9 | **12.47x** | 12.68 11.19 12.33 |
| `findMultiOffsetColonNoise@safere-utf8` | 240813.2 | 21341.7 | **11.28x** | 11.12 11.27 11.48 |
| `findMultiOffsetEqualsNoise@safere-string` | 61535.4 | 24647.0 | **2.50x** | 2.494 2.497 2.496 |
| `findMultiOffsetEqualsNoise@safere-utf8` | 61440.8 | 26560.6 | **2.31x** | 2.313 2.311 2.324 |
| `findSlashNoise@safere-string` *(single-offset control)* | 16427.6 | 16420.1 | 1.000x | 1.002 0.972 1.004 |
| `findSlashNoise@safere-utf8` *(single-offset control)* | 15778.5 | 15834.5 | 0.996x | 0.995 0.998 0.997 |

Over the 15-trial suite used for Project 60, the geomean is 1.0021x with no
trial outside 0.984x–1.007x.

`ApplicationBenchmark.urlExtraction` reads 1.046x here and 1.065x in the suite
run, but that number should not be leaned on: an all-`http://` variant of the
same trial read 1.133x and 1.048x in two independent runs, so the effect is at
or below what this rig resolves. `urlExtraction` is the weakest member of the
family, not a typical one — `://` occurs almost only inside URLs, so the literal
and the leading class agree and there is nothing to reject.

### Why this supersedes #900

#900 recovered the `urlExtraction` regression by keeping the longer leading
prefix whenever the fixed-offset literal was shorter. That treated the symptom:
the literal was not the problem, the candidate the accelerator derived from it
was.

It also had a cost. @eaftan's counterexamples in #903 are patterns where the
shorter literal is far rarer than the longer prefix, and preferring the prefix is
catastrophic there — measured on this machine, #900 is 122x slower on
`StartLiteralSelectionBenchmark.error` and 145x slower on `.support`, taking that
15-trial geomean to 0.52x. Both trials are unaffected by this PR.

The comment #900 edited still says "There is deliberately no length gate and no
margin here. Both were tried and both cost more than they saved" — that stays
true.

### Correctness

The result is a floor, not a guess: `Matcher` assigns `effectiveStart = idx` and
will not look before it. Starts implied by a literal occurrence are **not**
ordered by the occurrence, so the leftmost admitted start is not automatically
safe to return. On `(aq|b[a-z]{9})z` over `bxxaxzxxxxz` the offsets are 2 and 10;
the `z` at index 5 admits only a start at 3, while the `z` at index 10 starts the
match at 0. Returning 3 loses the match, and
`MatcherTest.multiOffsetAccelerationPreservesEarlierStartFromLaterLiteral`
catches it.

The resolved start is therefore clamped to `literalStart + 1 - maxOffset`, the
earliest start any later occurrence could imply. That yields the exact start
whenever the offset span is at most one, and falls back to today's conservative
floor when the span is wider. Both halves are pinned by new tests in
`StartAcceleratorTest`, over the `String` and UTF-8 accelerators.

Of the 46,866 patterns in the corpus, 3,117 compile to a `FixedOffset` start
plan, and 962 of those are multi-offset with a leading class and can change
behaviour here. 417 have an offset span of at most one and resolve exactly; the
remaining 545 fall back to the clamped floor but still skip occurrences the class
rejects.

### Bytecode budget

`nextFixedOffsetCandidate` sits on the inlining cliff documented for this file,
so the multi-offset path and the cold `retreatByCodePoints` branch both moved out
of line. The scan loop ends up smaller than before:

| method | before | after |
| :--- | ---: | ---: |
| `StringStartAccelerator$FixedOffset.nextFixedOffsetCandidate` | 250 | 240 |
| `Utf8StartAccelerator$FixedOffset.nextFixedOffsetCandidate` | 229 | 220 |

The single-offset path is byte-for-byte unchanged, which is what keeps
`findSlashNoise` at parity above. An earlier shape that inlined the loop grew the
method to 280 and cost 4.7% on `findSlashNoise`, a single-offset pattern whose
semantics it did not change at all.

23,457 tests pass.


cushon:
> I'm not done validating this yet, I think the claimed benefit for `urlExtraction` below is marginal but that this approach might still be worthwhile and that's the wrong benchmark to measure the benefit.

I updated this with some additional benchmarks and a rewritten PR description, the approach is unchanged from the initial draft.

eaftan (commented):
This makes sense, and I reproduced the reported benchmark gains:

| Workload | Input | main ns/op | PR ns/op | PR/main | Result |
|---|---|---:|---:|---:|---|
| findMultiOffsetColonNoise | String | 148,822 ± 3,930 | 16,753 ± 458 | 0.113 | 8.88× faster |
| findMultiOffsetEqualsNoise | String | 41,695 ± 388 | 18,384 ± 177 | 0.441 | 2.27× faster |
| findSlashNoise control | String | 13,146 ± 299 | 13,231 ± 410 | 1.006 | Flat |
| findMultiOffsetColonNoise | UTF-8 | 148,198 ± 1,937 | 14,814 ± 584 | 0.100 | 10.0× faster |
| findMultiOffsetEqualsNoise | UTF-8 | 42,001 ± 657 | 19,328 ± 583 | 0.460 | 2.17× faster |
| findSlashNoise control | UTF-8 | 11,032 ± 948 | 10,868 ± 677 | 0.985 | Flat |

This looks like a better path than #900's length rule, because it preserves rare shorter-literal filtering. LGTM

eaftan (approved):

PR #907, opened 2026-09-19
Compare multi-offset search bounds against the JDK

cushon:
`Utf8InputFuzzer.assertMultiOffsetLiteralOccurrencesMatchString` generates
`(aq|b[a-z]{n})z`, where the two alternation branches place the literal at
widely separated offsets. That is the family where the starts a literal
occurrence implies are *not* ordered by the occurrence — an earlier occurrence
can imply only a later start while a later one begins the leftmost match — so
it is exactly the shape a mistake in start-candidate resolution shows up on.

The assertion was weaker than the generator deserved:

```java
boolean stringFound = stringMatcher.find();
if (utf8Matcher.find() != stringFound || pattern.find(utf8Input) != stringFound) {
```

Three gaps. It compares the two SafeRE engines only against each other, and
both start accelerators are edited together, so an error that moves the
reported start the same way in both is invisible. It checks only *whether* a
match exists, never the reported bounds. And it stops after the first `find()`.

There is a fourth: the generated input is pure ASCII, so `utf8Offset` is the
identity function and the UTF-8 half of the comparison cannot detect an offset
error even in principle.

This takes the JDK as the oracle, walks every match, compares `start()` and
`end()` on both engines, and gives the input a multibyte prefix so byte offsets
and UTF-16 offsets actually differ.

Confirmed the new assertion has teeth rather than assuming it: with the
leftmost-match clamp removed from both accelerators, the old assertion passes
and the new one fails. On an unmodified tree it passes.


eaftan (approved):
This makes sense and improves fuzz test coverage.  LGTM

Issue #593, opened 2026-07-25
Investigate rebar benchmark suite


eaftan:
https://github.com/BurntSushi/rebar appears to be a comprehensive benchmark suite for regex engines. It is published under the especially permissive “Unlicense.” We should look into it, run SafeRE under it, and see how it compares. We also might want to contribute a SafeRE runner to it and maybe some of our real world benchmark cases.

eaftan:
I did investigate it and ended up adding support for Rust regex and .NET non-backtracking into SafeRE's benchmark suite.  I feel that's good enough to close this issue.

PR #908, opened 2026-09-19
Flush and rebuild the DFA state cache on budget exhaustion

cushon:
I tried running SafeRE on the rebar benchmark suite (see also #593), and this change helps with a few of the cases where SafeRE underperformed relative `java.util.regex`.

---

## Summary

When `Dfa.getOrCreate` reached `maxStates` (`Pattern.MAX_DFA_STATES = 10_000`), it returned `null` without ever flushing the state cache, causing the active search (and every subsequent search on the same `ThreadLocal<Dfa>` that touched an uncached transition) to abandon the DFA and fall back to the NFA.

On patterns with large bounded-repetition working sets over multi-megabyte inputs — such as `rebar`'s `curated/10-bounded-repeat/context` (`[A-Za-z]{10}\s+[\s\S]{0,100}Result[\s\S]{0,100}\s+[A-Za-z]{10}` over 7.4 MB of Rust source) — the 10,000-state budget fills after ~13% of the haystack (`~74` input bytes per state constructed). Because `BitState` is also bypassed above its input-size limit, 44 of the 54 `find()` calls (81% of the haystack) ran on the Thompson NFA, making SafeRE 3.3× slower than `java.util.regex`.

This PR implements RE2-style DFA state-cache flushing (`resetCache`) with cross-search progress tracking and exponential backoff when a cache generation fails to amortize its construction cost.

### 1. Flush and continue (`resetCache`, `transitionOrReset`, `computeNextOrReset`)
Following RE2's `DFA::ResetCache` (`dfa.cc`), when `computeNext` returns `null` due to state-budget exhaustion:
- If the current cache generation has covered enough input to pay for building its states (`>= 10 * cache.size()` positions), `resetCache` clears `cache`, resets the flat transition tables (`transitions`, `offsetToState`, `isAcceleratedStateOffset`), clears transitions recorded on `deadState` and `startStateByContext`, re-creates only the active state `s` via `recreateAfterReset(s)` (re-deriving `sId = s.id * numClasses` in `doSearchMany`'s ASCII loop), and retries `computeNext`.
- The per-character fast path (`transitions[sId + cls]` and `s.next[cls] != null`) is untouched; all reset logic sits strictly inside the cold `ns == null` branch when `cache.size() >= maxStates`.

### 2. Cross-search DFA step accounting & exponential backoff
Because SafeRE caches `Dfa` instances in per-`Pattern` `ThreadLocal`s and reuses them across `Matcher.find()` loops and across `Matcher` instances on the same thread, the profitability guard differs from RE2's single-search `resetp` check in two ways:
1. **Separating DFA productivity from post-bailout progress (`dfaStepsSinceReset` vs. `backoffProgress`):**
   - `dfaStepsSinceReset` counts only input positions actually stepped by the DFA itself (`Math.abs(pos - progressAnchor)` plus completed DFA search spans in `beginSearch`). Spans covered by the caller's NFA fallback after a DFA bailout do not increment `dfaStepsSinceReset`, so NFA fallback work is never mistaken for DFA cache productivity.
   - If `dfaStepsSinceReset >= MIN_PROGRESS_PER_CACHED_STATE * cache.size()` (`10 * cache.size()`) when the cache fills, the generation paid for itself: `unproductiveResets` resets to `0` and the cache is flushed immediately.
2. **Per-generation exponential backoff (`unproductiveResets`):**
   - If a cache generation fills the budget in `< 10 * cache.size()` DFA steps, it is penalized once (`unproductiveResets++`, up to `MAX_RESET_BACKOFF_SHIFT = 20`) and the flush is refused so the active search falls back to the NFA.
   - Unlike a permanent boolean latch — which would leave a `ThreadLocal<Dfa>` permanently stuck with a full cache for the lifetime of the thread after a single dense input — subsequent searches advance `backoffProgress` until `(10L << unproductiveResets) * cache.size()` input positions have elapsed (`200k` → `400k` → `800k` → `1.6M` ... at the 10,000-state budget). This bounds wasted rebuild work on adversarial inputs to a geometric series ($O(1)$ amortized overhead) while allowing long-lived `ThreadLocal<Dfa>` instances to self-heal on later inputs.

## Benchmark Results (`rebar`)

Measured on `curated/10-bounded-repeat/context` (`[A-Za-z]{10}\s+[\s\S]{0,100}Result[\s\S]{0,100}\s+[A-Za-z]{10}`, 7.4 MB Rust source haystack, 54 `find()` operations):

| Engine | Before (`MAX_DFA_STATES = 10_000`, no flush) | After (flush + exponential backoff) | Speedup | Ratio vs. `java/hotspot` |
| `java/hotspot` | 202.92 ms | **202.70 ms** | — | 1.00× |
| `safere` | 673.12 ms | **213.32 ms** (min 202.67 ms) | **3.16×** | **1.05×** (was 3.32× slower) |
| `safere/utf8` | 663.56 ms | **221.84 ms** | **2.99×** | **1.09×** (was 3.27× slower) |
| `safere/vector` | 697.33 ms | **224.48 ms** | **3.11×** | **1.11×** (was 3.44× slower) |

All 54 `find()` calls now finish on the DFA across ~13 cache resets (0 NFA fallbacks, down from 44/54 falling back to the NFA).

## Testing

Added unit tests in `DfaTest.StateBudgetTest`:
- `discardingTheStateCacheAgreesWithABudgetLargeEnoughToAvoidIt`: verifies that forward searches (`longest = false` and `true`, with and without end-context dependencies) at `maxStates = 500` across multiple cache flushes return the exact same match result and position as an un-exhausted `maxStates = 4_000_000` DFA.
- `repeatedFindsOverAResettingDfaMatchTheJdk`: walks an entire `Matcher.find()` loop over a 400 KB source-shaped input across repeated cache resets and checks every `(start, end)` span against `java.util.regex`.
- `unproductiveCacheExhaustionBacksOffAndThenRecovers`: verifies that a `Dfa` instance that exhausts its budget in $< 10$ positions/state refuses resets during the $2\times$ backoff window and then flushes and recovers once the backoff window elapses on productive input.


eaftan (approved):
Thanks for this change, and for highlighting rebar's bounded-repeat workload. During review I found gaps in initial-state cache resets and progress accounting for no-match and post-fallback paths. I pushed a fix with regression coverage for those cases.

I reproduced the exact rebar workload on the corrected PR head:

| Engine | Before (ms/op) | Corrected #908 (ms/op) | Result |
|---|---:|---:|---:|
| SafeRE String | 483.546 | 180.425 | 2.68× faster |
| SafeRE UTF-8 | 442.673 | 177.695 | 2.49× faster |
| JDK String control | 82.364 | 81.571 | Flat |

The gain is smaller than the figures measured in rebar, but it is clear and survives the fixes.

I filed #912 to investigate selective eviction policies such as FIFO or LRU. It would be interesting to see whether retaining useful states helps, though the DFA's cached transitions and active states make safe eviction more involved than replacing the cache map. I also opened PR #914 to bring this workload into SafeRE's benchmark suite with the Rust source attribution and license notice, and filed #913 to consider contributing SafeRE integration to rebar.

LGTM

PR #910, opened 2026-09-19
Subsume required character class when StartPlan.LeadingExpansion has inner literal

cushon:
Similar to #908 this was found looking at gaps between SafeRE and `java.util.regex` on the rebar benchmark suite.

---

### Background
In `MultiAnchorCompiler.extractRejectPlan`, patterns with a literal or fixed-offset start plan (`StartPlan.Literal` or `StartPlan.FixedOffset`) set `skipRequiredCharClass = true` because the required literal is already accelerated by `StartAccelerator` using SIMD or SWAR.

However, `StartPlan.LeadingExpansion` was omitted from the `switch (startPlan)` block. When a pattern had leading nullable or bounded repetition preceding an accelerable literal (e.g. `\s*# noqa` or `\s+[0-9]{2}404_NOT_FOUND`), `skipRequiredCharClass` remained `false`.

As a result, `extractRejectPlan` compiled a redundant scalar `RejectPlan.RequiredCharClass` (e.g. `['#']`). On every single `Matcher.find()` attempt, `Matcher.java` would scan the entire input with a scalar character-by-character loop for `['#']` before the authoritative SIMD/SWAR `StartAccelerator` ran for `"# "`.

### Changes
* In `MultiAnchorCompiler`, consolidate `StartPlan` inspection into exhaustive helper switches (`drivingLiteral`, `drivingCharClass`, `hasFixedOffset`) without `default` branches, strictly mirroring the pattern-matching style in `StringStartAccelerator` and `Utf8StartAccelerator`.
* Set `skipRequiredCharClass = drivingLiteral != null`, ensuring `StartPlan.LeadingExpansion` with an inner literal or fixed offset suppresses the redundant `RequiredCharClass` prefilter scan.
* Include the driving literal in `excludeStartLiterals` without ad-hoc AST checks, enabling downstream required literals (e.g. in `\s*foo.*RARE_KEYWORD`) to be properly retained as reject plans.
* In `MultiAnchorCompilerTest`, add unit tests verifying that `LeadingExpansion` with an inner literal or fixed offset skips `RequiredCharClass` and produces `RejectPlan.None`, and that downstream required literals are preserved.
* In `PatternInternalTest.requiredCharacterClassPrefersTheMostSelectiveMandatoryAtom`, use an unanchored wildcard prefix (`.*`) to isolate testing atom selectivity without triggering `StartPlan` subsumption.

### Benchmark Impact
On `rebar`'s `04-ruff-noqa/real` (`(\s*)((?:# [Nn][Oo][Qq][Aa])...)` on `cpython-226484e4.py`, 890,906 lines):
* Runtime improves from **226.0 ms down to 141.2 ms (1.60× speedup, -84.8 ms)**.
* Eliminates all 890,906 redundant scalar prefilter checks.
* Recovers **62% of the total deficit** between `04-ruff-noqa/real` and `04-ruff-noqa/tweaked`.


eaftan (changes requested):
The redundant character-class scan does look removable when `LeadingExpansion` already drives a literal accelerator, and I like the exhaustive `StartPlan` handling. I also see that this changes required-class selection for `MultiLiteral` and nested leading-expansion plans, and can add a downstream required-literal check; those changes look conservative from the code and tests.

I used eaftan/rebar#1 to run the reported Rebar workloads against the common base, #910, and the clean #910 + #915 tree. I couldn't find your `tweaked-with-ws` arm in the Rebar checkout, so I added an inferred arm by prepending `\s*` to `tweaked`: `\s*(?:# [Nn][Oo][Qq][Aa])(?::\s?(([A-Z]+[0-9]+(?:[,\s]+)?)+))?`. This may differ from your exact pattern. The table shows geometric means of medians from two standard runs; ratios are elapsed time over base, so lower is better.

| Engine | Arm | Base | #910 | #910/base | Combined | Combined/base |
|---|---|---:|---:|---:|---:|---:|
| String | `real` | 172.10 ms | 106.43 ms | 0.618 | 50.93 ms | 0.296 |
| String | `tweaked` | 56.52 ms | 56.79 ms | 1.005 | 57.08 ms | 1.010 |
| String | inferred whitespace | 156.10 ms | 92.14 ms | 0.590 | 51.79 ms | 0.332 |
| UTF-8 | `real` | 126.55 ms | 168.42 ms | **1.331** | 108.91 ms | 0.861 |
| UTF-8 | `tweaked` | 112.98 ms | 112.85 ms | 0.999 | 113.47 ms | 1.004 |
| UTF-8 | inferred whitespace | 114.16 ms | 156.56 ms | **1.371** | 108.98 ms | 0.955 |

The String gain is clear, and the longer confirmation run agrees. But #910 alone makes UTF-8 `real` about 33% slower and the inferred whitespace arm about 37% slower. The combined tree is faster than the original base on those paths, though it gives back much of #915's standalone UTF-8 gain. The `tweaked` control was essentially flat here. 

Could you take a look at the UTF-8 regression and see what's going on?

cushon:
Thanks! I looked into the UTF-8 regression and found another opportunity in #922. This PR is scoped to the exhaustive `StartPlan` handling.

---

I re-ran `04-ruff-noqa` against current `main` (`357fc8fc`, i.e. after #915)
rather than the older base, and the UTF-8 cost is larger there than in your
table: `real` goes 94.98 ms → 161.17 ms, a **1.70×** regression, with
`real-noncapturing-ws` and `tweaked-with-ws` at 1.74× and 1.76×. #915 made the
filtered path cheaper, so removing the filter now costs more than it did
against the common base. The String win also grows, to 0.42.

The tell is that with this PR applied the UTF-8 `real` arm (161.17 ms)
converges on `tweaked` (165.71 ms) — the arm that never had a filter at all.
On UTF-8 the required-class scan is paying for itself; on String it isn't.

Decomposing that with a count-vs-grep probe on the same pattern and haystack:
per-matcher setup costs ~92 ns/line on UTF-8 against ~58 ns/line on String,
while the scan itself differs by only ~7 ns/line. So UTF-8 needs the prefilter
for a reason that has nothing to do with scanning speed — it short-circuits
before the matcher is built — and String is hurt by it for a reason that has
everything to do with scanning speed: `indexOfCodePointClass` was walking
`codePointAt`/`charCount` to test a class with one member, `'#'`.

That points at fixing the scan rather than dropping the filter. #922 routes a
one-character ASCII reject class through `InputScanner.indexOfAscii`, and so
`String.indexOf`, which is intrinsified. On the same benchmark it gets **0.420
on `real`/String — the same win as this PR — with UTF-8 and vector flat.** So
the redundancy argument here was sound but the cost model under it wasn't.

What's left in this PR is the exhaustive `StartPlan` handling. It isn't a pure
refactor: `drivingCharClass` still returns a class for `MultiLiteral`, and a
`LeadingExpansion`'s driving literal still joins `excludeStartLiterals`. To put
a number on the first one I compiled a 385-pattern corpus at both commits and
diffed the flattened plans — 18 patterns change, all `MultiLiteral`
alternations, and in every case the dropped `RequiredCharClass` is
bit-identical to the start plan's driving class. No `LeadingExpansion` pattern
changes, which is the check that `skipRequiredCharClass` is really gone.

eaftan (approved):
Thanks for investigating the UTF-8 regression. Keeping the required-class filter for leading expansions addresses my concern with the earlier version. The exhaustive `StartPlan` switches make sense, and the remaining changes to `MultiLiteral` class selection and downstream required-literal extraction look sound. LGTM.

PR #911, opened 2026-09-19
Seed Surefire provider in CI Maven cache

### LINKED ISSUE #789: Investigate intermittent Maven Central HTTP 429 failures in CI

eaftan:
The existing `prepare-maven-cache` job resolves declared Maven dependencies but misses Surefire's dynamically selected `surefire-junit-platform` provider. On #905, the JDK 22 matrix job fetched that provider after fan-out and received Maven Central HTTP 429.

Run a focused Surefire test in the seed job so it resolves the provider before the matrix starts.

The vector matrix downloads precompiled runtime-test files from an artifact produced earlier in the workflow. That artifact previously expired after one day: when we tried to rerun #905's failed job, it could no longer download the files, so we had to rerun the whole workflow. Retain the artifact for seven days so a future failed vector job can be rerun on its own during that window.

The existing Maven cache for the current POM hash must be deleted manually so the seed job can save a complete replacement. The workflow does not rotate the cache key.

Verification run:
- `mvn -o dependency:go-offline dependency:resolve-plugins -Pcrosscheck-public-api-tests,work-counters,vector-tests,release -DskipTests --batch-mode --no-transfer-progress -q`
- `mvn test -pl safere -Dtest=StringLiteralSearchTest --batch-mode --no-transfer-progress -q`
- Parsed `ci.yml` and confirmed it has 16 jobs, the Surefire seed step, and no cache-key overrides; `git diff --check` passed.

Not run: full SafeRE test suite or the updated GitHub Actions workflow after the manual cache deletion.

Fixes #789


eaftan:
PR Shepherd is now monitoring this PR at `d358314bb4859adad3643326218daadfc18b58e4`.

eaftan:
All required checks have passed. PR Shepherd is about to merge `d358314bb4859adad3643326218daadfc18b58e4`.

PR #914, opened 2026-09-20
Add rebar bounded-repeat context benchmark

eaftan:
## What this adds

This PR adds the Rebar benchmark mentioned in #908 to SafeRE's benchmark suite.  I started doing it to repro the reported result, but figured I might as well include it in SafeRE.  I checked the license for both Rebar and the text in the actual benchmark and have attributed them properly.

## License and attribution

[Rebar](https://github.com/BurntSushi/rebar/blob/463d00f31887e84c38467805b9e3122c314b9521/UNLICENSE) publishes the benchmark definition under the Unlicense. Its [haystack](https://github.com/BurntSushi/rebar/blob/463d00f31887e84c38467805b9e3122c314b9521/benchmarks/haystacks/rust-src-tools-3b0d4813.txt) contains source from [`rust-lang/rust` commit `3b0d4813ab461ec81eab8980bb884691c97c5a35`](https://github.com/rust-lang/rust/tree/3b0d4813ab461ec81eab8980bb884691c97c5a35/src/tools). Rust's [copyright terms](https://github.com/rust-lang/rust/blob/3b0d4813ab461ec81eab8980bb884691c97c5a35/COPYRIGHT) offer MIT or Apache-2.0. We use the MIT option and include the Rust copyright attribution, complete MIT permission notice, source links, commit, size, and checksum in `safere-benchmarks/THIRD_PARTY_BENCHMARK_NOTICE.md`. SafeRE's BSD-3-Clause license is unchanged.


cushon (changes requested):
Thanks, adding this as a regression benchmark for #908 SGTM.

This makes `benchmark-data.json` go from ~350 KB to ~8 MB, which is too big for GitHub to display the diff. What do you think about a `file` recipe kind instead, with the haystack in something like `safere-benchmarks/third_party/rust-src-tools-3b0d4813.txt` marked `-diff` in `.gitattributes` (like `benchmark-results/published/**`)?

That would also avoid the placeholder carve-out.

Also I think the new tests don't run in CI. Nothing invokes `safere-benchmarks` tests — `mvn verify -pl safere,safere-crosscheck,safere-exhaustive -am` builds dependencies rather than dependents, and `run-java-benchmarks.sh` passes `-DskipTests`.

eaftan:
Thanks for the review!  I fixed both items.  I put the large input under `safere-benchmarks/data` as opposed to `safere-benchmarks/third_party` in case we want large inputs in the future that don't come from other projects.

cushon (approved):
LGTM, thanks!

